PR #1, opened 2015-08-28
Update index.php

TheMeq:
Fixed the display of the IP address to remove the port number if it is equal to 25565 which isn't required when entering the IP for a server.


samerton:
Thanks, merged.

PR #5, opened 2015-09-10
German translation added

LPkkjHD:
translated everything to German (Du)


samerton:
Thanks for this, very helpful indeed! I'll go ahead and merge now.

As for the word "ago", it's no longer used on any pages on its own, so it's now redundant.

PR #20, opened 2015-10-30
Add AuthMe support

nerdtron123:
Add AuthMe support, not done but i added a bit, ima search around in the api and find more


samerton:
Thanks for working on this!

Unfortunately the addon system isn't quite ready to handle larger addons such as this one; I've been thinking of including AuthMe support as a core module in an upcoming release however.

PR #21, opened 2015-10-31
Nginx support, as well as improved install process with new error messages and support for the MySQL dependency.

oliverdunk:
Removed htmlspecialchars from output, as the input has already been converted and this lead to "<b>hi</b>" for example, to be shown as "&lt;b&gt;hi..."


samerton:
Thanks for these changes, merging now.

PR #32, opened 2015-11-27
Add missing translation. Fixed Spanish translation

mrlost:


samerton:
Thanks for catching these!

PR #49, opened 2015-12-13
Improvements to Installer

dwilson390:
Made improvements to the installers design and functionality including;
+ Ability to install package to non root directory (eg. example.tld/beta/ as opposed to example.tld/)
+ Requirements & conversion information are now under expandable panels so as to draw attention to the primary question, if the user is upgrading from a previous version or if this is a new install.
+ When the install creates tables in the database, the page is no longer spammed with information regarding every tables creation, just one message on success or failure (hence large deletion in ` core/classes/Queries.php`)
+ Other smaller visual enhancements.


SystemUpdate:
While I like some of the changes, I think (could be wrong) but you removed the requirements testing code and never added it back else where?

dwilson390:
Lines `126` to `196` of the updated file still contain this code.  `upgrade_requirements` and `requirements` contained the same checks, so I merged them and added a check after the code to see where the user should progress to (lines `187` to `195`).

Thanks.

SystemUpdate:
Ahha ok. Looks good to me.. Did you test it?

dwilson390:
I have yes.  Tested both a clean install and upgrading from an old installation I had.  Everything works as expexted, seems ready to land.

samerton:
Thanks for these changes, I'll go ahead and merge them now.

PR #57, opened 2016-01-09
Update language.php

manuelgu:


SystemUpdate:
Looks like you made some improvements, but doesn't look like any translating went on.. ?

samerton:
Thanks for this!

PR #72, opened 2016-01-23
Fix JavaScript for CloudFlare

computerwizjared:
CloudFlare was causing these redirect pages to come up blank. The added tags prevent CloudFlare from caching the JavaScript, allowing the redirect to occur.


samerton:
Thanks, there will probably be a lot of these redirect scripts around the software which will need replacing; I'll try and keep this in mind whenever I come across one in the future.

SystemUpdate:
Maybe add a single php method which takes a url as a argument and then we can easily modify it in the future.

computerwizjared:
@SystemUpdate I would do that but sadly I don't have enough PHP knowledge. If someone else wishes to do that it would probably be useful.

samerton:
There is a redirect class already included in Nameless which does just this, however I've been phasing it out as it won't work if a page header has already been set.
I'll modify the existing redirect class at some point so it'll work in all cases; this should help this issue.

PR #108, opened 2016-02-29
Add Norsk (Norwegian) language 

partydragen:
Added Norsk (Norwegian) language


samerton:
Thanks for this! Merging now.

PR #121, opened 2016-03-12
Patch 4

partydragen:


samerton:
Thanks for these, however I'm going to change the way the Maintenance page is loaded instead of a page redirect.

PR #135, opened 2016-03-19
Fix custom icons in navbar

partydragen:


mentimjojo:
This will not fix the icons, it will only add icons as text to the navbar.

partydragen:
Nopp its working its like samertons https://github.com/NamelessMC/Nameless/commit/beea8ef9ad55f5208b5a091f3fbb01d52d4d6bff
Check that commit u see its like and its working ;P

mentimjojo:
I saw that, but when doing it like you are now  you get a plain text.

mentimjojo:
look i know my programming language

partydragen:
We write  class="" in language.php file and working 
	// Icons - will display before the text
	'home_icon' => '<i class="fa fa-cube"></i> ',
	'play_icon' => '<i class="fa fa-gamepad"></i> ',
	'forum_icon' => '<i class="fa fa-comments"></i> ',
	'staff_apps_icon' => '<i class="fa fa-book"></i> '

But i understand what u mean and it working too but if u fix <i class=""> instant u need to edit all addons again and navbar an modules

mentimjojo:
I know that, but it's better and cleaner then html in an language file :package:

partydragen:
Sure Both work we can see that samerton want

samerton:
Thanks, I've pushed my local files which I forgot to include previously now.

PR #141, opened 2016-03-22
Fix $p undefined when pointing to a specific post.

mentimjojo:
Fixed $p undefined when in the link is pointing to a specific post. Now you got a white page with only a error.

### DIFF
--- a/pages/forum/view_topic.php
+++ b/pages/forum/view_topic.php
@@ -42,6 +42,23 @@ if(!$list){
 	die();
 }
 
+// Get page
+if(isset($_GET['p'])){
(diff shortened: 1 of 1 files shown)


samerton:
I can't reproduce the issue you mention with the undefined $p variable, how do you get this error to occur?

mentimjojo:
When running in normal mode you don't get any errors, I found the error in debug mode when i was editting a post, and saved it. Normally it sends you back to the right post but in this case it crashed the site.

samerton:
Confirmed the issue now, thanks.

PR #142, opened 2016-03-22
Fixed password forgot function.

mentimjojo:
Some language text was missing so the mail etc was not send, and database not updated.


samerton:
Thanks for spotting the issue, however the 'password_email_set' language entry should only be displayed after a successful reset.
In this case the language value that needs outputting is `$email_language['thanks']` instead of $user_language...

PR #146, opened 2016-03-27
Updated Norsk language

partydragen:


samerton:
Thanks for these updates.

PR #156, opened 2016-03-31
Add reminder to remove install.php

Fiiily:


samerton:
The installer file deletes itself, assuming permissions are correct. 
I'll change this notice from one telling the user to delete it, to one which will only display if the installer couldn't delete itself.

PR #159, opened 2016-03-31
Display custom avatars in Members List

partydragen:
Change Members List avatar from always showing Minecraft avatar to showing a
custom avatar if the user has one.

(I have tested its much and working perfect) :P


samerton:
Thanks, merging now.

PR #160, opened 2016-03-31
Temp Fix: Display custom avatars in profile

partydragen:
Change profile avatar from always showing Minecraft avatar to showing a
custom avatar if the user has one.

(This is only a temp fix until we get the new profile page)


samerton:
The new profile system shouldn't be far away, thanks for the fix though.

PR #178, opened 2016-04-22
Issue template

ThemInspectors:
Added an issue template.


Schwertspize:
This doesn't seem like an appropriate issue template. Better would he something like this

``` md
# wanted behavior

# actual behavior

# specification of your environment (you can find it in the admin panel)

# if there is a relevant log, paste it here, attach it or use a paste service) 
```

samerton:
Thanks for creating this, I'll merge #229 which includes an example.

PR #203, opened 2016-05-25
.php extensions removed from browser urls

jesseke55:
Like the commit title says because of 404 errors.


samerton:
Thanks, usually the .htaccess file/nginx configuration finds the install file for you, but I agree that this needs changing as the URL can be confusing for people experiencing 404 errors, as they may be tempted to move the install file.

PR #211, opened 2016-06-06
Different types of forums

IonLogic:
This adds support for different types of forums. Currently, it only adds categories alongside discussion forums, however it could be used for other types as well. Categories still exist as normal forums, however they can not be accessed by URL or link and threads can not be moved into them. Their only use is as a visible grouping of forums.

This change is achieved by making one adjustment to the data, being that the column 'forum_type' has been added to the forums table. It records the value as either 'category' or 'forum'. The appropriate changes were made to the database initialisation and an update file has been added to support the changes.

I would recommend in a future update to completely overhaul the way forums work to a more XenForo-esqe nodes system. However this systems works and can fulfil those needs.


samerton:
Perfect, thanks!

I'll look into overhauling the forum system (thinking some sort of forum setup wizard would be best for this) in the near future.

PR #212, opened 2016-06-07
Changes "user_avatars" from true to 1 when installing

nfell2009:
When Nameless is installed, it sets "user_avatars" to 1 if the option is true in the database. But the actual code checking if users can use their own avatars is looking for the value of 1.


samerton:
Thanks for spotting this, I'll add the required code to the update files now.

PR #218, opened 2016-06-13
Updated the nginx.example

jesseke55:
Edited it with the most common way for LEMP users with php5 and php7 fpm instruction.


jesseke55:
Editing the file in my forked repo for better explanation.

Issue #220, opened 2016-06-13
Options for heads with helmets


ConnorLinfoot:
I personally hate that heads dont show skins with helmets, could add an option for this to be switched?

HappyHarm:
I think you can change that yourself by changing the links from 311 & 314 https://github.com/NamelessMC/Nameless/blob/master/pages/profile.php to something like http://cravatar.eu/helmavatar/<nickname|uuid>

HappyHarm:
Nvm, seeing much more cravatar links for the user profile picture. Would be a lot of work to edit 'em all.

samerton:
This was added back in 1.0.11 (commit 90b49e7d0a89bf77f7f89ecd026f36bad67d83ae)

PR #226, opened 2016-06-13
Fix for issue #220.

Azoraqua:
If any problems occur, feel free to deny it or change it back.


HappyHarm:
This is not a solution as I already said in the issue. this will just change it on one place and change it for **everybody** in stead of adding a option for it.

PR #227, opened 2016-06-13
Fixed the installer

partydragen:
Installer was broken because him used ` because of '

in 'forum' it was `forums` in default it should be ''


Azoraqua:
@partydragen Nice catch. @samerton to merge :)

PR #229, opened 2016-06-13
An Issue template written with example.

Azoraqua:
**Note:** None of the content in the example section is real. It's just for demonstration purposes made.


jesseke55:
I like this idea for faster bug fixes etc!

PR #234, opened 2016-06-15
Added Cancel button when editing a post

partydragen:
Added Cancel button when editing a post

**Note:** new_topic.php missing traslation on cancel button


samerton:
Thanks :)

PR #238, opened 2016-06-19
Update nginx.example

jesseke55:
Added a IPv6 listen to prevent issues when IPv6 has been rolled out.


samerton:
Thanks for this

PR #240, opened 2016-06-19
Update LICENSE.md

jesseke55:
Updated the year from copyright


samerton:
Thanks for spotting this.

PR #241, opened 2016-06-19
Big update

jesseke55:
Due the reason that you didn`t add a credits page so I decided to make this for you :)


jesseke55:
Feel free if you want to add more information :D

Ariuw:
I am not in the Spanish Translation Team :(

samerton:
Thanks for this, however instead of the page being a custom page (which would be configurable through the AdminCP), I'd probably rework this so it's a normal page (from within the pages folder).

@Ariuw Oops, sorry about that, I'll add you now :)

samerton:
Just another thought, if a contributor list is added (such as the one within this pull request, or potentially on the Wiki), the credits link could point to this instead of it being bundled within Nameless itself.

PR #251, opened 2016-06-21
Error 404 benig caused for phpinfo

jesseke55:
Like as the title says


samerton:
I presume this only occurs on certain server configurations (I've never seen this issue before, nor can I reproduce it), but the fix will help with the specific configurations this occurs on, thanks :)

Schwertspize:
I can at least understand why this is occurring. 

All of your pages are handled and get loaded by your php, which makes the file pages/signin.php available under /signin (and removes the .php suffix). This would also occur to pages/admin/phpinfo.php

PR #254, opened 2016-06-22
Allow people to use colors in private messages

partydragen:
Example me i pasted in a text and it was white text and page was white so it was hidden :/ and it was no options to edit the color


samerton:
Thanks

PR #267, opened 2016-06-24
Disable debug messages from the mail.

Tim-NL:
I think debug for PHPMailer should be off, now anybody that has SMTP & 2FA enabled can see the communication between server & client. 
(This is my first Pull request, I'm not sure I'm doing it correct)


samerton:
Thanks for this

PR #272, opened 2016-06-26
Create language_dutch.php

jesseke55:


jesseke55:
I hope that these additions will help with some extra translations for addons.

jesseke55:
Well.. I don`t have any idea for how to adding it into the initalisation.php or something but I think that those translations would help enough

samerton:
Thanks, I'll work on a way of automatically loading translations within addons soon.

PR #275, opened 2016-06-27
Add a "report closed" label

PugaBear:
If a moderator clicked a report alert after it had been closed, the button to close the issue still appeared. This fixes that.


samerton:
Thanks for spotting this!

PR #276, opened 2016-06-27
Fixed some bad grammar

melerpe:
Changed: Bercith Titel -  To: Bericht Titel
Changed: Start - To:  Home (We dont use Start for homepages on dutch sites!)


samerton:
Thanks for this!

PR #283, opened 2016-06-28
Sharing on social media's 

Root3287:
Added a option for the user to share the post(NOT REPLIES) to Facebook and Twitter.


samerton:
Thanks!

PR #287, opened 2016-06-28
White Text for report section

Ariuw:
It wasn't working well so what I have do was just add white text


samerton:
Thanks for spotting this

xBreaksx:
I think this is the same for the comments on Applications using the theme Cosmo.

PR #296, opened 2016-06-30
Change tachometer function slightly

PugaBear:
It was annoying me that the tachometer was clickable, so I was messing with it to make it not clickable but still show up correctly (cause if you remove href alltogher, the tachometer decides it doesnt wanna be in the right place anymore). I honestly came across this solution by accident when I typo'd, but hey, it works. Idk if its valid though. Feel free to send this to the trash can. Just something I liked having.

### DIFF
--- a/core/includes/template/footer.php
+++ b/core/includes/template/footer.php
@@ -21,7 +21,7 @@ $footer_nav = '
 
 if(isset($page_loading) && $page_loading == '1'){
 	$footer_nav .= '
-    <li><a href"" data-toggle="tooltip" id="page_load_tooltip" title="Page loading.."><i class="fa fa-tachometer fa-fw"></i></a></li>';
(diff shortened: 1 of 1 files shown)


samerton:
Thanks for this, however I don't think href"" in terms of HTML formatting is the best answer.

I've taken a quick look, and removing the href attribute altogether (so the tag is `<a data-toggle="tooltip" id="page_load_tooltip" title="Page loading..">`  seems to work okay. Could you try this again to confirm, and then I'll alter the code accordingly?

PugaBear:
Huh, okay, I really thought removing href broke it. Maybe I did something else. That works too

PugaBear:
I duplicated what I remember by removing the a tag alltogether. I have no idea why I did that in the first place- It was probably 3am.

samerton:
Thanks :)

PR #312, opened 2016-07-08
Little fixes

dgateles:
Some little fixes. The language is now 100% with the last Nameless changes.


samerton:
Thanks for this :)

Issue #264, opened 2016-06-24
Lots of stuff. Sorry, I'm picky. 


Pugabyte:
Small bugs: 
- [Calendar window is partially blocked by navbar](http://i.imgur.com/VINyJ70.png).
- [Forums box has an ugly category format](http://i.imgur.com/EHC6cXS.png)

Version 1.0.10, theme Darkly

Error: Tried to send a PM to someone, got this error: _SQLSTATE[HY000]: General error: 1364 Field 'content' doesn't have a default value_. I probably config'd something wrong, but not sure what.

And "a few" feature requests. There are duplicates, but I'm just going to say them again. I'm also aware I can edit the source/database to change some of these (and already have), but this can be for general future improvements, for users who don't know how to edit the source. :)
## Applications
- Multiple applications 
- Better support for different question types, like radio buttons, calendar windows, etc.
- Restrict an application to certain ranks, like a forum.
## Nav Bar
- Ability to re-name built in pages, such as "Play"
- Ability to restrict a page's nav bar link visibility based on rank
- Custom redirect links (See my manual solution in #249) 
- More dropdowns 
## Infractions
- Add option to disallow players who are banned in-game to post a new thread, and only allow them to appeal via application (obviously, this would have to come after the application improvements)
- Also add ability to restrict the "Infractions" footer link to certain ranks.
## Ranks
- Sync ranks ingame with vault groups
- Ability to add more than one rank to a user
- Custom page for a group (ability to redirect (e.g. to a wiki))
- Ability to re-order the ranks (changing the ID in the database (and don't forget about users too (and the initial rank you get assigned when you sign up))), and the same for the nav bar pages. 
## Users
- Format player's uuid with dashes 
- Ability to change user's banner on their profile page (Maybe there already is a way? Am I missing something?)
- **Promote someone ingame when they sign up!!**

Thank you for reading. I love the website so far, don't get me wrong, I just would like to help improve it a little bit. :)

I will very likely be posting more bugs/suggestions as I come across them / discuss with my staff team.

Schwertspize:
Pm error is probably an empty content? 😮 

Restrict to ranks.... It doesn't change anything if you accept it, afaik you have to change the group yourself

samerton:
Thanks for these, here are my initial thoughts:

- Calendar window is partially blocked by navbar:
Should be fixed in df7cf761cc9be80feb64432c2484a22cdc3720fc

- Forums box has an ugly category format
[...]

PR #314, opened 2016-07-09
Add uuid_formatted column in users table

PugaBear:
Not used for anything currently, but can be useful - I suggested this in #264 and you said you'd add it- but once I got familiar with the setup I figured I'd just do it myself. Obviously the table will need to be altered by an update script. I'm currently using this to parse my PEX file and pull a rank from it.

Hopefully I got everything.


samerton:
Thanks for this.

One reason why I've been reluctant to do this is that there's currently a `formatUUID` function in the `ProfileUtils` class (within core/integration/uuid.php), which I've used in a couple of places to obtain the formatted UUID from the UUID stored in the database. Something that will have to be decided is whether it's worth storing the formatted UUID alongside, or whether the function would suffice.

PugaBear:
Oh, I was not aware of this. :p

PR #319, opened 2016-07-13
Improvements to the forum permissions 

ConnorLinfoot:
I've moved all the permissions into one table, looks much nicer than each one being a new line in my opinion. 

EDIT: I've only changed the view, not the DB table. http://prntscr.com/bsdm2b


samerton:
Looks great, thanks for this!

PR #323, opened 2016-07-15
Make the disable avatar button in the admin panel more intuitive

PugaBear:
If the user doesnt have an avatar at all, it won't display the button
If the user does have an enabled avatar, it gives the option to disable it
If the user doesn't have an enabled avatar but the file for one exists, it gives the option to enable it.


Headhunterzpw:
You have conflicts in your PR

PugaBear:
Dsmn, didn't notice.  I know it was conflict free when I created it. A recent change must collide. Ill redo this with an updated branch.

PR #329, opened 2016-07-15
Android status bar colouring support

RobinMCNetwork:


samerton:
Thanks for this

RobinMCNetwork:
You might want to add something in the settings to change this, I didn't do that because I don't know this software's SQL layout. I figured some shade of gray would be appropriate for most themes

PR #331, opened 2016-07-18
Update README.md

Headhunterzpw:


aycgit:
Can u show us that ur really jesseke? I think that would be a bit better :)

aycgit:
And i would remove (the old jesseke) because it doesn look that good

samerton:
Thanks, I'll update your username in the readme, however I'll drop the "(the old jesseke)" text to keep things a bit tidier.

PR #337, opened 2016-07-20
Make the disable avatar button in the admin panel more intuitive

PugaBear:
If the user doesn't have an avatar at all, it won't display the button
If the user does have an enabled avatar, it gives the option to disable it
If the user doesn't have an enabled avatar but the file for one exists, it gives the option to enable it.


samerton:
Thanks for this, merging now.

PR #361, opened 2016-07-26
Added Name History list if you want it

partydragen:
Added Name History to see all players name changes
 
You can see how it looks here http://cubedcraft.com/profile/partydragen/

If you want to see a player who have changed name every time you can check out him http://cubedcraft.com/profile/Frozytime/


partydragen:
1. How do people add muliple commits in 1 pull request without using client :/

2. U may need to fix a option to name history not work if them using offline mode thing with no UUID

samerton:
Thanks for this, I'll merge this and make some alterations.

PR #367, opened 2016-07-29
Lots of translations

aycgit:


samerton:
Thanks for this

aycgit:
Going to add more later but this is what i had time for

PR #374, opened 2016-08-09
update to line 193

godleydemon:
Needed to change line 193 to work with twitters new widget system. No need for widget ID's anymore, which causes the widget to not display in the side bar when enabled in the admin panel. Utilizes the twitter url for it's twitter feed.


samerton:
Thanks for the fix!

PR #375, opened 2016-08-10
added Infractions to pull down

godleydemon:
Added Infractions to the pull down if admin logged in for easier access, also updated the language files for translation from various languages. 

Couldn't find a variable for if Infractions was enabled or not, so it just shows up all the time atm. Curious on if there is one and I'm just being an idiot and missing it.


samerton:
Thanks, I'll add a variable check now.

PR #378, opened 2016-08-11
Added ''Well'' in Wall posts, About, Name History To profile.php

IsS127:
Added  ''Well'' in Wall posts, About, Name History


samerton:
Thanks

PR #386, opened 2016-08-16
Added images in profile posts

ElectronicWizard:
Its entirely likely this code could be more efficient in PHP. This is simply my first go at it, added a feature I thought might be useful to the full repository.


ElectronicWizard:
Should note: this only works if the image link is the ONLY content in the Profile post. Again: this is my first go at PHP, you would definitely want to add a method to enable both text and images in the same post. This was more as a proof of concept than anything else.

samerton:
Thanks, I've altered the purification which should enable the use of HTML in posts in 888bebace84f96f6be5ad6d4a7112125ae730a96

PR #387, opened 2016-08-16
Finished German translation.

BukkitTNT:
This was many work :3


Cl1608Ho:
Looked over a few paragraphs, looks good! 

(as usual, all of these German translations don't sound so good, but it is as it is. This is one of the better translations I've read in the past time)

samerton:
Thanks!

PR #390, opened 2016-08-16
Fix banner URL

rodymol123:


samerton:
Hi,
The banner URL should work with the .png extension provided there is a file named ".htaccess"within the **core/integration/banner** directory containing [these rules](https://github.com/NamelessMC/Nameless/blob/master/core/integration/banner/.htaccess).

samerton:
I'll mark this as closed, as the included .htaccess should do the trick

PR #394, opened 2016-08-17
Add Title to users in admin panel

yshf9:
Needs to add user_title to all languages but I only know English


yshf9:
Needs to add user_title to all languages but I only know English

samerton:
Thanks for this

PR #400, opened 2016-08-19
Only show .panel-body with subforums if $parent.forums > 0

rodymol123:
Looks a bit cleaner for forums that don't have subforums


samerton:
Thanks for this

PR #422, opened 2016-09-25
Update infractions.php

dasilvaj4:
Add support Ban Management offline players


AlliedCore:
In the aspect of offline ban management do you mean banning players when they are offline or being able to manage banned players while offline?

dasilvaj4:
no, my server is offline mode

AlliedCore:
@dasilvaj4  that makes sense. I think there is a way to make a port extension in order to show a live display of the offline server

samerton:
Thanks for this

PR #437, opened 2016-10-04
Don't alert the user if they are posting on their own profile

PugaBear:
Its pointless. :x

Hopefully I've got this commit right, it's 3:30 AM. Anyone wanna double check for me?


samerton:
Thanks for this!

PR #438, opened 2016-10-04
Change scroll destination

PugaBear:
When you link to posts, the nav bar covers up a part of the reply panel.
image

This adds a bit of padding to the scroll so it shows up like this
image


MuhsinunCool:
This shud be merged! :D mew mew

samerton:
Thanks for this, I've just run a quick test and it works perfectly on mobile too :)

MuhsinunCool:
Yay! XD

PR #444, opened 2016-10-07
Update forum main posts

mentimjojo:
Fix for post counting, the old way also counted posts from deleted topics.


samerton:
Thanks for this!

PR #448, opened 2016-10-13
Updated installer?

partydragen:


Fellings1:
@partydragen I think it might be a good idea to leave beta in for any new releases in the near future giving indication that it is a beta file or version of the product. Ill leave @samerton to leave a decision on this matter.

PR #450, opened 2016-10-13
Create .gitignore (for future use)

nick-2017:
This is for if we need it in the future.


aycgit:
Really unneeded and if we need it we can create it sorry

aycgit:
👎

PR #456, opened 2016-10-26
Add checks if the user can view staff applications.

Cl1608Ho:
Currently, namelessmc is vulnerable that users who are in a group that can access the mod cp but aren't allowed to view staff applications don't see the link staff applications but can still browse them by simply going to /mod/applications. This adds the check that they can also view staff applications. If the user can't view staff apps, they cannot accept or deny either because view checks are made earlier.

Ps: dunno about the last no newline / newline edit. Github made that up

### DIFF
--- a/pages/mod/applications.php
+++ b/pages/mod/applications.php
@@ -8,7 +8,7 @@
 
 // Mod check
 if($user->isLoggedIn()){
-	if(!$user->canViewMCP($user->data()->id)){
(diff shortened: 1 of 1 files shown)


Cl1608Ho:
Well I already did something on my end, but I don't think it was intended that if you set the settings not to allow someone to view applications, that it just hides the link but basically allows them to view them. 

@samerton might I bump this?

samerton:
Thanks for this!

Issue #449, opened 2016-10-13
Question - Change the author title


partydragen:
Hey @samerton idk u want it or not but can u change the author on all pages to be website name instant?

Because when we always sending the link to like discord, skype then its stand Samerton as tittle because its loading the author i really don't know why its like that :/

And its really looks bad about its stand Samerton as title instant of Website name

You can see it here: http://imgur.com/a/XRIVV

Idk its any other may to solve it or its a bug but i have tryed and its the author code that set the title Samerton

Emoxus:
@partydragen Are you just a user of the software? I was going to suggest that if you knew how to read the code maybe you could edit it because this is a open source project.

samerton:
Fixed in #462

PR #462, opened 2016-11-02
Changed author to Website name

partydragen:
Changed all author from Samerton to <?php echo $sitename; ?> on all pages
That solving this problem:
https://github.com/NamelessMC/Nameless/issues/449

I don't want to push this, I want samerton to agree with this
Its sure other better ways to solve this


Cl1608Ho:
Wait, this isn't in some sort of templating? :o

partydragen:
What

Fellings1:
@samerton Fixed the "Samerton" title of the software.

samerton:
Thanks for this, merging now. This shouldn't be an issue with version 2.0, which will have the author tag in the global header.

Fellings1:
Close

PR #463, opened 2016-11-02
Added Help page

partydragen:
I see some other Softwares have a help page on admin panel so why not have it here too xD
So i created a simple page, Its may better to have a knowlegebase addon thats are planned on NamelessMC Website in v2

Then help page can load that page from NamelessMC to always display latest information/tutorials?

Its currently looks like this: http://imgur.com/a/GlA3d (without bakground)
(Its not using the language files its english only but all the urls is english so whatever

### DIFF
--- a/pages/admin/help.php
+++ b/pages/admin/help.php
@@ -1,0 +1,104 @@
+<?php
+/*
+ *	Made by Partydragen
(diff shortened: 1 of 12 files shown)


Fellings1:
Thank you! @partydragen

PR #465, opened 2016-11-03
Update language.php

dasilvaj4:
some fixs


Fellings1:
Thank you! @dasilvaj4

PR #470, opened 2016-11-09
Language.php Update | German

BukkitTNT:
I have translated some things and improved existing translations.


samerton:
Thanks for this!

Cl1608Ho:
Uhm all of these aren't quite right. 

The first one means "keine gelesenen Benachrichtigungen", you want to translate read messages, not unread messages. 

The second one means only banning, not warning. 

The third one needs to be "Seite wurde in x Sekunden geladen"

Fellings1:
@Cl1608Ho  If possible please do make sure to check for any other errors that may be within the translation. Thanks!

PR #511, opened 2017-01-03
updated Dutch translation

DoraKlikOpDora:
Removed and fixed some strange-formulated sentences, grammar-mistakes and other typo's

### DIFF
--- a/styles/language/Dutch/language.php
+++ b/styles/language/Dutch/language.php
@@ -8,6 +8,8 @@
  *
  *  Updated and removed syntax error by melerpe
  *
+ *  Updated by DoraKlikOpDora/Besbos
+ *  http://minevibes.com
  *  Dutch Language
  */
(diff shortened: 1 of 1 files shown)


DoraKlikOpDora:
@dangerORclose I have some other words English as well, because it is my opinion that it is 'better' to keep some words English. User is a well-known word and everybody knows what it means. I think UserCP should not be translated because it is a kind of thing of the software, just like you not call MineCraft something like mijnvak.

dangerORclose:
Owkay thats what i tought too.
But anyways MineCraft stays minecraft whatever language it is.. Maybe not for Chinese but they are just special.

DoraKlikOpDora:
@dangerORclose Couldn't come up with a better example. Blame my creativity

samerton:
Thanks!

PR #527, opened 2017-01-14
Remove extra line

PugaBear:
It bugs me :D


Fellings1:
Seems alright but we can always reverse the changes in case of any situation.

PR #530, opened 2017-01-23
Update navbar.tpl

dasilvaj4:
Fixed a bug in navbar, because in the file "core/includes/template/navbar.php" already includes the html tags


Fellings1:
@samerton Check it out.

samerton:
Looks good, thanks

PR #562, opened 2017-02-19
Added members list module

partydragen:
This is almost the same member list from v1 and i create it was pull request due i don't know we should use it


Awtra:
@samerton Should we use it?

samerton:
@partydragen @Awtra Going to rework the module as a new separate module, I'll go ahead and close this PR

Awtra:
Thanks

PR #572, opened 2017-03-11
Complete French translation of language.php

pandalyser:
Here is a complete French translation of language.php, since the last translator (CreaModZ) hadn't finished it. I hope it will serve players well.


samerton:
Perfect, thanks!

PR #585, opened 2017-03-24
Remove IRC from README because no one from the project uses it

PugaBear:
I have had a bouncer on the channel for many months now and its shitposting and people asking questions then leaving 20 seconds later. Better to just point people to discord.


samerton:
Thanks, probably wise.

PR #602, opened 2017-04-11
Update Chinese translation

BirkhoffLee:
Hey there, here's another Chinese translation update.

By the way, it would be much more convenient if you give me write access to this repository so I can update the translations fast. Every project I translate does this -- and you don't worry if your project is messed up by me, they are all recorded and you can easily rollback. I have found that a lot of pull requests in your repository are lanugage translate updates, it's such a waste of time to translators.

Oh if you wonder, I'll translate the remaining strings in v2 after it's in pre-release. I need to understand what it's talking about, you know, Chinese is a complicated lanugage.


samerton:
Thanks again for the translation.

I will improve the translation process around the time of v2's released, I have a few things in mind :)

PR #609, opened 2017-04-19
Efficient way of using Options

Headhunterzpw:
Made it a bit more efficient for the Options -Indexes by making it on 1 single line


samerton:
Thanks for this

PR #616, opened 2017-04-25
Patch for Norwegian-language

OscarWoHA:
A lot of errors were found during this process but I think i managed to fix the worst parts. Please merge!


samerton:
Thanks for the updates

PR #618, opened 2017-04-26
Update registration_disabled translation for Chinese

BirkhoffLee:


samerton:
Thanks for the update :)

PR #621, opened 2017-04-27
Update front_page.php

Aviorleking:
change the array show to show the last to the first news page


Aviorleking:
i saw something better

PR #622, opened 2017-04-27
Correction to show real latest news

Aviorleking:
reversing the array to show the real latest(not the first)


modules/Forum/classes/Forum.php
```diff
@@ -469,4 +469,4 @@ public function getPosts($tid = null){
 		}
 		return false;
 	}
-}
```
Aviorleking: wut

Aviorleking:
you missed one word in the EnglishUK translation i added it

samerton:
The order of news on the homepage should be correct without calling array_reverse:

Without array_reverse (latest news post is on top):
Before

With array_reverse (latest news post is now on the bottom):
After

Aviorleking:
it happen the reverse for me without i have my latest at the bottom and with it it is at the top

i'm gonna give the proof later not on my computer

Aviorleking:
with reverse : 
image

without : 

image

(first was the 1st april last was the 4th april)

i am using the latest of latest(cloning all changes)

Derkades:
Could it be something with your database server? For example, in the code it may say `SELECT * FROM news ORDER BY date`. This is supposed to sort in ascending order, but may for some reason be descending for you

samerton:
I've just been able to reproduce the issue, and I'll investigate at some point in the near future.

Aviorleking:
Ok thanks

samerton:
Thanks for bringing the issue up, I've just found the actual cause of the issue and have now solved it (https://github.com/NamelessMC/Nameless/commit/4711028eef58e9966089fa6288b5b2fed07814e8)

PR #630, opened 2017-04-29
Delete 1920x1080 background

Derkades:
I remember Samerton saying that this image is not used by anything, so I figured we should remove it to avoid confusion.


samerton:
The 1920x828 image is used by the homepage banner, it is the 1920x1080 image which isn't used at all :)

Derkades:
Oh, I mixed them up. I'll delete the 1920x1080 one instead

samerton:
Thanks for this

PR #633, opened 2017-05-02
Improved wording in the Norwegian translation

tlystad24:
I went through and quickly edited most of the grammatical errors in the translation. It seems like the previous translation was pulled from Google Translate, making it annoying to read and sometimes hard to understand.

Some words are hard to find replacements for, perhaps others have better suggestions. 

I'm not too sure if I'm completely finished with this translation, but I think most errors are now fixed.

It would be great if another Norwegian could have a look at this and correct whichever errors I've placed.

### DIFF
--- a/styles/language/Norsk/language.php
+++ b/styles/language/Norsk/language.php
@@ -1,7 +1,8 @@
 <?php
 /*
- *	Translation created by Partydragen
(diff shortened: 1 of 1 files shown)


samerton:
Thanks for the updates

PR #647, opened 2017-05-14
index.php improvements

HarveyHans:


samerton:
Thanks for the pull request, however as these are more down to personal preference, plus it may lead to inconsistencies in the code style across the package, I'll close this.

PR #667, opened 2017-06-15
Translated debugmode, minecraft servers and more

Thesevs:
Translated the query, debugmode and some other lines.


partydragen:
Thanks :)

PR #671, opened 2017-06-20
Integrate Infractions tab into Profile page

relavis:
I don't think I added any check if infractions is actually enabled so that might need to be added.


samerton:
Thanks, I'll make a couple of modifications such as to check if the addon is enabled etc

PR #685, opened 2017-06-28
Translated email

Thesevs:
Translated the email section


samerton:
Thanks for both translations :)

PR #693, opened 2017-07-05
Update languages: File general.php

dasilvaj4:


samerton:
Thanks for the updates :)

Issue #705, opened 2017-07-13
Missed php mail headers


baxandrei:
PHPMail header MIME-Version: 1.0 and Content-type: text/plain; charset=UTF-8 is missing from these files:
/pages/register.php
/pages/signin.php
/pages/api/v1/index.php

It looks like:`				$headers = 'From: ' . $siteemail . "\r\n" .
					'Reply-To: ' . $siteemail . "\r\n" .
					'X-Mailer: PHP/' . phpversion();
				mail($to, $subject, $message, $headers);`

But these need to look like: `				$headers = 'From: ' . $siteemail . "\r\n" .
					'Reply-To: ' . $siteemail . "\r\n" .
					'X-Mailer: PHP/' . phpversion() . "\r\n" .
					'MIME-Version: 1.0' . "\r\n" . 
					'Content-type: text/plain; charset=UTF-8' . "\r\n";`

On my server this cause an message (email) split before it to be send, and this is a bad thing, especially when is send an confirmation code to acces to activate account and the code is spited in 2 parts, and users don't realise that. 

The message will look like this: `Hi
Thanks for registering! In order to complete your registration, please =
click the following link:

[...]/validate/?c=tdx6VN0bkzTfXmwyY9JlEKLpr1WsRa73Z=
cM42Bgh5iUOvSeHuADjQFInoGqC

Thanks,
[...]`

samerton:
Thanks, I'll take a look into it :)

samerton:
Hopefully resolved in the PR :)

PR #714, opened 2017-07-19
Fixed mail headers.

baxandrei:
Fixed mail headers. (On php with msmtp sendmail and mail without MIME-Version and Content-type text will be striped and any url or long text will be affected.)

Reported here: https://github.com/NamelessMC/Nameless/issues/705


samerton:
Thanks :)

baxandrei:
With pleasure :D

PR #715, opened 2017-07-20
Updated Romanian lang and fixed mail headers.

baxandrei:
Update general.php and user.php in Romanian lang.

Ignore editor.css, i try to repair the smiley button with no icon but it`s from plugin because have a wrong icon.

Update contact.php, forgot_password.php and register.php for fix mail headers. (On php with msmtp sendmail and mail without MIME-Version and Content-type text will be striped and any url or long text will be affected.)


samerton:
Thanks again

baxandrei:
With pleasure :D

PR #717, opened 2017-07-20
update for romanian lang (again)

baxandrei:
I modified forgot_password translation to remove special romanian caracters because is displayed with wrong characters after email send. <<< user.php

Updated sign_in and log_out To use the same verbal time with register. <<< general.php


samerton:
Thanks :)

baxandrei:
With pleasure :D

PR #720, opened 2017-07-20
Adding missing terms

baxandrei:
I cecked my translation file and i find missing terms, now all terms is up to date!


samerton:
Thanks :)

baxandrei:
With pleasure :D

PR #727, opened 2017-07-21
Update profile.php

baxandrei:
Change group_html to group_html_lg and now group_html_lg and group_html is used. group_html_lg on profile view and group_html on forum pages


baxandrei:
Hmm.. anyone? Forgot abut this PR?

IsS127:
Can i see a preview?

samerton:
Thanks for the PR, I think instead of changing the profile pages from `html` to `html_lg`, I'll simply remove `html_lg` altogether and just use `html` :)

PR #729, opened 2017-07-22
Added some features to the donate module (CraftingStore)

Tim-NL:
I noticed that you implemented CraftingStore support, but it misses some features (mainly because of the API), so I added some of them. Now it's more complete ;)


samerton:
Thanks a lot :)

PR #740, opened 2017-08-08
Dutch translation for widgets

Derkades:


Derkades:
Why did I create a pull request, I have write access..

PR #783, opened 2017-08-29
Update index.php

baxandrei:
Add { and } betwen if and else to end


baxandrei:
Umm.. any editor?

samerton:
Sorry it's taken so long to get back to you.

This is something which is down to personal coding style, and if the code following the if statement is short, I personally use this format. As adding brackets is only a style change, I'll close this PR as the functionality doesn't change.

PR #788, opened 2017-09-24
Disallow access to .php + .tpl in custom/templates

BrightSkyz:
This is non tested


samerton:
Thanks, a slightly different method added in c99151a57ec21408e7c8638d10fd724f6f441dcc

PR #808, opened 2017-10-06
Translated a few sentences and words

TheSander562:


Derkades:
The other pull request is still open, you can push changes to that branch instead

PR #842, opened 2017-10-30
Add second group

smessie:
This adds a second group for users, the user only gets the permissions of the first group.
Useful if you want to display more groups (like a donator group + a staff group) on the profile page or forum.


pimkennedy:
Would be extremely nice with a second group. More groups would be even better.

Derkades:
What if we add a secondary table with the user id and group? This way you can assign unlimited groups to people. Example:

1: Developer
1: Owner
2: VIP
3: Moderator
3: Donator

samerton:
This is a nice update, until the upcoming version 2 update which will allow for unlimited secondary groups this is perfect. Thanks! :)

PR #921, opened 2017-12-15
V2, Fixed alot of requires to use ROOT_PATH constant for requiring.

TortleWortle:
This adds support for webservers, tools that use a different PHP entry point.
This removes the legacy $path (core/init.php:35) variable I'm not sure how that will affect legacy applications.

### DIFF
--- a/404.php
+++ b/404.php
@@ -55,7 +55,7 @@ define('PAGE', 404);
 	));
 	
 	// 404 template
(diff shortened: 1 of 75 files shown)


TortleWortle:
Another PR coming after this that fixes potential ip-spoofing exploit.

samerton:
Thanks for this

PR #922, opened 2017-12-15
Translated the new added sentences to dutch

TheSander562:
I dont know where nickname is being used, so maybe that translation is going to be changed.


samerton:
Thanks

PR #923, opened 2017-12-15
Translated english sentences to dutch

TheSander562:


TheSander562:
Thanks @TortleWortle , i think melding is bericht, and i think jouw is more accurate than jou.

TortleWortle:
Not sure about the "jou" vs "jouw" thing.
But for the "melding" vs "bericht" I don't think it matters alot.

PR #941, opened 2018-01-05
Update user.php

Thesevs:
Updated some stuff.


samerton:
Thanks for the updates :)

Issue #915, opened 2017-12-13
Add email field to contact form


samerton:
Add email field to contact form so staff can respond to messages

RichardB122:
Hi, I'm interested in contributing and I have made a commit for this on my fork. Is there any place where I can find a style guide and your policies about contributing?

samerton:
Hi @RichardB122, to be honest it's very relaxed on that front, we don't have anything like that set out at the moment.
If you have any questions or anything, I can try my best to answer them.

samerton:
Added in d1289657df4e14af0b8609e66580e9a27a89f76e, thanks to @RichardB122

PR #944, opened 2018-01-09
Add email field to contact form

RichardB122:
This pull request implements an email field to the contact page #915. Validation is handled on the back end and new error messages have been created to handle an invalid email. These new fields have been added to all the language files so that they remain in sync. 

The email the user provides in the contact form is set in the "Reply To" header for emails sent through the php mail function. However, I could not figure out how to do this with the PHPMailer library so this still needs to be done. 

I have tested the front-end error handling and backend validation. However, I was not able to get email setup locally so this should still be tested.

### DIFF
--- a/custom/languages/Czech/general.php
+++ b/custom/languages/Czech/general.php
@@ -58,8 +58,10 @@ $language = array(
      */
     'contact' => 'Kontakt',
     'message' => 'Zpráva',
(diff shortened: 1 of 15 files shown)


samerton:
Thanks a lot! :)

I'll take a look into setting the reply to header, it's probably something that still needs to be implemented in the email wrapper.

PR #964, opened 2018-02-02
Added Dutch translation for Ban appeal

jesgeerts:
However I don't know how to add it into the addon as a language so I made only the translation if it..


partydragen:
Thanks for this :)

PR #971, opened 2018-02-17
Missing Icon

Tyrelis:
Fixed missing favicon image to slightly improve page load time and fix the 404 error relating to it.


samerton:
Thanks for this, however I'm not going to merge this as I think that, if we're adding a favicon, it should be the Nameless icon as we don't own the Minecraft logo.

Tyrelis:
If you take a look at minecraft.net logo and the one in the pull request it's different. Why you would you put the namelessmc logo when minecraft logo makes the most since.

samerton:
Nameless v2 has an option to entirely disable Minecraft integration, so it's not necessarily Minecraft related.

If you're able to find a license for the logo in the pull request though, I can merge it.

Tyrelis:
It comes from this website considering they allow their users to use the favicons http://www.favicon.cc/?action=icon&file_id=900533

PR #1004, opened 2018-04-17
Replace Cravatar.eu

Tyrelis:
Replace unreliable and slow avatar site cravatar.eu for minotar.net. The website causes 12+ seconds increased load time for many sites as well as images not rendering as a result. I've tested thoroughly and find everything working and improved greatly.

Link: http://prntscr.com/j68tpr


Tyrelis:
I am using cloudflare cdn with **argo** and I have the pro plan. I have a 1gbps port dedicated server. Asides testing my own website I've done tests with cravatar.eu finding it often slow or offline. Each testing site will have slightly different results but nearly the same showing slow loading results. Not only that I find that some sites have the same issue and it will sometimes load the steve image instead of the users avatar as a result.

samerton:
Originally, minotar was used however it had serious reliability issues a couple of years ago, so I switched over to cravatar.

I'll take a look into it when I get a chance, thanks for the pull request.

samerton:
Thanks, that typo has gone almost 3 years without being spotted :P

PR #1014, opened 2018-04-28
footer at bottom of page 

jer3m01:


samerton:
Thanks, I'll take a look at this when I get a chance

Derkades:
I don't know if you should be modifying the bootstrap css files. Please add your css to the custom.css files in the same folder.

PR #1039, opened 2018-06-14
Update installer.php

baxandrei:


samerton:
Thanks for the updates!

PR #1071, opened 2018-06-19
Update profile.tpl

baxandrei:
Remove class="rounded" because I do not see it useful.
Added onerror="this.style.display='none'"to hide that annoying icon when UUID is not set and the image appears broken.

with onerror: https://i28.servimg.com/u/f28/17/31/96/85/screen13.png
without onerror: https://i28.servimg.com/u/f28/17/31/96/85/screen12.png

hope to be helpful!


samerton:
Thanks

PR #1084, opened 2018-06-21
Update general.php

baxandrei:
Question: Are these variables used here or just those in users.php?


samerton:
I don't *think* these are used, and they might just be duplicates, however I'm not 100% sure.

PR #1085, opened 2018-06-21
Update register.tpl

baxandrei:
Before: https://i28.servimg.com/u/f28/17/31/96/85/screen10.jpg
aster: https://i28.servimg.com/u/f28/17/31/96/85/screen11.jpg


baxandrei:
cancel the request, the desktop does not look good now, I'll come back with another code, thank you.

PR #1086, opened 2018-06-21
Update login.tpl

baxandrei:
same as here: https://github.com/NamelessMC/Nameless/pull/1085


baxandrei:
cancel the request, the desktop does not look good now, I'll come back with another code, thank you.

samerton:
Will do, thanks anyway :)

PR #1088, opened 2018-06-22
CraftingStore: Use new V4 API with header based auth.

Tim-NL:
Updated the CraftingStore Integration to support API V4, this API uses header based auth instead of placing the API token in the URL, thus it's safer.

And changing category names resulted in double categories, it now uses name checking instead of ID checking to prevent this.


samerton:
Thanks!

PR #1103, opened 2018-06-28
Update time.php

baxandrei:


baxandrei:
Same as here: https://github.com/NamelessMC/Nameless/pull/1104 but without icons, only uppercase Î

samerton:
Thanks, I'll stick with this one and allow users to decide whether they want to add their icons via the template.

PR #1104, opened 2018-06-28
Update time.php

baxandrei:
added fa fa-clock-o


baxandrei:
if it is not good with this icon, you can close this request, but please look at this: https://github.com/NamelessMC/Nameless/pull/1103

PR #1106, opened 2018-06-29
Update custom.php

baxandrei:
I added the 'title' variable for smarty in case someone wants to use it just like me.
http://prntscr.com/k0zkue

It's there only if someone wants to use it, it does not appear anywhere.


samerton:
Thanks

PR #1109, opened 2018-06-30
Update profile.tpl

baxandrei:
You should also create a translation for this sentence: `This user has not completed any fields yet.`

I made this code for a simple reason. To add something when the user does not complete any custom fields or there is nothing in the dedicated field because it looks ugly without content.

a single field or multiple fields completed (with custom code/without custom code)
http://prntscr.com/k153vq

no completed field or empty content in the element. (with custom code)
http://prntscr.com/k1549i

no completed field or empty content in the element. (without custom code)
http://prntscr.com/k154jh

### DIFF
--- a/custom/templates/Default/profile.tpl
+++ b/custom/templates/Default/profile.tpl
@@ -256,7 +256,7 @@
 			
 			<div class="col-md-8">
 			  <div class="card">
-			    <div class="card-body">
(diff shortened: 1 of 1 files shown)


samerton:
Thanks for this, I will change it to template based only to save the Javascript :)

PR #1116, opened 2018-07-02
Update search.php

baxandrei:
I added the 'SEARCH_TERM' variable for smarty in case someone wants to use it just like me.
http://prntscr.com/k1ldra

It's there only if someone wants to use it, it does not appear anywhere.


samerton:
Thanks for this, I've made a quick edit to prevent a potential undefined index notice.

PR #1163, opened 2018-07-16
fix ssl redirect when www is on and ssl off

baxandrei:


samerton:
The line in this change needs to be `https`, as it will be executed if:

- Force HTTPS is **enabled**
- The user is currently browsing the website with **http**

Thanks though :)

PR #1173, opened 2018-07-20
Update template.php

baxandrei:


samerton:
The reason for the `else` here is that if the route is `/login`, `/register` or `/complete_signup` then the first if statement is executed. In this case, the route *cannot* be `/forum/topic/`, so we use an else to make sure the second if statement is not evaluated.

This is a slight performance increase as it saves 1 if statement check.

PR #1190, opened 2018-08-07
V2 click to copy IP

### LINKED ISSUE #1181: v2 - click to copy IP
**What kind of issue is it (Question, Bug, etc.)?** Feature request (with possible PR?).  
**In which version of NamelessMC did it occur (1.x.x)?** v2 pre 4.  
**How would you describe the issue?** Click to copy the IP in the home header.  
**Do you have (relevant) resources about this issue (Logs, pictures, etc.)?** I implemented the functionality for my test installation (see https://forum.naturamc.com).

I would love to make this a configurable option and make a PR out of it. Is there interest for that feature? I would of course try to improve the look.

NiklasEi:
I removed the html from the language files and made the tooltip it's own message. 

In my opinion, it looks the best without an additional button around it. At least it did when I tried different buttons. 

resolves #1181

### DIFF
--- a/core/init.php
+++ b/core/init.php
@@ -614,7 +614,8 @@ if($page != 'install'){
             $smarty->assign('SERVER_QUERY', $result);
 
         if(!is_null($default) && isset($default->ip)){
-            $smarty->assign('CONNECT_WITH', str_replace('{x}', $default->ip . ((!is_null($default->port) && $default->port != 25565) ? ':' . $default->port : ''), $language->get('general', 'connect_with_ip_x')));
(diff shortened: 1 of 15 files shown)


NiklasEi:
Is it critical to add the new message to all language files, or does the system fall back to a default file if a message is missing?

baxandrei:
add the new message to all language files

NiklasEi:
Is there anything else I should improve?

samerton:
Thanks for this, I'll merge and then make a small change to shift the HTML to the template :)

PR #1218, opened 2018-09-20
Fix...It corresponded to UTF-8.

Mari0914:
In the state so far, letters were wrong when sending mails in Japanese etc. I fixed it.


samerton:
Thanks for spotting this :)

PR #1231, opened 2018-09-28
a more interesting scrollbar

baxandrei:
http://prntscr.com/kzo3ms

### DIFF
--- a/custom/panel_templates/Default/assets/css/custom.css
+++ b/custom/panel_templates/Default/assets/css/custom.css
@@ -11,3 +11,19 @@
 .thumbnails li img {
 	width: 60px;
 }
+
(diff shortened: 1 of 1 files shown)


samerton:
Thanks for this.

If I'm honest, I'm not sure adding a custom scrollbar into the default template is the best idea.
For example I have a dark theme in my browser which has darker scrollbars, which this would overwrite.

Awtra:
> Thanks for this.
> 
> If I'm honest, I'm not sure adding a custom scrollbar into the default template is the best idea.
> For example I have a dark theme in my browser which has darker scrollbars, which this would overwrite.

I agree, don't think it would make much of the difference.

PR #1236, opened 2018-09-30
Japanese Language Update(New Version) Errors.php

Mari0914:


samerton:
Thanks for the updates

PR #1247, opened 2018-10-01
Update admin.php

smessie:
Translate all untranslated messages to Dutch.


samerton:
Thanks for the updates

PR #1256, opened 2018-10-02
Update admin.php

smessie:
Translated latest untranslated messages to Dutch


samerton:
Thanks for the updates

PR #1268, opened 2018-10-15
Update language.php

JosGCS:
Some messages updated (grammar issues)


samerton:
Thanks for the updates!

PR #1278, opened 2018-11-02
Update installer.view.php

baxandrei:
missing "admincp.security.all":1


samerton:
Thanks for spotting this

PR #1305, opened 2018-11-18
Update profile.tpl

baxandrei:
Shows the user's title, profile visits, date of registration and last visit in case if minecraft integration is turned off.


baxandrei:
https://github.com/NamelessMC/Nameless/commit/b9af736878a51872c75b06949cd2790890547b8e

samerton:
Thanks! I've made a slight change so it will show the default avatar if Minecraft is disabled, instead of no avatar.

PR #1312, opened 2018-11-19
Update login.php

baxandrei:
I have a normal forum (without any link to minecraft) and I wanted to put on the login page instead of minecraft nickname, username, but it is not defined.


samerton:
Thanks for letting me know about this, https://github.com/NamelessMC/Nameless/commit/b9af736878a51872c75b06949cd2790890547b8e should resolve any issues.

PR #1320, opened 2018-11-23
Added missing language terms to Lithuanian

justasofficial:


samerton:
Thanks for the update

PR #1350, opened 2018-12-21
Translated remaining parts

Thesevs:


samerton:
Thanks for the updates :)

PR #1381, opened 2019-01-25
Data and Table Relationship

Root3287:
Allowed Users to view their data. Something to comply with the EU's GDPR. These also include the DB changes. 

Changes are to add the Data page. 

== V2 DB commit details == 
Modifcaitons:
    - Change DB Innitalizaiton order to allow for table relations
   -  Table Relations
    - Change Install order for table initalization.
    - Moved group initalization for table initalization.

Anything else? I don't know... I really like to make code and not keep track of changes I've made... If anything else is broken or modified blame this commit...

== End Dublicate Commit ==

**bug:** The data tables are not responsive... I've tried wrapping the table in `<div class="table-responsive">` but it wont render the scroll bar...


Root3287:
With the DB modification you could do a ON DELETE CASCADE on the users data. So the unnecessarily data would also be removed.

Aberdeener:
Closing as it is outdated. Feel free to comment and @ me if you'd like to work on it again

PR #1470, opened 2019-03-29
Updated to v2

NPCtom:


samerton:
The htaccess in the master branch needs to stay as it is so v1 continues to function correctly :)

PR #1473, opened 2019-03-29
Update errors.php

JosGCS:


JosGCS:
Wrong translation file (real: general.php)

PR #1488, opened 2019-04-24
Fully translated admin.php

dgateles:
Completely translated to portuguese and revised older translations.


samerton:
Thanks for the updates! :)

PR #1490, opened 2019-04-24
Fully translated general.php

dgateles:
Completely translated to portuguese and revised older translations.


samerton:
Thanks again

PR #1492, opened 2019-04-29
Optimized CSS

PixelNotFound:
Optimized CSS files,
I know I know, you have to use a keybind (crtl+alt+l) to modify code, but you will get faster load times. So isnt it worth it 🤔


Derkades:
I don't think we want to minify our css and js files. At least not in the source code, maybe only for releases. If you want your webserver to automatically minify files, see this: https://www.newmediacampaigns.com/blog/Automatic-file-minification-on-nginx

partydragen:
i prefer to have it like it currently is
its cleaner and more user friendly to do changes

Everything in same line getting everything so messy

samerton:
Thanks for the PR, although I'll have to agree with Derkades and partydragen here; custom.css is relatively small too so in its non-minified form it shouldn't have much of a performance impact.

PR #1496, opened 2019-05-11
Correction on French Language

x9nico:
Hi, I've done a correction on API Language. Because it wasn't complèté but complété.

:)


samerton:
Thanks for the update :)

PR #1499, opened 2019-05-15
Fixed "View all templates" link for specifically panel templates

Aberdeener:
Noticed today that the "View all templates" link for panel templates in my StaffCP just sent me to normal templates not panel templates. Thought I'd make a pull request rather than an issue for such a simple fix.


samerton:
Thanks for spotting this!

PR #1502, opened 2019-05-21
Update german admin.php

agenthighcastle:
some fixes and translation


samerton:
Thanks for the updates!

PR #1531, opened 2019-06-14
Add support for Twitter cards

TheRealPear:
If we are going to have support for the Open Graph protocol, then why not have support for Twitter cards?


samerton:
Thanks!

PR #1532, opened 2019-06-15
master

PixelNotFound:


PixelNotFound:
Made email plural in the language/Dutch

samerton:
Thanks for the fix!

PR #1535, opened 2019-06-23
Inline item content in online staff widget

Xemah:


Derkades:
Any reason this hasn't been merged yet? @samerton

PR #1540, opened 2019-06-25
Update general.php

baxandrei:


baxandrei:
We made this PR just to see that from this commit upwards it has to be translated.

PR #1549, opened 2019-07-17
Fixed translation

Nekomitsuki:
Fixed typos and translated some sentences (lol, why this even replaced my last translation that was fully translated without any typos)


samerton:
Thanks :)

PR #1558, opened 2019-08-16
Revised user.php

dgateles:
Translation revised, corrected and translated.


samerton:
Thanks for the updates

PR #1587, opened 2019-12-08
Fix names with special characters having no avatar

Govindass:
Fix accounts with special characters in name having no avatar

### DIFF
--- a/core/classes/User.php
+++ b/core/classes/User.php
@@ -410,11 +410,12 @@ class User {
 
         if($data[0]->uuid != null && $data[0]->uuid != 'none')
             $uuid = Output::getClean($data[0]->uuid);
-        else
+        else {
             $uuid = Output::getClean($data[0]->username);
             //fix accounts with special characters in name having no avatar                                             
             if(preg_match("#[^][_A-Za-z0-9]#", $uuid)) 
                 $uuid = 'Steve';
+	}
         // Get avatar type
         if(defined('CUSTOM_AVATARS')){
             // Custom avatars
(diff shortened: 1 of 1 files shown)


samerton:
Thanks, for this PR could you wrap the lines for the `else` statement in curly braces (`{}`) like so:
```php
        else {
            $uuid = Output::getClean($data[0]->username);
            //fix accounts with special characters in name having no avatar                                             
            if(preg_match("#[^][_A-Za-z0-9]#", $uuid)) 
                $uuid = 'Steve';
        }
```

Otherwise we'll be checking a UUID for special characters :)

Also I'm wondering if there's a way to get the avatars to display correctly, rather than fall back to a Steve avatar.

Derkades:
> Also I'm wondering if there's a way to get the avatars to display correctly, rather than fall back to a Steve avatar.

I don't believe Minecraft usernames support special characters, so if they used special characters in their NamelessMC username it wouldn't be linked to a Minecraft account anyway

Govindass:
> Thanks, for this PR could you wrap the lines for the `else` statement in curly braces (`{}`) like so:
> 
> ```
>         else {
>             $uuid = Output::getClean($data[0]->username);
>             //fix accounts with special characters in name having no avatar                                             
>             if(preg_match("#[^][_A-Za-z0-9]#", $uuid)) 
>                 $uuid = 'Steve';
>         }
> ```
> 
> Otherwise we'll be checking a UUID for special characters :)
> 
> Also I'm wondering if there's a way to get the avatars to display correctly, rather than fall back to a Steve avatar.

thanks for letting me know, I didn't notice that I made a mistake

PR #1597, opened 2020-01-12
es-ES translation

Marckos200:
Translation to es-ES


samerton:
Thanks for this!

@Marckos200 just wondering, should the current Spanish translation be renamed to something more descriptive?

PR #1617, opened 2020-03-05
Simplified Chinese Translation

ahdg6:
Nearly 95% of English translations are translated into Simplified Chinese, and most forum words are localized.
We will follow up the official version as much as possible during our spare time.
Translated by ahdg,lian20,LingDong,NEWLY_1129514


samerton:
Thank you for this!

PR #1624, opened 2020-03-26
Update profile_tab.php

baxandrei:
Prevent empty forum tab. (when user have only one post, and that post is deleted).


samerton:
Thanks!

baxandrei:
You are welcome!

PR #1648, opened 2020-05-06
Create Czech.php

Fjuro:


samerton:
Thank you for the translation!

PR #1666, opened 2020-05-18
Chinese(Simplified) Translation for module forum

ahdg6:
Chinese(Simplified) Translation for module forum
Most of its words have been Localized,Translated by ahdg


samerton:
Thanks!

PR #1674, opened 2020-05-30
Remove Nameless Installer file on install (v1)

Aberdeener:


Derkades:
Can't this be done by the easy installer? Send a redirect header then destroy itself. You have made the choice to keep the easy installer separate from the main project, so in my opinion you shouldn't include code related to the easy installer in the main project.

Aberdeener:
I didn't really make a choice to make the installer separate, that is the whole purpose of it - to be a 1 file downloader. I'll do that instead though...

PR #1689, opened 2020-06-25
Add permission for users to edit their own topics

Aberdeener:
Removes confusion when users did not know they needed the "create replies" permission to edit their own topic.
*Can Moderate permission overrides this.*
I also added an updater file and updated the Queries to add this to the database upon installation.


samerton:
Just to check, is this specifically for editing topics and not replies?

Aberdeener:
Correct!

modules/Forum/pages/panel/forums.php
```diff
@@ -753,6 +759,7 @@
 									'forum_id' => $_GET['forum'],
 									'view' => $view,
 									'create_topic' => $create,
+									'edit_topic' => $create,
```
samerton, line 762: This line should be `'edit_topic' => $edit,`

Aberdeener: Fixed

PR #1703, opened 2020-07-01
Check if Discord Server ID is valid length + numeric

Aberdeener:
Seen a lot of people enter the wrong length server ID for the discord widget, and also threw in a numeric checker just in case.


modules/Core/includes/admin_widgets/discord.php
```diff
@@ -21,21 +21,60 @@
...
+            if ($validation->passed()) {
+                $discord_api_key = $_POST['discord_api_key'];
+            } else {
+                // Validation errors
+                foreach ($validation->errors() as $validation_error) {
+                    if (strpos($validation_error, 'minimum') !== false) {
+                        // x must be a minimum of y characters long
+                        switch ($validation_error) {
+                            case (strpos($validation_error, 'discord_api_key') !== false):
+                                $errors[] = str_replace('discord_api_key', 'Discord Server ID', $validation_error);
```
Derkades, line 43: Does this need translation support?

Aberdeener: I left it with no translations since other validations didn't have translations. What do you think?

modules/Core/includes/admin_widgets/discord.php
```diff
@@ -36,29 +36,13 @@
...
                     else if (strpos($validation_error, 'numeric') !== false) {
-                        // x must be a maximum of y characters long
-                        switch ($validation_error) {
-                            case (strpos($validation_error, 'discord_api_key') !== false):
-                                $errors[] = str_replace('discord_api_key', 'Discord Server ID', $validation_error);
-                                break;
-                        }
+                        $errors[] = $language->get('admin', 'discord_id_numeric');
                     }
+                    break;
```
Derkades, line 45: Shouldn't it show all errors, instead of breaking after showing the first one? Otherwise, what is the point of the for loop in the first place?

Aberdeener: Wow, how did I miss that.... Fixing

PR #1712, opened 2020-07-02
Add backend email + timezone validation

Aberdeener:
I am shocked this hasn't been found out before - Thank you KFD for finding this issue.


Aberdeener:
Uhhh seems like I messed something up during pushing

PR #1751, opened 2020-07-11
Add fully customizable announcements!

Aberdeener:
I have done a lot of testing on this, and feel quite confident about it. Would appreciate another lookover. Thanks guys

Also, I can give you owner access to my testing site if you'd like 

Options:
- Icons
- Text Colour
- Background Colour
- Module pages + built in pages
- Group control
- Closable or not


Aberdeener:
Ok, with all the queries and translations done, I think those are all my commits now!

modules/Core/module.php
```diff
@@ -791,10 +793,27 @@ public function onPageLoad($user, $pages, $cache, $smarty, $navs, $widgets, $tem
 				$navs[2]->add('groups', $language->get('admin', 'groups'), URL::build('/panel/core/groups'), 'top', null, $order, $icon);
 			}
 
+			if ($user->hasPermission('admincp.core.announcements')) {
+				if (!$cache->isCached('announcements_order')) {
+					$order = 4;
```
samerton, line 798: Does using an existing order cause issues on existing sites? As the integrations tab already has order 4, do they display in the correct order?

Aberdeener: I dont think so? I bumped the order for all other navbar items up by 1, so it should be fine I think...

samerton: @Aberdeener It won't modify the orders for existing installations, since they're already cached - just checking having 2 items in the list with the same order of 4 still results in the links displaying properly

Aberdeener: Oh I see, okay. I'll take a look after work today - thanks for the detailed review

Aberdeener: @samerton, According to the Cache class documentation, after setting the cache file and then calling eraseAll() it will erase all the key/values in that file - so my solution for this issue is just to `$cache->setCache('panel_sidebar')->eraseAll()` on upgrade from pre7 -> pre8.
If the Cache documentation is correct (which I assume it is), then this shouldnt cause any issues. I will not merge this until you confirm this is the case, as I would hate to merge damaged code. 
I searched NMC files and couldnt find any usage of eraseAll(), which is why I am hesitant.
Anyways, if this is all good, then you can merge :^)

Aberdeener: Party said this would work, so I am closing and merging

partydragen: well i only meant that eraseAll() is working, i don't know about rest of the code ¯\_(ツ)_/¯ or if samerton actually want the order changed

Aberdeener: Yeah, I know that the rest of the code works - I did lots of testing with it

PR #1782, opened 2020-07-29
Revamp Installer fix (Bug fix)?

ItzArty:
I have fixed my version of revamped installer, there wasnt a lot of work to do, really. The main issue with installer is the changes in Bootstrap, so I had to change few class names and added popper.js, which is required by now to power popups and further.


Aberdeener:
Thanks!
Has this been tested?

ItzArty:
> Thanks!
> Has this been tested?

I have not really fully tested Nameless, but the installer should work just fine, I found no problems, atlhough seems, like popper itself has some issues, but it doesnt damage installer or nameless, just sometimes opens dropdowns configured to appear at the top of the item tempt to appear down anyway :-/ Overall it has been tested

core/assets/css/custom.css
```diff
@@ -342,3 +342,6 @@ form.install > .input-group {
   padding-right: 32px;
   padding-top: 12px;
 }
+.input-group-prepend > i {
+  padding-top: 1.35vh;
```
Derkades, line 346: It looks like this should use px/em not vh, but I may be wrong. Did you check how it behaves when resizing your browser?

ItzArty: Well, I'll check it, but it should be okay, but yeah, will change it to px, it was too late, so I was little bit off :D

Aberdeener:
Nice, that is a lot better.
Two minor things: 
- The black at the bottom, I dont know if it fits. Just my opinion
- [Black at bottom sometimes has white under, sometimes not](https://imgur.com/a/impN4FC) (read the comments on the images)

ItzArty:
Okay, sorry for late responses but my family decided without me that we will visit neighbour country so I had not enough time to fix those issues.
About the dark footer, well it has been done so in the actual revamp theme for nameless so I wanted to keep the look but I can surely change that and about that white stripe that sometimes would appear, Im gonna fix that, it will be just minor javascript change, while to reveal the enitre footer, well, that depends on the size of the actual content, if content is not that big you will see the stripe and if it is bigger, then you gotta scroll, so this is how it is, if those two things are required then let me know, but for now I will stick to fixing the bottom white stripe

Aberdeener:
No worries, thank you for the update

Derkades:
I personally like the dark footer (if it's actually stuck to the bottom of the page :P)

ItzArty:
Yep, on my way to fix that bug, soon I will update it

StormyIceLeopard:
any updates on this bug?

ItzArty:
Oh my. I just quite forgot about this, since I have been  on vacation in
england, I thought I have fixed it already

StormyIceLeopard:
Didn't seem finished so I thought I would bump it in case it was forgotten.

ItzArty:
Ofcourse, thanks for reminding me, in like 10 hours I will fix things, I
think I wasn't able to force bootstrap to display the language dropup to
appear up, since I was offline the entire time tho

Derkades:
Xemah is improving the installer, you don't need to work on this anymore.
image

ItzArty:
Well to be honest I think Xemah suits better for this, I wasn't able to fix
even that thing so, yeaah

PR #1784, opened 2020-07-29
Update README.md

zJerino:


Aberdeener:
Not yet

Derkades:
This is somewhat complicated. The readme file is in the v2 development branch with upcoming pr8 code. While it's not yet pr8, it's definitely not pr7 either. On the other hand, some people might see the readme file as a general repository description (not a description for the latest code at that time), in which case pr7 would be better.

maybe just remove the version from the title altogether..

PR #1791, opened 2020-08-01
Update moderator.php

WhenUHackUNoob:
Just made things a little easier to understand. (Should be no spelling mistakes)


Derkades:
Usually these messages are written with brevity in mind, are you sure making them longer is a good idea?

WhenUHackUNoob:
Hm very true, I'll review the file again and change some stuff over.

WhenUHackUNoob:
Well I'll see tomorrow, closing.

PR #1814, opened 2020-08-25
Update for german translation (+ note)

Boy132:
The german translation is very inconsistent and sometimes very poorly translated!


Derkades:
We depend on the community for translations. Thanks!

PR #1834, opened 2020-09-13
Update admin.php

kacperleague9:
Damn this is so hard, my finger hurts, I will complete tomorrow


Aberdeener:
Hahaha, thank you so much

PR #1835, opened 2020-09-14
Change Invalid API Method message to be more clear

Aberdeener:
Making PR so Sam can have the final say


samerton:
Not sure on altering the message, doesn't feel very API like.
Maybe we can be clever and detect if we're accessing via a browser and display an info page instead

Aberdeener:
Good thought, I'll see about doing that in the rewrite. thanks

PR #1871, opened 2020-10-16
Update Chinese (Simplified) translations

Dreta:


Aberdeener:
Great work as always, thanks buddy

PR #1887, opened 2020-10-25
Installer defaults from environment variables

### LINKED ISSUE #1886: Define installer defaults using environment variables
For example
```sh
NAMELESS_HOSTNAME="example.com"
NAMELESS_FRIENDLY="true"
NAMELESS_DATABASE_USERNAME="root"
NAMELESS_DATABASE_ADDRESS="localhost"
```
https://www.php.net/manual/en/function.getenv.php

This would be nice for docker users since container configuration in docker is done primarily using environment variables

Derkades:
Speeds up installation for development environments (see docker-compose file) and my hosting website.

Closes #1886

### DIFF
--- a/core/installation/views/steps/admin_account_setup.php
+++ b/core/installation/views/steps/admin_account_setup.php
@@ -52,7 +52,7 @@ if ($_SERVER['REQUEST_METHOD'] == 'POST') {
 				$error = $language['email_invalid'];
 			}
 		}
-	
+
 	} else {
 		$user = new User();
 		$password = password_hash(Input::get('password'), PASSWORD_BCRYPT, array('cost' => 13));
@@ -87,9 +87,9 @@ if ($_SERVER['REQUEST_METHOD'] == 'POST') {
 				Redirect::to('?step=conversion');
 				die();
 			}
-			
+
 			$error = $language['unable_to_login'];
-			
+
 			$queries->delete('users', array('id', '=', 1));
 
 		} catch(Exception $e) {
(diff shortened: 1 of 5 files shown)


core/installation/views/steps/admin_account_setup.php
```diff
@@ -118,8 +118,8 @@
 				<div class="sixteen wide mobile twelve wide tablet ten wide computer column">
 					<div class="ui form">
 						<?php
-							create_field('text', $language['username'], 'username', 'inputUsername');
-							create_field('email', $language['email_address'], 'email', 'inputEmail');
+							create_field('text', $language['username'], 'username', 'inputUsername', getenv('NAMELESS_ADMIN_USERNAME') ?: '');
```
samerton, line 121: Apparently `getenv` returns false if the variable isn't set - do you know if this results in the string "false" being added to the field without the environment variable?

Derkades: `getenv('NAMELESS_ADMIN_USERNAME') ?: ''` is short for `getenv('NAMELESS_ADMIN_USERNAME') ? getenv('NAMELESS_ADMIN_USERNAME') : '' `. I'm pretty sure the string `"false"` evaluates to the boolean `false` in php?

samerton: Sorry I didn't see `?:`, oops 😄

Derkades:
"documentation" here: https://github.com/NamelessMC/Nameless/wiki/Environment-variables

PR #1891, opened 2020-10-26
Delete old Default template

### LINKED ISSUE #1785: Retire old "Default" template
The "DefaultRevamp" template looks much better than the old "Default" template, I don't think anyone still uses the old theme (maybe do a poll in discord?).

I propose removing the "Default" template and renaming "DefaultRevamp" to "Default". This saves time during development because there is no longer the need to add every feature to two templates.

Derkades:
The time has come.

Closes #1785

TODO: Warn Default template users https://github.com/NamelessMC/Nameless/issues/1785#issuecomment-709777498


Aberdeener:
Okay, good to go. Will let Sam make the final choice here

PR #1893, opened 2020-10-28
Fix banner image path on install

Xemah:


core/installation/views/includes/site_initialize.php
```diff
@@ -670,7 +670,8 @@
 	'Statistics' => 1
 ));
 
-$config_path = Config::get('core/path');
+$config_path = $conf['core']['path'];
```
samerton, line 673: This should be equivalent to `Config::get('core/path')` - does the existing code not work?

Xemah: O yeah, the problem was just that there wasn't a slash before the path.

PR #1898, opened 2020-10-31
Update forum module japanese translation

ViaSnake:
Japanese strings added in v2 by other contributors were not well translated.
Therefore, all strings added after September 20, 2018 have been re-translated.

Changed translator credits to the list.

*Update*
- `modules/Forum/language/Japanese/forum.php`


Aberdeener:
Thank you

Issue #1786, opened 2020-07-29
Code style guide


Derkades:
I'll edit this issue description with any conclusions drawn in the comments, when it seems like we have defined everything it can be put in a markdown file in the repo or on the wiki.

Default phpcs settings with the following differences:

- Use keywords like functions (e.g. `echo("Something")`)
- 120 character line limit instead of 85

How to use on vscode on ubuntu:
```
apt install php-fpm php-codesniffer
```
Install `phpcs` extension in vscode

Derkades:
Most important right now: tabs or spaces? how many spaces? possibly different number of spaces for HTML and PHP? Should code in `<?php` blocks be indented? How about css in `<style>`?

Xemah:
I personally use tabs, and I don't indent the code in `<?php` block

samerton:
I'd say 2 space indentations, like so:

```php
<!doctype html>
<html>
  <head>
    <title>Test</title>
    <style>
      body {
        background-color: #000;
        color: #fff;
      }
    </style>
  </head>

  <body>
    <?php
    foreach($var as $item){
      echo $item;
    }
    ?>
  </body>
</html>
```

Derkades:
phpcs suggests no parentheses for statements (so `require 'file.php'` not `require('file.php')`, what do you think?

partydragen:
well require('file.php') looks nicer xD

samerton:
Agree with partydragen, for me require() looks nicer

RE https://github.com/NamelessMC/Nameless/commit/7e0e98862a6fa95a2219313a83834df06790bca6#diff-59736aecc611cda3c8a10fbb4681b15045c56f8c2a786bd3c154dacdcb03abbb

Because HTML can end up with some deep indents, can we keep HTML as 2 space indentations, also indent <head> after <html>

Derkades:
> Because HTML can end up with some deep indents, can we keep HTML as 2 space indentations

I would do this for HTML templates that are separated from PHP code, but in this case one file had both and IDEs only support one indentation type per file. It can be done manually if you really want to though.

PR #1899, opened 2020-11-01
Use 4 spaces for indent in languages

ViaSnake:
The indentation of the language file contains a mix of tabs and spaces.
This is inconvenient and confusing when translating, so I replace everything with 4 spaces.

Perhaps the mix of tabs and spaces was causing some of the indentations in some language files to go wrong. I fix that too.

*Change*
`modules/Forum/language/.*/forum.php`
`custom/languages/.*/.*.php`


Derkades:
Thank you for this, I've been wanting to so it for a long time but forgot

ViaSnake:
`Nameless/custom/languages/*` also needs to be unified into two spaces from the tab? If so, I will add commit to this PR.

Derkades:
Can you format it it to 4 spaces for indentation? That seems to be the standard for all other php files. Unfortunately we have no clear style guidelines yet, #1786

ViaSnake:
Ok. Replaced the tab in all language files with 4 spaces.

Derkades:
I'll leave the final decision to @samerton
it should be merged as soon as possible though to prevent conflicts

Derkades:
@samerton conflicts are already accumulating, please have a look at this PR

PR #1908, opened 2020-11-07
Browser-based spell checking for TinyMCE

Marckos200:
This activates the spell checker and changes the contextmenu of the editor to the browser contextmenu, allowing the words to be corrected with the browser's own dictionary.


Derkades:
It might be my browser but it seems to work
image

samerton:
A recent update to the editor in Nameless lost the pre-7 configuration we had - I'll be reverting it back to the pre-7 configuration for the next update

PR #1912, opened 2020-11-13
Make bold number consistent in player online

Govindass:
I noticed a really weird thing today - the server online player count is bold only when there's more than 1 player online, but when it's only 1 online, it's not bold, so I thought it'll be better to make it bold.


Derkades:
Thank you for your contribution. To avoid inconsistencies, would you mind updating this for all language files?

Govindass:
> 
> 
> Thank you for your contribution. To avoid inconsistencies, would you mind updating this for all language files?

done

Aberdeener:
Thanks!

Derkades:
A few languages are still missing, I can add those myself though

PR #1985, opened 2020-12-26
Translate untranslated strings

Fjuro:
fjuro is dumb and forgot to translate new strings 🤦


Derkades:
Thank you! Fortunately pr9 won't be that long, so you should see your translations soon.

PR #1990, opened 2020-12-26
Update user.php

Fjuro:
Bad translation


Fjuro:
got more translations

PR #1991, opened 2020-12-26
Update translations

Fjuro:
Found bad translations, sorry for two PRs today


Aberdeener:
Thanks as always, Fjuro :)

PR #2046, opened 2021-01-08
Completely rewritten translations

Fjuro:
Made from 0, because the old translations were very inconsistent


Derkades:
Thank you for this!

PR #2102, opened 2021-02-13
Added function to send alert without specific language file

Vladimir-Urik:
I added a function to create an alert without a language file.

### DIFF
--- a/core/classes/Alert.php
+++ b/core/classes/Alert.php
@@ -37,6 +37,21 @@ class Alert {
             }
         }
     }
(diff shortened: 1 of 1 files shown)


Derkades:
What are possible use cases for this method? Where do you use it?

Aberdeener:
Code is fine but as Derek said, what is the purpose?
Did you encounter an error at one point which this would fix?

Aberdeener:
Closed for inactivity

PR #2103, opened 2021-02-13
Fixed a bug that allowed to 2 times close/open report.

Vladimir-Urik:
Fixed a bug that allowed to 2 times close/open report.


modules/Core/pages/panel/users_reports.php
```diff
@@ -285,24 +286,29 @@
...
+                            'commenter_id' => $user->data()->id,
+                            'comment_date' => date('Y-m-d H:i:s'),
+                            'date' => date('U'),
+                            'comment_content' => str_replace('{x}', Output::getClean($user->data()->username), $language->get('moderator', 'x_closed_report'))
+                        ));
+                        Session::flash('report_success', $language->get('moderator', 'report_closed'));
+                        Redirect::to(URL::build('/panel/users/reports/', 'id=' . Output::getClean($report[0]->id)));
+                    } else {
+
+                        Session::flash('report_failure', "You cannot close this report 2x");
```
Derkades, line 308: Could you add a translation for this, just like the other Session::flash lines?

Vladimir-Urik: > Could you add a translation for this, just like the other Session::flash lines?

Done.

PS: Sorry for a large number of commits 😅

Derkades: don't worry, fortunately github has a squash option

Aberdeener:
Thanks for the PR - How was this bug created or replicated in the first place?

Derkades:
Regardless of whether this is a bug that can be reproduced using the UI, it's probably a good idea to prevent creating a new report when there already is one, for example if a malicious user decides to spam POST requests.

Aberdeener:
Closing for inactivity

PR #2193, opened 2021-03-06
design modern

zJerino:
screenshot


Derkades:
For reference, the default design:
image

Derkades:
Maybe remove "Value" in the right column? Right now the information density is a bit low

Aberdeener:
Yeah, if the "Value" is removed on right hand side I'm fine with this

PR #2194, opened 2021-03-07
Error handling revamp

Aberdeener:
Screen Shot 2021-03-06 at 4 59 12 PM

-Clickable frames for each file and line that was called before the error

Currently doesn't catch all errors and if an error happens before the class is loaded bad things happen. I think it'll need a revamp to catch everything.

Still lots to do in terms of QOL, design and general impl.
Would appreciate reviews, dont wanna keep working if there is anything fatal anyone else disagrees with


Aberdeener:
Updated style: (obviously scroll down to see the highlighted line of code + click on other frames)
Screen Shot 2021-03-08 at 10 35 30 PM

Aberdeener:
Marking as ready for review, I can finish the details this evening before merging.

core/init.php
```diff
@@ -11,51 +11,55 @@
 
 // Nameless error handling
 require_once(join(DIRECTORY_SEPARATOR, array(ROOT_PATH, 'core', 'classes', 'ErrorHandler.php')));
-set_error_handler("ErrorHandler::catchError");
-register_shutdown_function("ErrorHandler::catchFatalError");
+set_exception_handler([ErrorHandler::class, 'catchException']);
```
samerton, line 14: Does this replace the need for `register_shutdown_function`?

core/init.php
```diff
@@ -11,51 +11,55 @@
...
+// Any errors thrown now will not be caught by the error handler, as they require some classes be preloaded
 spl_autoload_register(function ($class) {
     $path = join(DIRECTORY_SEPARATOR, array(ROOT_PATH, 'core', 'classes', $class . '.php'));
-    if (file_exists($path)) require_once($path);
+    if (file_exists($path)) {
+        require_once($path);
+    }
 });
-
+
```
samerton, line 62: Extra spacing 😄

core/init.php
```diff
@@ -65,8 +69,8 @@
 
 if ($page != 'install') {
     /*
-     *  Initialise
-     */
+    *  Initialise
```
samerton, line 72: Can we keep the asterisks inline? They look neater imo

core/init.php
```diff
@@ -157,9 +163,9 @@
             ini_set('display_errors', 0);
         }
     }
-	
-	// Configurations
-	$configuration = new Configuration($cache);
+
```
samerton, line 166: Extra space

modules/Core/init.php
```diff
@@ -19,4 +19,7 @@
 
 require_once(ROOT_PATH . '/modules/Core/module.php');
 
+// TODO: start_line is negative when issue LOC is less than line_buffer
```
samerton, line 22: Is this still needed?

samerton (changes requested):

core/init.php
Aberdeener: Good catch, must have been a weird copy/paste issue 😄

core/assets/css/prism.css
```diff
@@ -0,0 +1,278 @@
+/* PrismJS 1.23.0
```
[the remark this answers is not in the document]
Aberdeener: Thanks, fixed

core/init.php
Aberdeener: Good point, I misunderstood the documentation. 
I have added it back with a new function.

Aberdeener:
Good to go for me now. See #development channel for some notes on latest commit. Thanks!

PR #2200, opened 2021-03-12
Rework captcha system and add Recaptcha 3

### LINKED ISSUE #2080: Update from reCAPTCHA v2 to v3

samerton:
Closes #2080

### DIFF
--- a/core/classes/CaptchaBase.php
+++ b/core/classes/CaptchaBase.php
@@ -19,7 +19,7 @@ abstract class CaptchaBase {
      * @param CaptchaBase $provider Provider instance to register
      */
     public static function addProvider(CaptchaBase $provider) {
-        CaptchaBase::$_providers[$provider->_name] = $provider;
+        self::$_providers[$provider->_name] = $provider;
     }
 
     /**
--- a/core/classes/CaptchaBase.php
+++ b/core/classes/CaptchaBase.php
@@ -1,0 +1,98 @@
+<?php
+/*
+ *	Made by Samerton
+ *
+ *  https://github.com/NamelessMC/Nameless/
+ *  NamelessMC version 2.0.0-pr10
+ *
+ *  License: MIT
+ *
+ *  CaptchaBase class
+ */
+abstract class CaptchaBase {
+    protected static $_providers = array(), $_activeProvider = '';
+
+    protected $_name, $_publicKey, $_privateKey;
+
+    /**
+     * Register a provider
+     * @param CaptchaBase $provider Provider instance to register
+     */
+    public static function addProvider(CaptchaBase $provider) {
+        CaptchaBase::$_providers[$provider->_name] = $provider;
+    }
+
+    /**
+     * Get provider name
+     * @return string Provider name
+     */
+    public function getName() {
+        return $this->_name;
+    }
+
+    /**
+     * Get public key
+     * @return string Provider name
+     */
+    public function getPublicKey() {
+        return $this->_publicKey;
+    }
+
+    /**
+     * Get private key
+     * @return string Provider name
+     */
+    public function getPrivateKey() {
+        return $this->_privateKey;
+    }
+
+    /**
+     * Set active provider
+     * @param string $provider Provider name to set as active
+     */
+    public static function setActiveProvider($provider) {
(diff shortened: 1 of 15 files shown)


samerton:
> Might be missing something, but why do we call addProvider($this) here as well as in the Core module init file with addProvider()? Isnt this redundant?

Good spot, thanks 😄

core/classes/CaptchaBase.php
```diff
@@ -0,0 +1,98 @@
...
+    protected static $_providers = array(), $_activeProvider = '';
+
+    protected $_name, $_publicKey, $_privateKey;
+
+    /**
+     * Register a provider
+     * @param CaptchaBase $provider Provider instance to register
+     */
+    public static function addProvider(CaptchaBase $provider) {
+        CaptchaBase::$_providers[$provider->_name] = $provider;
```
Aberdeener, line 22: Should we use `self` here for consistency with the rest of the class?

modules/Core/classes/Recaptcha2.php
```diff
@@ -0,0 +1,44 @@
+<?php
+class Recaptcha2 extends CaptchaBase {
```
Aberdeener, line 2: Missing title comment, not sure if this is intentional

modules/Core/classes/Recaptcha3.php
```diff
@@ -0,0 +1,51 @@
+<?php
+class Recaptcha3 extends CaptchaBase {
```
Aberdeener, line 2: Title comment

modules/Core/pages/login.php
```diff
@@ -29,44 +29,22 @@
...
-			curl_setopt($ch, CURLOPT_POST, 1);
-			curl_setopt($ch, CURLOPT_POSTFIELDS, $post_data);
-			curl_setopt($ch, CURLOPT_HEADER, 0);
-			curl_setopt($ch, CURLOPT_RETURNTRANSFER, 1);
-
-			$result = curl_exec($ch);
-
-			$result = json_decode($result, true);
+		if (!isset($_SESSION['tfa']) && $captcha == 'true') {
+            $captcha_passed = CaptchaBase::getActiveProvider()->validateToken($_POST);
```
Aberdeener, line 42: Not sure if this is just GH preview, but line 42 appears to be not indented properly.

Aberdeener (approved):

modules/Core/pages/login.php
Derkades: it's because the file contains tabs, which many editors display as 4 spaces but github displays as 8

modules/Core/pages/register.php
```diff
@@ -593,26 +565,22 @@
...
-        );
-    } else {
-        $template->addJSFiles(
-            array(
-                'https://www.google.com/recaptcha/api.js' => array()
-            )
-        );
+if ($captcha === 'true') {
+    $smarty->assign('CAPTCHA', CaptchaBase::getActiveProvider()->getHtml());
+    $template->addJSFiles(array(CaptchaBase::getActiveProvider()->getJavascriptSource() => array()));
```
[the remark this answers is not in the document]
samerton: It's part of the `addJSFIles()` and `addCSSFiles()` methods in the template to add attributes such as integrity, crossorigin etc: https://github.com/NamelessMC/Nameless/blob/v2/core/classes/TemplateBase.php#L31

core/classes/CaptchaBase.php
samerton: Yep good catch, too used to TS... 😄 

In terms of the blank line, personally I'm not a fan of blank lines under class/method declarations - I just feel like they're unnecessary since the declaration is already pretty short and unobtrusive.

modules/Core/classes/Recaptcha2.php
samerton: It was intentional, I'll add them in though for consistency

modules/Core/pages/login.php
samerton: For now I'll change line 42 to tabs too, until we run a formatter through the whole project to convert to spaces

PR #2203, opened 2021-03-15
Update Discord role after verification

### LINKED ISSUE #2187: Trigger discord role sync after linking
After a user links their discord account to the website, use the `/roleChange` bot endpoint to send the website user's current roles to discord

PikaMug:
This attempts to resolve #2187 and would be important for servers which require verification in order to attain a higher Discord role.

It should be noted that I have zero PHP experience, no way to test this change, and have merely copied part of the function from https://github.com/NamelessMC/Nameless/blob/103cab4e01e27f2de061f2e0d6cdc357fcbaca79/modules/Core/includes/endpoints/AddGroupsEndpoint.php


Aberdeener:
Would be nice to get some testing done before merging. Is anyone available to do so?

PikaMug:
Am finally capable of testing and no, this doesn't work.

PR #2207, opened 2021-03-16
Validation revamp

Aberdeener:
Will modernize the Validate class and handling of messages and errors.

Example usage (at time of writing):
image


Aberdeener:
Forgot to checkout to `master` first. Redoing

PR #2208, opened 2021-03-16
Validation Revamp

Aberdeener:
Modernize and streamline validation class, while **keeping it 100% backwards compatible!**
Will let us do fun things like this:
image


Aberdeener:
Valid usage:
```php
$validate->message(
    // Fallback message for any errors that occur AND that do not have specific messages set
    'Invalid input. Please ensure you have entered a correct label name, forums and type.'
)->messages([
    'label_id' => [
        Validate::REQUIRED => 'Please select a label type' // this will override the message() above
    ],
    'label_forums' => 'Please select at least one forum for this label'
])->check($_POST, [
    'label_name' => [
        'required' => true, // 'required' == Validate::REQUIRED. constants are just syntax sugar
        Validate::MIN => 1,
        Validate::MAX => 32
    ],
    'label_id' => [
        Validate::REQUIRED => true
    ],
    'label_forums' => [
        Validate::REQUIRED => true
    ]
]);
```

Aberdeener:
Converted all but 2 validators to the new style. Will review tomorrow and finish the last two ASAP

core/classes/Token.php
```diff
@@ -36,7 +36,10 @@ public static function get() {
     // Check a token in session matches
     // Params: $token (string) - contains the form token which will be checked against the session variable
     public static function check($token = null) {
-        if ($token == null) $token = Input::get('token');
```
BaxAndrei, line 39: I don't think it was necessary to add {}.

Derkades: https://stackoverflow.com/questions/359732/why-is-it-considered-a-bad-practice-to-omit-curly-braces

PR #2220, opened 2021-03-28
Add page with NamelessMC and third-party licenses

Derkades:
To make sure we comply with the licenses of libraries, like mobile apps with a "legal" section listing all projects they use and their licenses.

This is the first time I'm creating a new page in NamelessMC so feedback is very welcome!


Derkades:
Screenshots
image
Mobile
image

samerton:
Could we review the libraries used to ensure this page is necessary? MIT for example doesn't require a page like this

PR #2238, opened 2021-04-26
Updated some license information to please LibreJS

Jochyoua:
Added some magnet links to identify the correct license for LibreJS to understand
[Setting Your JavaScript Free](https://www.gnu.org/software/librejs/free-your-javascript.html)


custom/panel_templates/Default/assets/js/dataTables.bootstrap4.min.js
```diff
@@ -1,3 +1,4 @@
+// @license magnet:?xt=urn:btih:d3d9a9a6595521f9666a5e94cc830dab83b65699&dn=expat.txt Expat/MIT
```
Derkades, line 1: Thanks for this! Are you sure this is datatables version >1.10? Apparently older versions are licensed under GPL/BSD. All other licenses look fine.

Jochyoua: After some digging it appears the version of dataTables.bootstrap4.min.js is 1.10.19 as this [file hosted on their CDN](https://cdn.datatables.net/1.10.19/js/dataTables.bootstrap4.min.js) is identical to this one without any changes

Issue #2234, opened 2021-04-22
Issue with forum index


Jochyoua:
**Describe the issue**
On the Forums index page, attempting to click on the latest post takes you to a page labeled as page 0
I assume this is because I have deleted the first and last post on this topic

**To Reproduce**
Steps to reproduce the behaviour:
1. Create a post
2. Be the first comment on post
3. Remove the comment
4. Behold the page 0, screenshot below

**Expected behaviour**
I expected it to take me to the latest post instead of page 0

**Screenshots**
![Image of error](https://i.fairy.land/HU6LPJ.png)

**NamelessMC version (from AdminCP -> Overview)**
2.0.0-pr9
**Additional information**
At first I thought this was the fault of a template but the issue persists across all my tested templates

PR #2239, opened 2021-04-27
Fixed issue with pagination

Jochyoua:
This fixes a bug I came across a while ago. 
If a reply was deleted and it was the last reply it would show a strange page, this fixes that issue


Aberdeener:
Thanks, you're saying this fixed #2234 for you?

Jochyoua:
Yes I'm glad I found a fix because it was bugging me for a while

PR #2241, opened 2021-04-29
Added feature of receiving timezone on registration

Jochyoua:


custom/templates/DefaultRevamp/js/core/core.js
```diff
@@ -83,6 +83,13 @@ $(function() {
         delay: {show: 500, hide: 200},
         onShow: function(e) {this.html(cachedUsers[$(e).data('poload')].html)}
     });
+
+		var timezone = document.getElementById('timezone');
```
Aberdeener, line 87: can you use `const` instead of `var` here?

modules/Core/pages/register.php
```diff
@@ -353,6 +353,7 @@
                                     'last_online' => $date,
                                     'language_id' => $language_id,
                                     'active' => $active
+                                    'timezone' => ((isset($_POST['timezone']) && $_POST['timezone']) ? Output::getClean($_POST['timezone']) : Output::getClean('Europe/London'))
```
Aberdeener, line 356: instead of hardcoding `Europe/London`, we have a const of `TIMEZONE` you can replace it with.

Aberdeener, line 356: as well, on line ~173 in this file, can you add timezone to the validation array?
```php
'timezone' => [
	Validate::TIMEZONE => true
]
```

Aberdeener (changes requested):

PR #2261, opened 2021-05-08
Revamp/cleanup (external) avatar handling

Aberdeener:
This pr will convert avatar handling from its current state into a flexible system which is extremely easy to use and add more sources to.

### DIFF
--- a/core/classes/User.php
+++ b/core/classes/User.php
@@ -404,127 +404,56 @@ class User {
         return $this->data()->signature;
     }
 
-    // Get a user's avatar, based on user ID    
     /**
      * Get this user's avatar.
      *
-     * @param mixed $path TODO: NOT USED. REMOVE.
      * @param int $size Size of image to render in pixels.
-     * @param bool $full Whether to use full URL or not.
+     * @param bool $full Whether to use full site URL or not, for external loading - ie discord webhooks.
      * @return string URL to their avatar image.
      */
-    public function getAvatar($path = null, $size = 128, $full = false) {
-        $data = $this->data();
-        if (empty($data)) {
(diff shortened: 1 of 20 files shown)


core/classes/User.php
```diff
@@ -404,127 +404,56 @@ public function getSignature() {
...
+            $uuid = $this->data()->uuid;
+        } else {
+            $uuid = $this->data()->username;
+            // Fallback to steve avatar if they have an invalid uuid
+            if (preg_match("#[^][_A-Za-z0-9]#", $uuid)) {
+                $uuid = 'Steve';
             }
         }
+
+        return AvatarSource::getAvatarFromUUID($this->data()->uuid, $size);
```
samerton, line 456: Should the first parameter be `$uuid` instead of pulling it from data?

custom/panel_templates/Default/core/avatars.tpl
```diff
@@ -155,6 +152,45 @@
     <!-- End Wrapper -->
 </div>
 
+<script>
+
+const perspective_selector = document.getElementById('inputAvatarPerspective');
```
samerton, line 157: Just a side note, should we officially drop support for IE and show some form of banner to IE users, asking them to switch browsers?

core/classes/User.php
Aberdeener: good catch

custom/panel_templates/Default/core/avatars.tpl
Aberdeener: yeah, i agree. i'll make an issue now so we can remember that

PR #2284, opened 2021-05-23
Update japanese translation

ViaSnake:
Update japanese translation.

### Update
```
custom/languages/Japanese/admin.php
```


Derkades:
Thanks!

PR #2313, opened 2021-06-29
Fix upgrade script by removing switch

smessie:
I've spent hours and hours into looking what I was doing wrong till I take a look into the code and saw this potential issue here.
This code uses a variable `$s` which is uninitialised. As a consequence of this, the switch below will only execute the first case as `$s` will be 0 by default.
This causes the upgrading script to only convert the first batch of tables leaving all others empty which was exactly my issue.

In this PR I have removed this (possibly left over?) variable `$s` and its associated switch, the case bodies are placed one after the other so that all tables are converted.

### DIFF
--- a/core/installation/views/includes/upgrade_perform.php
+++ b/core/installation/views/includes/upgrade_perform.php
@@ -1,1114 +1,1083 @@
 <?php
 
-if ($s < 9) {
-    $conn = DB_Custom::getInstance($_SESSION['db_address'], $_SESSION['db_name'], $_SESSION['db_username'], $_SESSION['db_password'], $_SESSION['db_port']);
-}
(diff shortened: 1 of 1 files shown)


partydragen:
With the current information we gonna fix the switch, not remove it

samerton:
@partydragen is this something you've looked into?

partydragen:
> @partydragen is this something you've looked into?

Not much, No idea how we gonna do it better you do it i still need to learn more about ajax/js etc must run the steps links somehow

Aberdeener:
@partydragen should this be closed/set as draft for the time being?

partydragen:
> @partydragen should this be closed/set as draft for the time being?

Believe we can close it then have a issue for it?

PR #2316, opened 2021-07-01
Change DB password field to password type

smessie:
In this step, the password was displayed clear text on the screen.
In this PR I change the field type to password so that the input is hidden.


Aberdeener:
I seem to remember there being a good reason for this. @Xemah any ideas why?

Xemah:
I'm not sure ):

PR #2317, opened 2021-07-02
Update Slovak translation

RobiNN1:
Synchronized with EnglishUK using Crowdin.


Derkades:
thanks for translating, and especially finding missing strings and removing some old strings!

RobiNN1:
With Crowdin it's so easy

PR #2333, opened 2021-07-08
Fix update checking/upgrading to work with new API

Aberdeener:
Based on https://github.com/Aberdeener/Nameless-Releases


Aberdeener:
Will rewrite

Issue #2328, opened 2021-07-08
Shrinked panel navbar no scroll


skyrowl:
**Describe the issue**
The shrinked panel navbar is too long, don't have scrollbar and can't be unshrinked as it has no scrollbar to go on the bottom

**To Reproduce**
Steps to reproduce the behaviour:
1. Go to AdminCP
2. Shrink navbar on screens that don't show the bottom

**Expected behaviour**
Having a scrollbar

**Screenshots**
(The screenshot is full height of the page)
![image](https://user-images.githubusercontent.com/66191586/124950833-0fdbde00-e013-11eb-9baf-f242e0ee6448.png)

**NamelessMC version (from AdminCP -> Overview)**
v2-pr10

PR #2336, opened 2021-07-11
possible fix to #2328

ItsLynix:
Possible temporary fix to #2328
If there's any way to align the icons and text in the overflow  too, please make that edit.

### DIFF
--- a/custom/panel_templates/Default/assets/css/custom.css
+++ b/custom/panel_templates/Default/assets/css/custom.css
@@ -314,6 +314,7 @@ i,
     height: 100vh;
     z-index: 999;
     overflow-x: hidden;
+    overflow-y: auto;
 }
 
 .sidebar::-webkit-scrollbar-track {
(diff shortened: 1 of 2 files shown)


Derkades:
CSS should probably be done here: https://github.com/NamelessMC/Nameless/blob/c83b85108a938d3fda022721659e362ec046d622/custom/panel_templates/Default/assets/css/custom.css#L312

You can see `overflow-x: hidden` is already present there. Is there a way to add the scrollbar only when necessary? Your commit is essentially the reverse of 0fee1cdb9

ItsLynix:
I agree that there would be a big ugly white bar, overflow-y: auto; would likely help as it would only appear when it actually needs to. 

Is it just me or does Firefox completely ignore customized scrollbars?

ItsLynix:
I just tested overflow-y locally and it worked fine.

Derkades (approved):

PR #2356, opened 2021-08-03
Set Custom Pages url and title length to 255

### LINKED ISSUE #1680: Support for longer titles and paths
As the title suggests, I want to set the page title and page path to be longer than they are now.

Currently the page title is limited to 30 characters and the path is limited to 20 characters.

ViaSnake:
Allows custom pages to be created with longer title and path.

- Close #1680 

Change the "url" and "title" of "nl2_custom_pages" from varchar(20) to varchar(255).
Change `page_title_maximum_30` -> `page_title_maximum_255`
Change `page_url_maximum_20` -> `page_url_maximum_255`

I didn't know the optimal maximum length, so I set it to 255 for now. If there is a length that is most suitable, it is better to change it to that.


samerton:
Thanks for this

PR #2357, opened 2021-08-03
Display non-alphanumeric characters in forum URLs

### LINKED ISSUE #1683: When topic title is Japanese, it does not show in the title and URL
**Describe the issue**
When write a forum topic title in Japanese, friendly URL don't show the title, nor do they show the page title.

**To Reproduce**
Steps to reproduce the behaviour:
1. Create "あいうえお" topic.
2. Post.
3. See the URL and page title.

**Expected behaviour**
If the friendly URL, need to encode the Japanese title
For example, if the topic title is `あいうえお`, the URL should be `%E3%81%82%E3%81%84%E3%81%86%E3%81%88%E3%81%8A`.
The page title must display the topic title.

**Screenshots**
title

**NamelessMC version (from AdminCP -> Overview)**
2.0.0-pr7

ViaSnake:
Topic URLs will now contain non-alphanumeric characters. For example, the [CJKV characters](https://en.wikipedia.org/wiki/CJK_characters).

- Fix #1683

image

In a past commit(https://github.com/NamelessMC/Nameless/commit/a50fb572181675f0686a54ad94a4e873e65889b4), non-alphanumeric characters were removed from the forum URL. I don't know the reason for this change, but if some reason exists, need to consider changing this Pull Request.

### DIFF
--- a/modules/Forum/classes/Forum.php
+++ b/modules/Forum/classes/Forum.php
@@ -460,7 +460,7 @@ class Forum {
     // Transform a topic title to URL-ify it
     public function titleToURL($topic = null) {
         if ($topic) {
-            $topic = preg_replace("/[^A-Za-z0-9 ]/", '', Util::cyrillicToLatin($topic));
+            $topic = Util::cyrillicToLatin($topic);
             return Output::getClean(strtolower(urlencode(str_replace(' ', '-', htmlspecialchars_decode($topic)))));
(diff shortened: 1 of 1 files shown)


samerton:
Thanks for this, I can't remember why they were removed so we'll get this into the next release and monitor for any potential incompatibilities

PR #2392, opened 2021-08-30
Update japanese translation

ViaSnake:
All Japanese translations have been updated to NamelessMC 2.0.0-pr11.

The Japanese translation of installer.php was very old, so I retranslated everything.
And fixed some orthographic variants.


Derkades:
Thanks!

PR #2407, opened 2021-09-10
Custom CSS Revamp

CqldFire:
Fix some spacing, organize things better, add dark mode variables to make changing dark mode colors easier in the future, remove unneeded semicolons. Nothing has been removed except some old code that was commented out that I accidentally left from when I developed this, just re-organized.

Note: **UNTESTED,** you would have to test before merging. I cannot test atm because I am at school.

### DIFF
--- a/custom/panel_templates/Default/assets/css/custom.css
+++ b/custom/panel_templates/Default/assets/css/custom.css
@@ -1,11 +1,11 @@
 /* --------------- Prevents light flicker on dark mode --------------- */
 
 body {
-    display: none
+    display: none;
 }
 
 .visible {
-    display: block
+    display: block;
 }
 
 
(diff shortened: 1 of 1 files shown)


custom/panel_templates/Default/assets/css/custom.css
```diff
@@ -1,65 +1,77 @@
 /* --------------- Prevents light flicker on dark mode --------------- */
 
 body {
-    display: none;
+    display: none
```
samerton, line 4: Can we keep semi-colons at the end of lines please?

CqldFire: We can... my only issue with that is shipping it without the semi-colons saves users a small amount of code they need to download, and for users with slow connections the smallest amount of code can make a decent difference. Should I add them back or no?

Derkades: the real solution to that problem would be minimizing these files for the upgrade and install packages

CqldFire: That would be a great addition :D 
I would still like to merge this in some form though because improvements like the variables

CqldFire:
All right added semi-colons back and changed dark mode colors slightly to look MUCH better, is it ok now sam?

Aberdeener:
Can we get before/after screenshots @CqldFire ?

Aberdeener:
Yeah

CqldFire:
Screenshot 2021-09-22 at 2 27 35 PM
Screenshot 2021-09-22 at 2 27 08 PM

Darker is after, lighter is before

PR #2420, opened 2021-09-26
Update install.php

SamertonBby:


Aberdeener:
Thanks for the pull request. Looks like something may have gone wrong, though.
What was the intended addition/change here? :)

PR #2423, opened 2021-09-26
Add new dynamic Group Sync injectors system + separate Discord into own module

Aberdeener:
Module name: "Discord Integration" (open for different ideas, but this seems good to me)

It incorporates:
- bot integration
- widget
- etc

This will be beneficial to us maintainers as we can touch Discord things without being super worried about breaking other things, and better for users as they can now disable this module if they want absolutely no Discord features to keep their site simpler to use.

Adds the `Util::isModuleEnabled(string $name)` function. I have taken liberty and replaced some legacy check for if the Forum module is enabled with this new way.

Should be done in next few days!
Open for reviews, help and constructive criticism.


Aberdeener:
Initial testing showing good results, will do much more tomorrow

modules/Discord Integration/init.php
```diff
@@ -0,0 +1,7 @@
+<?php
+
+require_once(ROOT_PATH  . '/modules/Discord Integration/module.php');
```
Derkades, line 3: double space :-)

Aberdeener:
All that is left is languages :)

Aberdeener:
Good to merge right after pre12 is released as this is targeting 2.0.0 release

Derkades:
> Good to merge right after pre12 is released as this is targeting 2.0.0 release
Or before, since the `v2` branch is now separate from pr12

PR #2437, opened 2021-10-05
Make hook fire even if no hooks are there

supercrafter100:
Due to a coding mistake, the event was only fired when a webhook was configured for this forum. The code update will always fire the event.


partydragen:
Thanks

PR #2458, opened 2021-10-16
Change naming of HookHandler class properties and methods

tadhgboyle:
PRing so others have a chance to veto or change any language I have used

With so much stuff being named "hook", this class was always a headache to work on.

Now, modules register "events", and can register a "listener" for any registered event.
As well, "hooks" has been renamed to "webhooks" where applicable, though I think splitting webhook management into its own class would be good.

### DIFF
--- a/core/classes/EventHandler.php
+++ b/core/classes/EventHandler.php
@@ -20,10 +20,6 @@ class EventHandler {
      * @param array|null $params Array of available parameters and their descriptions.
      */
     public static function registerEvent(string $event, string $description, array $params = []): void {
-        if (!isset(self::$_events[$event])) {
-            self::$_events[$event] = [];
-        }
-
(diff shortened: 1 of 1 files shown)


Derkades:
Maybe change the class name HookHandler to something else as well? EventHandler? Or simply Events?

samerton:
I guess the one thing I'd comment is that these aren't specific to webhooks

One example of an existing implementation is the delete hook (event) - this is used by other modules to hook into when a user gets deleted to ensure all their data is removed from the database

samerton:
... personally I'd keep hooks 😄 

Hooks are something which plugs into a system to extend its functionality, in this case internally

Whereas I'd say webhooks are a way to communicate with an external system

Aberdeener:
Right, so you’d veto the event/listener wording?

PR #2479, opened 2021-11-02
Add clickable link in toastr when requesting debug url

supercrafter100:
Add the ability to click a "here" text when the toaster pops up when the debug link is generated. This will ensure that people can always get the URL, even if the copying to their clipboard for some reason fails


Aberdeener:
Thank you

PR #2490, opened 2021-11-14
German updated

Justman10000:


Aberdeener:
Thank you

Justman10000:
> 
> 
> Thank you

I want to update the whole german package, which form of address should I use? Du or Sie?

Aberdeener:
I would follow whatever convention is already in place. But informal "Du" seems fine to me!

Justman10000:
Okay, not that I have to rewrite the whole thing then 😅

PR #2501, opened 2021-11-21
Add permission check to debug link query

supercrafter100:
This is to prevent any random from the internet from generating a debug link and getting all the debug data from your website.


Aberdeener:
Can you also add a check on the error exception page and the debugging + maintenance page so that the debug link button is only shown if they have this permission?

supercrafter100:
The current permission is the one required to view the debugging & maintenance page. If you rather have the permission seperate, let me know

Aberdeener:
Thanks

PR #2517, opened 2021-11-29
Use Composer for dependencies

Aberdeener:
Why this should occur:
- *Much* easier to update our deps later on.
- Less to think about when using these deps, since all we need to do is run the composer autoloader to use the packages.
- In the future can use other composer packages (such as Swagger-PHP / OpenAPI and PHPUnit).

Change needed for releases:
- Since most of our users are on cheap webhosts, they won't be able to run `composer install --no-dev` themselves, so we should offer (and default to) .zips with the `vendor/` directory already there and `composer install --no-dev` pre-run.
- Maybe we can also offer a .zip without the `vendor/` directory to allow the user to run `composer install --no-dev`, but this isn't needed since it would be the exact same as the first option I mentioned. 

Tested:

- [x] SqlFormatter
- [x] Smarty
- [x] HTMLPurifier
- [x] Emojione - updated from 1 deprecated method
- [x] Bulletproof
- [x] PHPMailer
- [x] SitemapPHP
- [x] TwoFactorAuth
- [x] PHPMarkdown
- [x] HTMLToMarkdown


Aberdeener:
Complete 🎉 
Waiting for feedback internally.

samerton (approved):
Looks good to me, thanks!

PR #2518, opened 2021-11-30
Add userBanned event

supercrafter100:
Adds a userBanned event that will get fired when a user is banned. Yes its that simple


Aberdeener:
Thanks!

Derkades:
missing languages!!!

PR #2520, opened 2021-12-04
Make switchery be in the template and not in the `case`s

PadowYT2:


Aberdeener:
Personally against this, as *most* pages don't need the JS switches - we shouldn't waste time attempting to initialize them.

Derkades:
it's probably not worth the code repetition though, imo

PR #2527, opened 2021-12-08
Server Compatibility icons on the left side

TeemoCell:
Server Compatibility  iocns on the left side

Looks way more clean.

**Before**
image

**After**
image

### DIFF
--- a/custom/panel_templates/Default/index.tpl
+++ b/custom/panel_templates/Default/index.tpl
@@ -104,8 +104,8 @@
                                 <!-- Card Body -->
                                 <div class="card-body">
                                     {$NAMELESS_VERSION}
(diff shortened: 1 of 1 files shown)


Aberdeener:
I prefer it how it is currently, let's see if anyone else has thoughts

partydragen:
> I prefer it how it is currently, let's see if anyone else has thoughts

agree

TeemoCell:
The icons on the left side looks way more cleaner xD

partydragen:
> The icons on the left side looks way more cleaner xD

Yea i kinda agree more now after i thinked about it

Aberdeener:
Thank you

PR #2532, opened 2021-12-12
Add proper example server block

supercrafter100:
This PR adds a complete server block for nginx. This is to prevent confusion for people that have never touched nginx and don't understand where to put the server blocks. This example will likely not work as-is as it still requires some configuration on the user their end.


samerton:
Not sure if we should go with this:

- Maintenance overhead with updating the PHP versions
- People will likely simply copy + paste, the advantage of just having the server block means users should be reading into it more to understand what they need to do

supercrafter100:
Understandable.

PR #2533, opened 2021-12-13
Correct Update Message for SEO Page

TeemoCell:


Aberdeener:
Can you add the new string to all of the languages? Good to me as soon as that is done.

TeemoCell:
> Can you add the new string to all of the languages? Good to me as soon as that is done.

Yeah will do that

PR #2534, opened 2021-12-13
Froum Topic - Group Badges

TeemoCell:
Old Group badges formating is way to big and looks really ugly when you have a group with a long name or your have a lot of groups.
image

That is my suggestion 
image


Aberdeener:
Brilliant, thank you!

PR #2536, opened 2021-12-15
Custom pagination arrows

graphiteisaac:
This is a pretty simple PR, just allows people to set custom content for the pagination left / right buttons in a similar fashion to how pagination classes are currently set in the template.

Sorry for the big mess this became, I mixed another PR I'm working on into it. Hopefully it's still able to be used, this is a first for me contributing to something open source lol


samerton (changes requested):
Just as a general comment, I wonder if we should be looking at making the pagination options some sort of object/array instead of having to define 3 variables separately (class, left + right content)

graphiteisaac (commented):
Thank you, much cleaner!

graphiteisaac (commented):
Cleaned up my proposal thanks to a suggestion from Samerton

Aberdeener:
Thanks!

PR #2545, opened 2021-12-27
UltimateNamelessMaster

Justman10000:
UltimateNamelessMaster specially created for NamelessMC

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,6 +71,9 @@ To discuss changes with fellow translators, visit the [NamelessMC Translators](h
 #### Translation credits
 Translation credits can be found within CONTRIBUTORS.md
 
+## Automatic Managment
+[UltimateNamlessMaster](https://github.com/Justman10000/UltimateNamelessMaster) was created for the automatic management of NamelessMC
+
 ## Contributing
 We welcome all contributions of code and translations. Please feel free to fork the repository on GitHub and create any pull requests.
 - We generally keep a todo list in the Milestones tab of the the Issues page.


PadowYT2:
I don't think this is needed

Auto install is already a [thing](https://github.com/NamelessMC/Nameless/blob/v2/cli_install.php) and theres an another [installer](https://github.com/NamelessMC/Nameless-Installer)
Auto uninstall, u can just remove the folder

Aberdeener:
Thank you so much for taking the time to make this.

However, I will not be merging this PR, for the same reasons Padow mentioned above + some more;
- We have 3 installation methods already
- Uninstalling is just deleting the folder and database
- Maintenance overhead - this would be another thing we have to worry about when releasing an update
- 3rd party code - since this repo is managed by you, I wouldn't feel 100% certain about adding a link to it in our official README.md since it is not maintained by the official NamelessMC team

Justman10000:
> Thank you so much for taking the time to make this.
> 
> However, I will not be merging this PR, for the same reasons Padow mentioned above + some more;
> 
>     * We have 3 installation methods already
> 
>     * Uninstalling is just deleting the folder and database
> 
>     * Maintenance overhead - this would be another thing we have to worry about when releasing an update
> 
>     * 3rd party code - since this repo is managed by you, I wouldn't feel 100% certain about adding a link to it in our official README.md since it is not maintained by the official NamelessMC team

The maintenance would of course be included, you can also open pull requests or create issues! It has already cost me time and I am sure that I will not develop a bash script for 15-30 minutes, just so that it will then be rejected!

In addition, even more functions are planned in the script...

Justman10000:
> I don't think this is needed
> 
> Auto install is already a [thing](https://github.com/NamelessMC/Nameless/blob/v2/cli_install.php) and theres an another [installer](https://github.com/NamelessMC/Nameless-Installer) Auto uninstall, u can just remove the folder

Also, this is an installer in PHP, not Bash...

PR #2547, opened 2021-12-28
German updated

Justman10000:


Aberdeener:
Looks like there’s some issues with the language files. Can you check them out and then fix? Thanks

Justman10000:
> Looks like there’s some issues with the language files.

In what way?

Justman10000:
> Looks like there’s some issues with the language files.

Hat sich erledigt! Seltsam, es kam die Meldung, dass der String nicht existierte!

Justman10000:
> Looks like there’s some issues with the language files.

It's been dealt with! Strange, the message came up that the string did not exist!

Aberdeener:
Thanks!

Derkades:
> Strange, the message came up that the string did not exist!

If you're using pr12, it has been fixed since then

PR #2561, opened 2022-01-12
API route binding

Aberdeener:
API "route binding" is a feature which will allow us to simplify the DX of most of our API endpoints, especially simple GET ones.

### Without route binding:
- GET requests are ugly and annoying to debug
- Endpoints have to validate the params themselves
- To use variables passed to the endpoint, they are stored in a simple `<string, string>` array, with no real type checking.
- Example request: `https://example.com/api/users/ban?id=1`

### With route binding:
- "Seamless" routes
- All param validation is done before the request is sent to an Endpoint class
- Params are passed as arguments to the `execute(Nameless2API $api, ...)` method in the Endpoint class, with type checking.
- Example request: `https://example.com/api/users/1/ban`

### How does it work?
Given the second example request above:
- An Endpoint, in this case probably something called `UserBanEndpoint` would register the route as `users/{user}/ban`
- When a request is made, we look for a "transformer" for the `{user}` type
    - Transformer closures are required to have 2 arguments, the first one being a `Nameless2API` instance and the second is the raw value from the request (1)
	- This transformer is built into the Core module (see it's module.php)
	- We call the transformers closure to generate a User object from it's ID (1)
		- Since transformers are required to have an `Nameless2API` argument, if something goes wrong while generating the User object we can call `$api->throwError(...);`
- Finally, the `execute(...)` method in the Endpoint class will be provided the resolved User object in its params. 

### Notes
- Modules can (obviously) register their own transformers using `Endpoints::registerTransformer(string $type, Closure $transformer);`
- Code is yucky, I'll deal with it tomorrow!
- Might add something to the API section of StaffCP to see all registered transformers...
- This will require us to change API endpoints in the next update when merged, so anything which uses an old Endpoint route will not work. Since I think its mainly stuff maintained by us that uses the API, we should be fine - but yeah. Breaking changes! whee

### DIFF
--- a/core/classes/Endpoints/Endpoints.php
+++ b/core/classes/Endpoints/Endpoints.php
@@ -16,6 +16,11 @@ class Endpoints {
     private iterable $_endpoints = [];
 
     /**
+     * @var array Mapping of key names to closures to transform a variable into an object (ie, a user ID to a User object)
+     */
+    private static array $_transformers = [];
+
+    /**
      * Register an endpoint if it's route is not already taken.
      *
      * @param EndpointBase $endpoint Instance of endpoint class to register.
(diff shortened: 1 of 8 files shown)


modules/Core/includes/endpoints/GetUserEndpoint.php
```diff
@@ -0,0 +1,19 @@
+<?php
```
samerton, line 1: What do we want to do with this? 😄

core/classes/Endpoints/Endpoints.php
```diff
@@ -28,55 +33,162 @@ public function add(EndpointBase $endpoint): void {
...
+        $endpoint_parts = explode('/', $endpoint->getRoute());
+        $endpoint_vars = [];
+
+        $route_parts = explode('/', $route);
+        $route_vars = [];
+
+        $i = 0;
+        // first, find any variables (e.g. {user}) in the endpoint's route
+        // we save them to an array with their index, so we can reference them later
+        foreach ($endpoint_parts as $part) {
```
samerton, line 129: As we've already discussed, this could be
```php
foreach ($endpoint_parts as $i => $part) {
```
to save manually creating + updating the $i var

modules/Core/includes/endpoints/GetUserEndpoint.php
Aberdeener: it's been deleted in the next API PR 😅

samerton (approved):

PR #2567, opened 2022-01-16
Emailing refactor!

Aberdeener:
Instead of having a condition for sending mail over the `mail()` function vs PHPMailer, the class handles this on it's own now.

`Email::send(...)` now takes:
- array of "email" and "name" strings to determine the recipient
- string email subject
- string email message
- optional array of "email" and "name" strings to determine the reply-to address

`Email::addPlaceholder(...)` method for modules to add placeholders to their emails

### DIFF
--- a/core/classes/Core/Email.php
+++ b/core/classes/Core/Email.php
@@ -22,6 +22,11 @@ class Email {
     public const MASS_MESSAGE = 6;
 
     /**
+     * @var array Placeholders for email templates
+     */
+    private static array $_message_placeholders;
+
+    /**
      * Send an email.
      *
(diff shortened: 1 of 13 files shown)


core/classes/Core/Email.php
```diff
@@ -10,25 +10,48 @@
 */
 
 use PHPMailer\PHPMailer\PHPMailer;
+use PHPMailer\PHPMailer\SMTP;
 
 class Email {
 
+    public const REGISTRATION = 1;
```
samerton, line 17: Could do with making this extensible (at some point, I appreciate it wasn't before!) by other modules (FORUM_TOPIC_REPLY for example is obviously not part of the core, rather the Forum module)

samerton (approved):

PR #2570, opened 2022-01-17
OAuth register/login/linking!

### LINKED ISSUE #451: Social logins
I was hoping for the ability to register and login using those "Sign In With Google" and "Sign In With Facebook" buttons and to link those accounts to your regular account. Maybe you can link your Minecraft account with your account using something like Enjin's system?

Aberdeener:
5-10 mins of local configuring needs to be done for a site to use this.

Closes #451
Closes NamelessMC/Nameless-Modules#15


zJerino:
**Good work**

modules/Core/pages/user/oauth.php
```diff
@@ -33,13 +33,11 @@
         $oauth_user_provider_id = $oauth_user[OAuth::getInstance()->getIdName($provider_name)];
 
         if (OAuth::getInstance()->userExistsByProviderId($provider_name, $oauth_user_provider_id)) {
-            Session::flash('oauth_error', 'User already exists with that provider ID');
-            Redirect::to(URL::build('/login'));
+            Session::flash('oauth_error', "Another NamelessMC user is already linked to that $provider_name account.");
```
Derkades, line 36: translations!

Aberdeener: thanks, added to the Todo list

PadowYT2 (commented):
Looks all great, also I've modified the [docs page](https://docs.namelessmc.com/en/oauth).
Also, why not making the OAuth methods to register too?

Aberdeener:
> Looks all great, also I've modified the [docs page](https://docs.namelessmc.com/en/oauth). Also, why not making the OAuth methods to register too? And what is the redirect url should be put?

Thanks so much for doing that!
The redirect URLs:
- `http://example.com/index.php?route=/login/oauth/&provider=<provider>`
- `http://example.com/index.php?route=/user/oauth/&provider=<provider>`
*where `<provider>` is "google" or "discord"*

I would like to add OAuth registration - but it brings a whole other layer of complexity and potential bugs.
If a user registers with OAuth, all we would do is get their email from the provider, and autofill that on the registration page - but this could get complicated with things like AuthMe configured.

I'm going to take another shot at it, but for the little convenience it brings (since they dont have to link the provider afterwards), I'm not sure it is totally worth it.

custom/templates/DefaultRevamp/user/oauth.tpl
```diff
@@ -0,0 +1,113 @@
...
+                            </tbody>
+                        </table>
+                    {/nocache}
+                </div>
+            </div>
+        </div>
+    </div>
+</div>
+
+{foreach $OAUTH_PROVIDERS as $provider_name => $provider_data}
```
partydragen, line 79: Modal missing translation

partydragen (changes requested):

PadowYT2:
Is discord really working?
a
a
The url I'm getting:
`https://discord.com/oauth2/authorize?scope=identify%20email&state=8547128e5ab6e9e8341f66d293b05527&response_type=code&approval_prompt=auto&redirect_uri=https%3A%2F%2Fdev.namelessmc.xyz%2Findex.php%3Froute%3D%2Flogin%2Foauth%2F%26provider%3Ddiscord&client_id=876678761582563338`

custom/templates/DefaultRevamp/user/oauth.tpl
Aberdeener: Added

PR #2578, opened 2022-01-20
Quick_Action

xHelloC:
I picked up the latest version now, but I'm really not sure about the .tpl
image
image


Derkades:
Unfortunately it looks like these files are still based on pr12

xHelloC:
Ofcourse, but its only changed few rows

so 22. 1. 2022 v 2:10 odesílatel Robin ***@***.***> napsal:

Aberdeener (changes requested):
Are you able to revert these pr12 files & if possible, implement the style and static analysis changes requested by the bot? Thanks!

PR #2602, opened 2022-02-13
Add PHPDocs to core classes for new class docs site

Aberdeener:


Aberdeener:
well that was boring to do

PR #2604, opened 2022-02-18
Select menu

### LINKED ISSUE #2043: Improve the functionality and style of Group select
At the moment the Groups section of user edit is quite bland and due to it being just a select list can easily be messed up due to you clicking on a new group without holding ctrl will remove all other groups from the player, and since the group box is quite small if you have a lot of groups it can be hard to find any of them. 

It would be nice if the group selection area could basically be made bigger and  more visually pleasing & that clicking and then clicking on a group is how you select/deselect them without having to hold ctrl.

One idea could be to make the entire group area just two list, one which is like "User Groups" and then another one with "Available Groups", like this:
nvnvbnvbn
and then you can just drag a group to the other side:
gsfjhfgkfgk

supercrafter100:
Makes the select menus a lot less confusing and work better generally. Also looks better.

Closes #2043 

**Example:**
Gif of the action

Let me know if anything needs to be changed


supercrafter100:
Still has the commits from my other PR whoops. Whatever

Derkades:
Nice!

Aberdeener:
a

PR #2613, opened 2022-03-01
Add Old PHP Versions support

OskiBoski483:


samerton:
Hi,

Thanks for the PR! Unfortunately, we can't do something like this because if you are using a PHP version less than 7.4, you will immediately see a syntax error as a lot of our code has been refactored to make use of 7.4 features.

I think this is perfectly justified as it makes our lives easier developing new features (so we can take advantage of the new features + syntax), and also 7.3 has had no security updates from PHP themselves since November 2021.

PR #2618, opened 2022-03-10
Verify captcha before allowing to save

supercrafter100:
Altough this doesn't work for every single captcha type. It made more sense to have it for the ones that support it.

**hCaptcha**: 
- [ ] Verify secret key
- [x] Verify public key

**reCaptchaV2**:
- [x] Verify secret key
- [ ] Verify public key

**reCaptchaV3**:
- [x] Verify secret key
- [x] Verify public key

This may not be ideal that it doesn't support everything but it does mean there is some protection in place to prevent people from activating captcha with invalid configurations.

### DIFF
--- a/modules/Core/classes/Captcha/Recaptcha2.php
+++ b/modules/Core/classes/Captcha/Recaptcha2.php
@@ -28,6 +28,21 @@ class Recaptcha2 extends CaptchaBase {
         return $result['success'] == 'true';
     }
 
+    public function validateSecret(string $secret) : bool {
+        $token = "Verification";
+        $url = 'https://www.google.com/recaptcha/api/siteverify';
+
+        $result = HttpClient::post($url, [
+            'secret' => $secret,
+            'response' => $token
+        ])->json(true);
+        return !($result['error-codes'][0] == 'invalid-input-secret');
+    }
+
+    public function validateKey(string $key) : bool {
+        return true; // No way to verify
+    }
+
     public function getHtml(): string {
         return '<div class="g-recaptcha" data-sitekey="' . $this->getPublicKey() . '"></div>';
     }
--- a/modules/Core/classes/Captcha/Recaptcha2.php
+++ b/modules/Core/classes/Captcha/Recaptcha2.php
@@ -36,7 +36,7 @@ class Recaptcha2 extends CaptchaBase {
             'secret' => $secret,
             'response' => $token
         ])->json(true);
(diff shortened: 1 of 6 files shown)


modules/Core/classes/Captcha/hCaptcha.php
```diff
@@ -28,6 +28,20 @@ public function validateToken(array $post): bool {
...
+    public function validateSecret(string $_secret) : bool {
+        return true; // Haven't found a way to verify this
+    }
+
+    public function validateKey(string $key) : bool {
+        $url = 'https://api.hcaptcha.com/checksiteconfig?sitekey=' . $key;
+        $client = HttpClient::get($url);
+        if ($client->hasError()) {
+            return false;
+        } else {
```
Aberdeener, line 40: unneeded `else`

modules/Core/classes/Captcha/Recaptcha2.php
```diff
@@ -28,6 +28,21 @@ public function validateToken(array $post): bool {
...
 
+    public function validateSecret(string $secret) : bool {
+        $token = "Verification";
+        $url = 'https://www.google.com/recaptcha/api/siteverify';
+
+        $result = HttpClient::post($url, [
+            'secret' => $secret,
+            'response' => $token
+        ])->json(true);
+        return !($result['error-codes'][0] == 'invalid-input-secret');
```
Aberdeener, line 39: can be simplified to `!==`

modules/Core/pages/panel/registration.php
```diff
@@ -109,18 +65,79 @@
...
+                        (
+                            CaptchaBase::isCaptchaEnabled('recaptcha_login')
+                            || CaptchaBase::isCaptchaEnabled()
+                        )
+                        && (
+                            CaptchaBase::getActiveProvider()->validateSecret(Input::get('recaptcha_secret')) == false
+                            || CaptchaBase::getActiveProvider()->validateKey(Input::get('recaptcha')) == false
+                        )
+                    ) {
+                        $errors = [$language->get('user', 'invalid_recaptcha')];
```
Aberdeener, line 80: Hmm, maybe an error specific to "hey person you didnt enter the right info" is better than a generic one?

Aberdeener (changes requested):

Aberdeener (approved):

Aberdeener:
Thank you!

PR #2624, opened 2022-03-11
Update - SB Admin 2 to v4.1.3

TeemoCell:
Updated the panel template to the latest version . Tested it and everything is still working.


custom/panel_templates/Default/assets/js/sb-admin-2.js
```diff
@@ -6,22 +6,21 @@
...
-                    l = o.wheelDelta || -o.detail;
-                (this.scrollTop += 30 * (l < 0 ? 1 : -1)), e.preventDefault();
-            }
-        });
-
-})(jQuery);
-// @license-end
+    !(function (s) {
+        "use strict";
+        s(window).resize(function () {
```
supercrafter100, line 14: Is it correct that you have "s" instead of "$" here? And in the other places

Issue #1528, opened 2019-06-07
Change "Admin" user to "Root" or "WebAdmin"


Derkades:
So it doesn't conflict with existing roles people might have on their Minecraft server and make everything more confusing

PixelNotFound:
People can change the name of the admin rank, I always rename it to CEO like the ranks on the mc server.

A248:
This is already configurable in v1. I've changed the name of the Admin role to `SysAdmin`

Derkades:
People seem to think it should be "root", I agree.
![image](https://user-images.githubusercontent.com/15892014/71782835-dfab1d80-2fde-11ea-8ffd-ddfb1ecb5015.png)

A248:
Yes, I voted for `B` as well.

Derkades:
Updated poll:
![image](https://user-images.githubusercontent.com/15892014/136838813-fbd8d40c-4d17-4de9-94d4-0f4c012d919b.png)

PR #2628, opened 2022-03-11
Change "Admin" user to "Root"

TeemoCell:
I hope that is correct, as it was meant in [#1528](https://github.com/NamelessMC/Nameless/issues/1528)


partydragen:
No, Owner was choosed and there will be new admin group and require permissions changes a few places within the installer etc

TeemoCell:
Oh okay, I close the PR then.

Issue #1691, opened 2020-06-27
Maintenance page not saving image Height/Width values


HeavenCrafter:
**Describe the issue**
I was trying to setup a custom maintenance information page. Wanted to make it look cool by adding some images. While doing so, I set the width to fix the text alignment by making it smaller which can be done simply on the image settings itself. Before submit, everything looked the way I wanted. After submission, it reverts the image size to the original size of the image.

**To Reproduce**
1. Go to StaffCP > Configuration > Maintenance
2. Proceed to the Maintenance Message editor
3. Add an image of your choice, in my case, is from an Imgur image link.
4. Change the size, in my case, 16x11 (16 Width x 11 Height)
5. Submit the changes.

**Expected behavior**
![image](https://user-images.githubusercontent.com/22242534/85913148-c20b9200-b864-11ea-8553-d9fa03be278a.png)

**Behavior that I got instead**
![image](https://user-images.githubusercontent.com/22242534/85913157-d8195280-b864-11ea-84ec-a38b16984e33.png)

**NamelessMC version (from AdminCP -> Overview)**
Latest (v2.0 Pre7)

Aberdeener:
This appears to be fixed with the use of TinyMCE - please feel free to reopen if the problem persists.

PR #2629, opened 2022-03-11
Maintenance Page HTML Code

TeemoCell:
This should fix the issue in #1691 . In case i used the right `Output` method.

### DIFF
--- a/core/classes/Core/Output.php
+++ b/core/classes/Core/Output.php
@@ -57,7 +57,7 @@ class Output {
             $purifierConfig->set('CSS.AllowedProperties', ['text-align', 'display', 'float', 'color', 'background-color', 'background', 'font-size', 'font-family', 'text-decoration', 'font-weight', 'font-style', 'font-size', 'width']);
             $purifierConfig->set('CSS.AllowTricky', true);
             $purifierConfig->set('CSS.MaxImgLength', null);
-            $purifierConfig->set('HTML.AllowedAttributes', 'target, rel, href, id, src, height, width, alt, class, style');
+            $purifierConfig->set('HTML.AllowedAttributes', 'target, rel, href, id, src, height, width, alt, class, *.style');
             $purifierConfig->set('Attr.AllowedFrameTargets', ['_blank', '_self', '_parent', '_top']);
             $purifierConfig->set('Attr.AllowedRel', ['noopener', 'nofollow']);
             $purifierConfig->set('HTML.SafeIframe', true);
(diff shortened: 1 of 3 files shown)


Aberdeener:
Did you test this?

TeemoCell:
Yeah, tested it. `Height/Width` and all other inline styles are working now.

partydragen:
@Aberdeener its a bug with the getPurified itself, discussed in discord

TeemoCell:
I try to fix the `getPurified` then.

TeemoCell:
Fixed

Derkades:
Is this safe? Is getPurified used anywhere else?

TeemoCell:
Yeah it is used in quite a few files 
image

core/classes/Core/Output.php
```diff
@@ -53,9 +53,10 @@ public static function getPurified(?string $input, bool $escape_invalid = false)
             $purifierConfig->set('HTML.Doctype', 'XHTML 1.0 Transitional');
             $purifierConfig->set('URI.DisableExternalResources', false);
             $purifierConfig->set('URI.DisableResources', false);
-            $purifierConfig->set('HTML.Allowed', 'u,a,p,b,i,small,blockquote,span[style],span[class],p,strong,em,li,ul,ol,div[align],br,img,figure,figcaption');
-            $purifierConfig->set('CSS.AllowedProperties', ['text-align', 'display', 'float', 'color', 'background-color', 'background', 'font-size', 'font-family', 'text-decoration', 'font-weight', 'font-style', 'font-size']);
+            $purifierConfig->set('HTML.Allowed', 'u,a,p,b,i,small,blockquote,img[style|width|height|alt|src],span[style|class],div[style|align],p,strong,em,li,ul,ol,br,img,figure,figcaption');
+            $purifierConfig->set('CSS.AllowedProperties', ['text-align', 'display', 'float', 'color', 'background-color', 'background', 'font-size', 'font-family', 'text-decoration', 'font-weight', 'font-style', 'font-size', 'width']);
             $purifierConfig->set('CSS.AllowTricky', true);
+            $purifierConfig->set('CSS.MaxImgLength', null);
```
samerton, line 59: Why is this change necessary? According to the HTMLPurifier docs, it opens up to imagecrash attacks (imagine someone adding a height + width of millions of pixels 😄)

TeemoCell: I've tested it myself with a 50000x50000 image but we can add add a limit.

TeemoCell:
Changed it to `1920px` in the last commit which should prevent imagecrash attacks

Derkades:
So this is no longer necessary since the TinyMCE switch? @Aberdeener

core/classes/Core/Output.php
partydragen: > 

i still believe its a issue, i had way to big image many years ago with made the website extremly slow with being the case @samerton telling i guess

Aberdeener:
> So this is no longer necessary since the TinyMCE switch? @Aberdeener

i believe so. closing - if this happens again we can reopen to investigate.

PR #2631, opened 2022-03-13
Update portal.tpl

progremmerpro2021:


WavingCatDevs:
Why is this necessary?

Aberdeener:
I think this would just introduce more confusion - is there a specific situation you're thinking of here? :)

TeemoCell:
I would keep the old Placeholder.

PR #2633, opened 2022-03-14
Use native bootstrap switches

Derkades:
image

Looks much better imo, please let me know what you all think before I convert everything. We can make them bigger if needed for looks. Clicking text now also toggles the switch so it's not a big deal that they're small.

Conversion steps for all StaffCP `<input type="checkbox"` elements:
1. Add `custom-control custom-switch` to form-group `<div>` class
2. Replace `class="js-switch"` with `class="custom-control-input"` for `<input>`
3. Move `<label>` below <input> instead of above `<input>`
4. Remove `style="margin-right:10px"` from `<label>` or `<span>`
5. Add `class="custom-control-label"` to `<label>`


Aberdeener:
This is excellent, I'm totally in favour of swapping to this. Thanks Robin

CqldFire:
It looks good, but I think the switches should be a bit larger
- drew

PadowYT2:
How will this look for switching dark/white?

TeemoCell:
check #2634 :D

Aberdeener:
> For some reason switches don't work properly in Layout > Templates > DefaultRevamp Settings > Permissions, I do not know why. Only select/deselect all works

are we [you] replacing the green + red permission selection boxes?

TeemoCell:
> For some reason switches don't work properly in Layout > Templates > DefaultRevamp Settings > Permissions, I do not know why. Only select/deselect all works

No idea what happen there https://assets.uedit.co/02802c74-1ab2-4cfc-bfa9-8f0165825748.gif

TeemoCell:
Found the problem.
https://github.com/NamelessMC/Nameless/blob/932411e87247702eff07c62164ba4e9587e762c3/custom/panel_templates/Default/core/template_permissions.tpl#L62
Change to `id="perm-use-{$group->id|escape}"`

Derkades:
I see, apparently `for` on `<label>` refers to the id not the name.

TeemoCell:
Looks fine for me. Everything working so far.

Aberdeener:
:shipit: Thanks so much you two!

PR #2637, opened 2022-03-21
Update translations

ViaSnake:
Two changes are being made in this pull request.

- Japanese translation has been updated.
- Compared to the original translation(EnglishUK), missing strings were added and unused strings were removed.


custom/languages/Czech/installer.php
```diff
@@ -103,5 +104,5 @@
     'no_converters_available' => 'Nejsou dostupné žádné konvertory.',
     'config_not_writable' => 'Konfigurační soubor není zapisovatelný.',
 
-    'session_doesnt_exist' => 'Nepodařilo se detekovat relaci. Pro správnou funkci Nameless je potřeba funkce ukládání relací. Zkuste to prosím znovu. Pokud chyba přetrvává, kontaktujte podporu vašeho webhostingu.'
```
Aberdeener, line 106: Why was this translation overwritten?

custom/languages/Chinese(Simplified)/installer.php
```diff
@@ -104,5 +105,5 @@
     'no_converters_available' => '没有可用的转换器',
     'config_not_writable' => '配置文件无法写入',
 
-    'session_doesnt_exist' => '无法检测会话。若想使用 Nameless，您必须启用会话。请您再试一次，若此问题始终存在，请咨询您的网页提供商。'
```
Aberdeener, line 107: Why was this translation overwritten?

custom/languages/Danish/installer.php
```diff
@@ -102,5 +103,5 @@
     'no_converters_available' => 'Der er ingen omdannere tilgængelige.',
     'config_not_writable' => 'Konfigurationsfilen kan ikke skrives i.',
 
-    'session_doesnt_exist' => 'Ikke i stand til at opdage session. Session gemning er et krav for at bruge Nameless. Prøv igen, og hvis problemet fortsætter, bedes du kontakte din webvært for støtte.'
```
Aberdeener, line 105: Why was this translation overwritten?

custom/languages/Dutch/installer.php
```diff
@@ -101,5 +102,5 @@
     'no_converters_available' => 'Er zijn geen converters beschikbaar.',
     'config_not_writable' => 'Het configuratiebestand is niet beschrijfbaar.',
 
-    'session_doesnt_exist' => 'Niet in staat om de sessie te detecteren. Sessies opslaan is een vereiste om Nameless te gebruiken. Probeer het opnieuw, en als het probleem blijft bestaan, neem dan contact op met uw webhost voor ondersteuning.'
```
Aberdeener, line 104: Why was this translation overwritten?

custom/languages/SwedishSE/installer.php
```diff
@@ -103,5 +104,5 @@
     'no_converters_available' => 'Det finns inga konverterare tillgängliga.',
     'config_not_writable' => 'Konfigurationsfilen är inte skrivbar.',
 
-    'session_doesnt_exist' => 'Det gick inte att upptäcka sessionen. Sessionsbesparingar är ett krav för att använda Nameless. Försök igen, och om problemet kvarstår, kontakta din webbhotell för support.'
```
Aberdeener, line 106: Why was this translation overwritten?

Aberdeener (changes requested):
Thanks so much for this.
Looks like you used some sort of automation to do it? Would be curious to see how that works!
The downside to this is that some of the translations are overwritten - I've made comments to mark them, they all seem to be the session error message...

ViaSnake:
> Thanks so much for this. Looks like you used some sort of automation to do it? Would be curious to see how that works! The downside to this is that some of the translations are overwritten - I've made comments to mark them, they all seem to be the session error message...

Unfortunately all changes were made by myself. Therefore, I did not use any automation or tools.

As for `session_doesnt_exist`, I overwrite it if the translation has not been updated since 88b1b0d.
With this change, the user will see the untranslated string until the next time someone updates it. Instead, when someone updates the translation, they will easily notice that the string has changed.

For me, it is very helpful to know immediately that a string has been changed. However, not all users will benefit from this. If this change is undesirable, I will revert.

Aberdeener:
Ah ok, I understand!

Thank you so much for the PR as always

PR #2638, opened 2022-03-24
Remove markdown + CKEditor

Aberdeener:
Closes #2638


core/classes/Core/Input.php
```diff
@@ -55,20 +55,29 @@ public static function get(string $item) {
...
-          menubar: \'table\',
-          convert_urls: false,
-          plugins: \'autolink,codesample,directionality,emoticons,hr,image,link,lists,spoiler,code,table\',
-          toolbar: \'undo redo | bold italic underline strikethrough formatselect forecolor backcolor ltr rtl | alignleft aligncenter alignright alignjustify | codesample code emoticons hr image link numlist bullist | spoiler-add spoiler-remove\',
-          spoiler_caption: \'' . $language->get('general', 'spoiler') . '\',
-          default_link_target: \'_blank\',
-          skin: "' . (defined('TEMPLATE_TINY_EDITOR_STYLE') ? TEMPLATE_TINY_EDITOR_STYLE : 'oxide') . '"
-        });
-        ';
+        // TODO: use oxide-dark skin when dark theme is enabled?
```
Derkades, line 58: doesn't this already happen because defaultrevamp sets TEMPLATE_TINY_EDITOR_STYLE when dark theme is enabled?

Aberdeener: Good point, fixed. Thank you

samerton (approved):

custom/templates/DefaultRevamp/template.php
```diff
@@ -68,7 +68,7 @@ public function __construct($cache, $smarty, $language, $user, $pages) {
 
         if (isset($darkMode) && $darkMode == '1') {
             $smartyDarkMode = true;
-            define('TEMPLATE_TINY_EDITOR_STYLE', 'default-revamp');
```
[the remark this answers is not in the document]
Aberdeener: It was a dark mode version of the built in "oxide" theme, but now "oxide-dark" is built in so we don't need to supply our own version :D

PR #2641, opened 2022-03-25
Disable blank issues

PadowYT2:
People open wrong issues and this anyways is not needed


Derkades:
Thanks for the PR. Not sure what the others think but imo having the blank issue button is fine. Some people might indeed click the wrong button but I don't think we should inconvenience everyone else who does truly want to create an issue without a template because if this. Not really that big of a deal anyway

PR #2606, opened 2022-02-20
User Integrations systems

partydragen:
Add user integrations systems where modules can register their own integration types

Progress:
- [x] Allowing modules to register new integration types
- [x] Add "Connections" page to UserCP where they can link their integrations
- [x] Convert Minecraft part into the new user integrations system
- [x] Convert Discord part info the new user integrations system
- [x] Add Integrations page to StaffCP where they can modify settings for the integrations
- [x] Add integrations section to User page on StaffCP where they can view and manage integrations
- [x] Add option to enable or disable integration
- [x] Add option to force users to link a integration
- [x] Add option to disable unlinking
- [x] Add VerifyIntegration endpoint for Universal verification system
- [x] Show user integrations on user info endpoint
- [x] Update all endpoints to the new user integrations system
- [x] Update the installer to include the new tables
- [x] Update the upgrader script and convert Minecraft and Discord into the new user integrations system
- [x] Add new hook for when user integration being linked or updated
- [x] Add support for offline mode uuid
- [x] Load uuid into user data from user integrations

Registering from the website will make the minecraft integration unverified
Registering from the api will make the minecraft integration verified
Currently only way to verify mc integration after registering on website is by having api verification enabled so they execute /verify code ingame

Closes: https://github.com/NamelessMC/Nameless/issues/2052
Related: https://github.com/NamelessMC/Nameless/issues/1577#issuecomment-699061950

PR #2646, opened 2022-03-26
Support looking up a user in the API many different ways

Aberdeener:
code is gross, will change this to the user integration system when #2606 is merged

examples:
```
/users/3
/users/id:3
/users/username:Aberdeener
/users/integration_name:minecraft:Derkades
/users/integration_id:discord:271510274475819008
/users/integration_name:discord:aberdeener#0001
```


partydragen:
Hm there being 2 more queries releated to this

1 query to load integrations
and a another query to search for user integration

then it can load the normal user

Issue #2635, opened 2022-03-16
Update to Font Awesome v6


S4MON1X:
### Describe the feature

Font Aweasome v6 adds a lot of new icons and its styles.

WavingCatDevs:
I'll open a pr with this soon

WavingCatDevs:
Hmm, they seem to have changed the font awesome file things, ill let one of the developers sort this out.

samerton:
We need to do this whilst maintaining compatibility for version 5 so existing website icons don't break

S4MON1X:
This might help: https://fontawesome.com/docs/web/setup/upgrade/whats-changed#backward-compatibility

WavingCatDevs:
> We need to do this whilst maintaining compatibility for version 5 so existing website icons don't break

Yeah

TeemoCell:
Check #2651

samerton:
Removing the link since we still need to update Default Revamp

Aberdeener:
Fixed in 937c518d94a8df3ac4df8d9928392d4562abd872

Derkades:
Nice, with this commit the last bit of third party css/js was removed!

Aberdeener:
Ha, what do you mean? There is still so much

Derkades:
I meant js/css served from third party CDNs instead of locally

PR #2651, opened 2022-03-27
Panel Template - Font awesome 6

TeemoCell:
Related to #2635 . Checked all pages and looks like everything is working fine.


RobiNN1:
You don't need JS if you used FA6 with all.min.css file. JS version is redundant in this case.

TeemoCell:
> You don't need JS if you used FA6 with all.min.css file. JS version is redundant in this case.

Yeah true, forgot that. Just used FA6 like 2 times and for some reason i always include the `all.min.js`.

PadowYT2 (commented):
Maybe keep Font Awesome v5 because websites use it and all icons doesn't break next release?

TeemoCell:
As far as I know and have tested, FA5 also works with FA6. So far I have not found any icons that are not displayed correctly.

custom/panel_templates/Default/assets/css/sb-admin-2.min.css
```diff
@@ -7,4 +7,4 @@
  * Copyright 2011-2021 The Bootstrap Authors
```
samerton, line 7: Just wondering what the changes to this file are for?

PadowYT2: ~~well i just wasted my life~~
From: `font-family:'Font Awesome 5 Free'` to `font-family:'Font Awesome 6 Free'`

Derkades: Nice find, maybe that could be an override in custom.css instead?

samerton: I'm happy for this to be "baked" into the minified CSS for now since that's probably how it is in the original.

samerton:
Thanks!

custom/panel_templates/Default/assets/css/sb-admin-2.min.css
Derkades: Downside is it will break if someone updates sbadmin like in #2624

PR #2653, opened 2022-03-30
Allow uploading images through TinyMCE image uploader!

Aberdeener:
@Derkades What would be the standard here for returning the error header with the error text? Is what I have now okay?

Couple things to do:
- [ ] Show error message in TinyMCE popup, not just "http error 500"
- [ ] Token check

Deals with #2170 (can't see any open issues for it?)

### DIFF
--- a/core/classes/Core/Input.php
+++ b/core/classes/Core/Input.php
@@ -72,6 +72,7 @@ class Input {
               spoiler_caption: '{$language->get('general', 'spoiler')}',
               default_link_target: '_blank',
               skin: '$skin',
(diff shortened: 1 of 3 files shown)


Derkades:
I'd change it to
```php
if ($upload === false) {
    http_response_code(500);
    die("Failed to upload image" . $image->getError());
}
```
```php
http_response_code(400); // Bad request
die("No image uploaded");
```

Your current code tries to set the header "HTTP/1.1 500 Server Error" which doesn't make sense. The status code is sent before the headers in an HTTP response and messages belong in the response body.

samerton (approved):

PR #2655, opened 2022-03-31
Add mention suggestions in TinyMCE

### LINKED ISSUE #2351: @mention username suggestions
image

Aberdeener:
https://user-images.githubusercontent.com/26070412/160975475-f199eaae-30c0-4712-9ea3-3f2267dc3527.mov

This also cleans up `MentionsParser` a tiny bit, and makes the `parse` method static.
Additionally, to avoid slowing down the user list endpoint, I had to rework how user avatars are generated so that I can use data already queried in the API endpoint to generate the avatars.

~~We might want to consider adding a `&like=<string>` GET query param to the user list endpoint, because if the suggestions take > 200ms to show up it will be super annoying for the users. @Derkades what are your thoughts on this?~~

~~Also, do you want to keep it as `.includes(...)` or would `.startsWith(...)` be better for showing suggestions?~~
~~- Currently it is using `includes`, which is fine for my dev site since it has very few users - but if someone with many users used this, it would show many more than if we used `startsWith`. Thoughts?~~

(Ignore the random Collection changes, I meant to commit them separately)

Closes #2351

### DIFF
--- a/core/classes/Core/Input.php
+++ b/core/classes/Core/Input.php
@@ -53,10 +53,67 @@ class Input {
      *
      * @param Language $language Instance of language class to use for translation.
      * @param string $name Name of input field ID.
+     * @param bool $mentions Whether to enable mention autocompletion/parsing or not.
      */
-    public static function createTinyEditor(Language $language, string $name): string {
+    public static function createTinyEditor(Language $language, string $name, bool $mentions = false): string {
         $skin = defined('TEMPLATE_TINY_EDITOR_DARKMODE') ? 'oxide-dark' : 'oxide';
-        return "
+        $js = '';
+
+        if ($mentions) {
+            $js .= "
+                (function () {
+                    let flags = (function () {
+                        'use strict';
+                        tinymce.PluginManager.add('mentions', function (editor, url) {
+                            editor.ui.registry.addAutocompleter('autocompleter-mentions', {
+                                ch: '@',
+                                minChars: 2,
+                                columns: 1,
+                                fetch: function (pattern) {
+                                    return new tinymce.util.Promise(function (resolve) {
+                                        fetch('" . URL::build('/api/v2/users', 'avatars=true') . "')
+                                            .then((resp) => resp.json())
+                                            .then(function (data) {
+                                                let results = [];
+                                                const users = data.users.filter(p => p.username.toLowerCase().includes(pattern.toLowerCase()));
+    
+                                                for (const user of users) {
+                                                    results.push({
+                                                        value: '@' + user.username,
+                                                        text: user.username,
+                                                        icon: '<img src=\"' + user.avatar_url + '\">'
+                                                    });
+                                                }
+    
+                                                results.sort(function (a, b) {
+                                                    let x = a.text.toLowerCase();
+                                                    let y = b.text.toLowerCase();
+                                                    if (x < y) {
+                                                        return -1;
+                                                    }
+                                                    if (x > y) {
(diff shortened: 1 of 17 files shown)


Derkades:
Uh, are you sure the list users endpoint is safe to expose publicly? It contains quite a lot of user info. One example is profile fields, some of which might be private. I think adding a new endpoint (with a 'starts with' or 'contains' filter like you said) is a better idea.

samerton:
> Uh, are you sure the list users endpoint is safe to expose publicly? It contains quite a lot of user info. One example is profile fields, some of those might be private. I think adding a new endpoint (with a 'starts with' or 'contains' filter like you said) is a better idea.
> 
> Other than that, this is very exciting!

Agreed - I wonder if we can have an endpoint which supports both API key verification _and_ session authentication (based on user permissions or am I being too optimistic? 😄 )

Aberdeener:
Thanks guys - I agree Derk, did not consider the fact that profile fields are displayed in the users endpoint. What about "users-mentions" for the name?

Sam, *technically* that is possible with the CustomAuthEndpoint class, but now that Derk mentioned the profile fields, it's probably not the safest option

samerton:
> Sam, _technically_ that is possible with the CustomAuthEndpoint class, but now that Derk mentioned the profile fields, it's probably not the safest option

imo we should do this even with a new endpoint - I'm not sure a list of usernames needs to be accessible without authentication

Derkades:
It's a very common feature, spigotmc.org has the same thing

partydragen:
Add add a new endpoint/queries with is made just for this, users endpoint will also contain integrations data etc

Then you can also use that new endpoint/queries on messaging page where its also searching for users when writing their name

core/classes/Core/Input.php
```diff
@@ -53,10 +53,56 @@ public static function get(string $item) {
...
+                    let flags = (function () {
+                        'use strict';
+                        tinymce.PluginManager.add('mentions', function (editor, url) {
+                            editor.ui.registry.addAutocompleter('autocompleter-mentions', {
+                                ch: '@',
+                                minChars: 2,
+                                columns: 1,
+                                fetch: function (pattern) {
+                                    return new tinymce.util.Promise(function (resolve) {
+                                        fetch('" . URL::build('/queries/mention_users') . "&nickname=' + pattern)
```
samerton, line 74: Not sure if this will work if friendly URLs are enabled as the URL would be
```
<domain>/queries/mention_users&nickname=abc
```
ie there is no `?`

This could work as a "hack":
```php
URL::build('/queries/mention_users', 'nickname') . "='
```

modules/Forum/pages/forum/edit.php
```diff
@@ -120,6 +119,7 @@
         if ($validation->passed()) {
             // Valid post content
             $content = Output::getClean(Input::get('content'));
+            $content = MentionsParser::parse($user->data()->id, $content, URL::build('/forum/topic/' . $topic_id, 'pid=' . $post_id), ['path' => ROOT_PATH . '/modules/Forum/language', 'file' => 'forum', 'term' => 'user_tag'], ['path' => ROOT_PATH . '/modules/Forum/language', 'file' => 'forum', 'term' => 'user_tag_info', 'replace' => '{x}', 'replace_with' => Output::getClean($user->data()->nickname)]);
```
samerton, line 122: _Maybe_ worth adding a parameter to `parse` to disable sending alerts when editing a post?

samerton (approved):

PR #2662, opened 2022-04-02
Render post event

samerton:
Includes colour coded and interactive user mentions, fixed html code blocks, pagination fixes


custom/templates/DefaultRevamp/template.php
```diff
@@ -58,16 +58,7 @@ public function __construct($cache, $smarty, $language, $user, $pages) {
...
-        } else {
-            if (Cookie::exists('night_mode')) {
-                $darkMode = Cookie::get('night_mode');
-            }
-        }
-
-        if (isset($darkMode) && $darkMode == '1') {
+        if (defined('DARK_MODE') && DARK_MODE == '1') {
             $smartyDarkMode = true;
             define('TEMPLATE_TINY_EDITOR_DARKMODE', true);
```
Aberdeener, line 72: Do we need to define another constant?

modules/Forum/hooks/MentionsHook.php
```diff
@@ -35,6 +35,21 @@ public static function preEdit(array $params = []): array {
         return $params;
     }
 
+    public static function parsePost(array $params = []): array {
+        if (parent::validateParams($params, ['content'])) {
+            $params['content'] = preg_replace_callback(
+                '/\[user\](.*?)\[\/user\]/ism',
+                function($match) {
+                    $user = new User($match[1]);
+                    return '<a href="' . $user->getProfileURL() . '" data-poload="' . URL::build('/queries/user/', 'id=' . $user->data()->id) . '" class="user-mention" style="' . $user->getGroupClass() . '">@' . Output::getClean($user->data()->nickname) . '</a>';
```
Aberdeener, line 44: Awesome

modules/Forum/module.php
```diff
@@ -120,14 +120,43 @@ public function __construct(Language $language, Language $forum_language, Pages
...
         require_once ROOT_PATH . '/modules/Forum/hooks/DeleteUserForumHook.php';
         EventHandler::registerListener('deleteUser', 'DeleteUserForumHook::execute');
 
+        require_once ROOT_PATH . '/modules/Core/hooks/ContentHook.php';
         require_once ROOT_PATH . '/modules/Forum/hooks/MentionsHook.php';
         EventHandler::registerListener('prePostCreate', 'MentionsHook::preCreate');
         EventHandler::registerListener('prePostEdit', 'MentionsHook::preEdit');
         EventHandler::registerListener('preTopicCreate', 'MentionsHook::preCreate');
         EventHandler::registerListener('preTopicEdit', 'MentionsHook::preEdit');
+        EventHandler::registerListener('renderPost', 'ContentHook::purify');
```
Aberdeener, line 150: Random thought, all of these pre/post render "events" aren't exactly the same type of event as the other usage of the eventhandler class. 

To avoid confusion and separate them, could we create something like a "pipeline" class? Basically the same thing, but used for """under the hood""" events instead of normal events such as user register/verify etc.

core/classes/Core/Output.php
```diff
@@ -45,8 +45,6 @@ public static function getDecoded(?string $input): string {
      * @return string Purified string.
      */
     public static function getPurified(?string $input, bool $escape_invalid = false): string {
-        $input = Output::getDecoded($input);
```
Aberdeener, line 48: Reminder to us that we need to migrate all old data to decoded

modules/Core/hooks/ContentHook.php
```diff
@@ -0,0 +1,73 @@
...
+ *  Content hooks
+ */
+
+class ContentHook extends HookBase {
+
+    public static function codeTransform(array $params = []): array {
+        if (parent::validateParams($params, ['content'])) {
+            $domDocument = new DOMDocument();
+            $domDocument->loadHTML(mb_convert_encoding($params['content'], 'HTML-ENTITIES', 'UTF-8'));
+            echo $domDocument->nodeValue;
```
Aberdeener, line 16: Forgotten debug statement?

Aberdeener (approved):

modules/Core/hooks/ContentHook.php
samerton: Oops, good spot 😄

modules/Forum/module.php
samerton: When you register an event, it's possible to mark it as internal which distinguishes the event from ones the user should be aware of: https://github.com/NamelessMC/Nameless/blob/v2/core/classes/Events/EventHandler.php#L40

partydragen (approved):

modules/Forum/hooks/MentionsHook.php
partydragen: yea its very awesome!

PR #2667, opened 2022-04-05
Update swedish translations

ItsLynix:
I've done some minor updates to the translations as well as keep them up to date with new and changed NamelessMC features and APIs.

Great job guys, keep on evolving this software.
Even though, I myself no longer use NamelessMC, I can't help but still contribute to it. 

I wish you well and a good continuation of the project.

👍


ItsLynix:
Also, just another thing, I've noticed google plus still being in some of the language files and going around, google plus is a service that as you know, google has shut down, so shouldn't we remove it? Just a question.

PR #2669, opened 2022-04-05
Update Prism + error handler fix

Aberdeener:
Todo:
- [x] sql trace line highlighting is not working
- [x] make font a little smaller?

Note: the "Default" prismjs theme does not work with the line numbers plugin, so we are using Coy for the light theme


Aberdeener:
@samerton Fixed the issues we were having earlier.

- We use the "coy" theme for the error page. This is because it fits the style of the error page much better, but also because the line numbers plugin just... doesnt work on the "default" theme (as you saw haha)
- We use the "default" theme for everything TinyMCE related (except dark mode), since this is the builtin theme TinyMCE supports for Prism.

PR #2670, opened 2022-04-06
Template asset rework

Aberdeener:
This reworks how template assets are managed to make including them a lot easier and not as repetitive.
Might not be merged, just an idea i want to experiment with...

TODO:
- [ ] Use yarn
- [ ] Clean up TemplateAssets


partydragen (approved):

core/classes/Templates/AssetResolver.php
```diff
@@ -0,0 +1,147 @@
...
+        foreach ($asset['depends'] as $dependency) {
+            // Don't throw an exception if the dependency has already been resolved
+            // since it is a dependency, there is a high chance it has already been resolved anyway
+            if ($this->validateAsset($dependency, false)) {
+                $this->gatherAsset($dependency, parent::ASSET_TREE[$dependency], $css, $js);
+            }
+        }
+
+        // Unideal solution to get the correct light/dark Prism dependency
+        if ($name === parent::TINYMCE) {
```
samerton, line 73: Might be worth creating an issue for this now so we can come back to it.

Perhaps an asset definition can optionally be presented as an array with a `dark` and `light` option

modules/Forum/pages/forum/view_topic.php
```diff
@@ -259,7 +259,7 @@
         ]);
 
         if ($validate->passed()) {
-            $content = Output::getClean(Input::get('content'));
+            $content = Input::get('content');
```
samerton, line 262: Can we verify this has no effect on inserting code snippets which include HTML tags please?

package.json
```diff
@@ -0,0 +1,29 @@
+{
+  "name": "nameless",
+  "version": "2.0.0-pr13",
+  "repository": "https://github.com/NamelessMC/Nameless",
+  "license": "MIT",
```
samerton, line 5: Can we add `"private": true,` to here so it's not possible to accidentally publish this to npm

rewrite_test.php
```diff
@@ -14,8 +14,7 @@
 <head>
     <title>Rewrite Test &bull; NamelessMC</title>
 
-    <!-- Global CSS -->
-    <link rel="stylesheet" href="core/assets/css/bootstrap.min.css">
+    <link rel="stylesheet" href="core/assets/vendor/bootstrap/dist/js/bootstrap.min.js">
```
samerton, line 17: Is the change from `.css` to `.js` intentional here

samerton (changes requested):

rewrite_test.php
Aberdeener: Haha, no. Good catch

modules/Forum/pages/forum/view_topic.php
Aberdeener: Yes, this seems to still work. Editing and viewing works normally.

This is what is stored in the DB: 
```<pre class="language-markup"><code>&lt;p&gt;Hello, I am &lt;strong&gt;HTML&lt;/strong&gt;&lt;/p&gt;</code></pre>```

Screen Shot 2022-05-02 at 14 27 54

samerton (approved):

PR #2674, opened 2022-04-09
Ensure people don't overwrite staffcp urls

Derkades:
I still need to add a proper message


modules/Core/pages/panel/pages.php
```diff
@@ -23,6 +23,48 @@
 // Load modules + template
 Module::loadPage($user, $pages, $cache, $smarty, [$navigation, $cc_nav, $staffcp_nav], $widgets, $template);
 
+function validate_page_form() {
```
Aberdeener, line 26: While I like the idea of removing duplicate validation checks, I don't know if this is the most elegant way we can do it...

Derkades: Any other ideas? A collection of "form validators" in a separate directory somewhere?

Aberdeener: Perhaps - that seems good.
`FormValidator` class which contains a bunch of static functions seems alright. Ideally we use something like https://laravel.com/docs/9.x/validation#form-request-validation but can't get too ahead of ourselves 😉

modules/Core/pages/panel/pages.php
```diff
@@ -23,6 +23,48 @@
...
+        'page_title' => [
+            Validate::REQUIRED => true,
+            Validate::MIN => 2,
+            Validate::MAX => 255
+        ],
+        'page_url' => [
+            Validate::REQUIRED => true,
+            Validate::MIN => 2,
+            Validate::MAX => 255,
+            Validate::NOT_START_WITH => '/panel'
```
samerton, line 38: We need to consider subdirectory installations here

partydragen: Unfriendly urls don't even require to begin with / , so should make sure page path always begins with a /

modules/Core/pages/panel/pages.php
Derkades: should that be included in this PR, or can it be done later?

modules/Core/pages/panel/pages.php
Derkades: I believe the entered custom page path wouldn't contain subdirectories, so I don't think we need to do anything special to support subdirectory installations?

partydragen: yea you're right, we don't enter subdir there

modules/Core/pages/panel/pages.php
Aberdeener: let's do that later:tm:

for now, let's keep the repeated validation in the file.

PR #2678, opened 2022-04-11
Dark Mode fixes

TeemoCell:


custom/panel_templates/Default/assets/css/custom.css
```diff
@@ -44,9 +44,9 @@ body {
 .dark .dropdown-item,
 .dark .dropdown-menu,
 .dark .topbar .dropdown-list .dropdown-item:active {
-    background-color: var(--dark-color-5);
+    background-color: #161c25;
```
Derkades, line 47: Instead of replacing the variable with a hex color, why not update the variable? Doing that in the first place might have prevented this issue

TeemoCell: bec the color is used in different classes which would change it for everything that use that variable.

PR #2658, opened 2022-03-31
Use i18n for translations

Aberdeener:
workflow:
- ~~replace all variables in current language files~~
	- ~~update code accordingly to use new system variables array~~
- ~~check for english strings in non englishuk languages & remove~~ 
- ~~compile all languages as json, delete old language files~~
- ~~remove unused strings (derk script)~~
- ~~add `short_code` column to language table & use for lookup in db~~
- ~~check language files for un-replaced variables in code~~
- ~~move all links to the link variable style~~
- ~~update user integrations language to json when merged~~

todo:
- ~~deal with html in language files, `<i...`~~
- ~~i marked areas we need to update things with `// TODO: new language system`~~
- ~~rework `set` method~~
- ~~timeago in words is mega broken~~
- ~~use the i18n plural builtin stuff where applicable (nvm, it only support very basic plurals)~~
- ~~`Failed to initialize the editor as the document is not in standards mode. TinyMCE requires standards mode.` error in quickreply on forum~~
- ~~page load time not working~~

PR #2688, opened 2022-04-17
Use proxy headers, properly, securely

Derkades:
Fixes a security issue, NamelessMC used the X-Forwarded-For header without verifying whether the source was trustworthy which means a user could impersonate any IP address.
 
Also fixes an issue where NamelessMC didn't use the proxy port but the backend server port when building URLs.

* Removed the old `allowedProxies` option in config.php which wasn't used
* Added three new methods to Util.php for getting the correct protocol, port and remote address
* Added new `core/trustedProxies` option, which is an optional list of CIDR network ranges to trust.
* Removed `getIP` from User.php because this function has nothing to do with the User object (replaced with `Util::getRemoteAddress`)
* Uses `Symfony\Component\HttpFoundation\IpUtils` because I am lazy but also do not trust myself to check network addresses securely. IP addresses are complicated!


Aberdeener:
We will have to remember to update the upgrade guide with these method removals

core/classes/Core/Util.php
```diff
@@ -128,6 +129,80 @@ public static function isExternalURL(string $url): bool {
         return !(str_replace('www.', '', rtrim(self::getSelfURL(false), '/')) == str_replace('www.', '', $parsed['host']));
     }
 
+    /**
+     * Checks whether the client making the request is a trusted proxy. If not,
+     * abruptly aborts the request using die().
+     */
+    private static function ensureTrustedProxy(): void {
+        $trustedProxies = Config::get('core/trustedProxies');
```
Aberdeener, line 137: Are we leaving this new `trustedProxies` configuration setting undocumented? Would make sense since it is for more advanced setups to not include it by default in the config...

modules/Core/pages/panel/general_settings.php
```diff
@@ -218,19 +218,6 @@
                 $errors = [$language->get('admin', 'config_not_writable')];
             }
 
-            /*
```
Aberdeener, line 221: We will have to remember to remove the `allowed_proxies` and `allowed_proxies_info` translations in #2658

core/classes/Core/Util.php
```diff
@@ -143,26 +218,21 @@ public static function getSelfURL(bool $protocol = true): string {
         }
 
         // https and www checks
-        if (self::isConnectionSSL()) {
-            $proto = 'https://';
-        } else {
-            $proto = 'http://';
-        }
+        $proto = self::getProtocol() . '://';
```
Aberdeener, line 221: Might read easier if we change the param from `protocol` to `show_protocol`, then we can rename `proto` to `protocol`

Aberdeener (changes requested):

core/classes/Core/Util.php
Derkades: Good point, while it might sound advanced I think potentially many users need this option, especially with the prevalence of CloudFlare. Maybe it could be part of the installer.

Derkades:
If you really wanted you could do that by adding 0.0.0.0/0 as a trusted proxy

AToska21:
> If you really wanted you could do that by adding 0.0.0.0/0 as a trusted proxy

That should be at least documented, but I think it would be easier to just add an option.

core/classes/Core/User.php
```diff
@@ -229,23 +229,6 @@ public function getGroupStyle(): string {
         return $css;
     }
 
-    /**
-     * Get the logged in user's IP address.
-     *
-     * @return string Their IP.
-     */
-    public function getIP(): string {
```
Aberdeener, line 237: @Derkades what would you think about keeping this method and just forwarding it to `Util::getRemoteIp()`? `$user->getIP()` reads more fluently 🤔

Derkades: But it has nothing to do with the User object, instead with the current connection. Someone might do
```
$user = new User(4)
$user->getIP()
```
And expect to get the IP address for user 4, which is not the case.

Aberdeener: Makes sense, thank you for the explanation!

core/classes/Core/Util.php
Derkades: Documented now: https://docs.namelessmc.com/en/trusted-proxies

core/classes/Core/User.php
partydragen: public function getIP(): string {
    return ($this->exists() && !$this->isLoggedIn() ? $this->data()->ip : Util::getRemoteIp())
}
???

core/classes/Core/Util.php
```diff
@@ -128,41 +129,140 @@ public static function isExternalURL(string $url): bool {
...
+        $trustedProxies = self::getTrustedProxies();
+
+        if (count($trustedProxies) === 0) {
+            die('Received proxy header but no trusted proxies are configured. Please see <a href="https://docs.namelessmc.com/trusted-proxies">https://docs.namelessmc.com/trusted-proxies</a> for more information.');
+        }
+
+        $trusted = false;
+
+        foreach ($trustedProxies as $trustedProxy) {
+            if (IpUtils::checkIp($_SERVER['REMOTE_ADDR'], $trustedProxy)) {
```
Aberdeener, line 171: I'm probably missing something here, but shouldn't we be using `self::getRemoteAddress()` here, instead of `$_SERVER['remote_address']`? (I know that `getRemoteAddress` will call `ensureTrustedProxy`, but we could add a param to skip the trusted proxy check)

Aberdeener (approved):

PR #2689, opened 2022-04-20
German Language Update

enno123:
Added missing translations


Derkades:
Indeed, language contributions unfortunately cannot be accepted at this time because all of the old language files have been converted to a new system. The new language files have so many changes to the strings that it'd be way to much work to merge contributions to the old language files. I'll create a pinned issue so it's more obvious.

When the i18n PR is merged, you'll be able to copy paste the translations you've done here to https://translate.namelessmc.com/

PR #2693, opened 2022-04-23
Fix long dropdown names overflowing

KaspianDev:
before
after

### DIFF
--- a/custom/panel_templates/Default/assets/css/custom.css
+++ b/custom/panel_templates/Default/assets/css/custom.css
@@ -396,7 +396,7 @@ body {
 }
 
 .collapse-item {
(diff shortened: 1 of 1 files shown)


custom/panel_templates/Default/assets/css/custom.css
```diff
@@ -395,6 +395,10 @@ body {
     max-width: none;
 }
 
+.collapse-item {
+    white-space: normal !important;;
```
samerton, line 399: Really small one but there are two `;` characters at the end of this line

samerton (changes requested):
Thanks for this - just a tiny comment!

samerton (approved):

PR #2694, opened 2022-04-23
Fix small sidebar not having scrollbar

KaspianDev:


custom/panel_templates/Default/assets/css/custom.css
```diff
@@ -351,6 +351,12 @@ body {
     .sidebar {
         width: 17rem !important;
     }
+    
+    .sidebar.toggled {
+        width: 7rem !important;
```
samerton, line 356: This seems to force the sidebar to be always open on mobile (it should be wrapped in `@media (min-width: 768px)` to only apply to bigger screens)

samerton (changes requested):

samerton:
Thanks for this - just had a look on the latest dev code on my side and the sidebar appears to have a scrollbar already?

KaspianDev:
> Thanks for this - just had a look on the latest dev code on my side and the sidebar appears to have a scrollbar already?

Problematic on firefox

custom/panel_templates/Default/assets/css/custom.css
KaspianDev: it is in media query

Derkades:
Indeed, there was a bug where the collapsed sidebar didn't have a scrollbar in firefox

custom/panel_templates/Default/assets/css/custom.css
samerton: Thanks, this is me being lazy and not checking the above lines

samerton (approved):

PR #2698, opened 2022-04-23
Add ability to clone group

partydragen:
Adds the ability to clone group with also executes a event so modules can listen for it to copy their group permissions

Discussion: https://github.com/NamelessMC/Nameless/discussions/2129


modules/Core/hooks/CloneGroupHook.php
```diff
@@ -0,0 +1,31 @@
...
+        if ($permissions->count()) {
+            $permissions = $permissions->results();
+
+            $inserts = [];
+            foreach ($permissions as $permission) {
+                $inserts[] = '('.$permission->page_id.',' . $new_group_id . ',' . $permission->view . '),';
+            }
+
+            $query = 'INSERT INTO nl2_custom_pages_permissions (page_id, group_id, view) VALUES ';
+            $query .= implode('', $inserts);
```
samerton, line 26: Alternatively you could implode using the `,` character as the glue and then remove the `,` from line 22 and `rtrim` from line 28 😄

modules/Core/pages/panel/groups.php
```diff
@@ -294,78 +298,213 @@
...
+                            'html' => $language->get('admin', 'html_maximum')
+                        ]);
 
-                        try {
-                            $queries->update('groups', $group->id, ['permissions' => $perms_json]);
+                        if ($validation->passed()) {
+                            try {
+                                if (isset($_POST['default']) && $_POST['default'] == 1) {
+                                    $default = 1;
+                                    $cache->setCache('default_group');
```
samerton, line 341: I assume this was already here but it should be removed since we do it on line 381 anyway

modules/Core/hooks/CloneGroupHook.php
partydragen: Good idea!

samerton (approved):

PR #2712, opened 2022-04-24
Use standard Authorization header

Derkades:
* Replace non standard `X-API-Key` header with standard `Authorization: Bearer` header
* Retire specific error for invalid API key, always return the same unauthorized error. Different authentication implementations returning different error codes is annoying for clients, especially if modules decide to add their own authentication schemes and client applications receive an unexpected status code
* Remove CustomAuthEndpoint, it is unnecessary because a custom authentication endpoint can just extend EndpointsBase directly. 
* Make EndpointsBase an abstract class, it should never be instantiated directly.


Aberdeener:
Can you make sure to update the documentation for this? Thank you

PR #2721, opened 2022-04-27
Only read last part of error log

Derkades:
Before:
before
After:
after


partydragen:
Still not sure about this as fields can contain field inputs with might be personal Derk told its been removed in newer php versions i only just remember seeing my emails and passwords in my errors before

and that file path is full path with many people want to hide with sometimes include the guys real name @samerton

Derkades:
This PR was mostly a fix for the error log page, maybe discussion about including logs in debug reports should be in a separate issue?

PR #2728, opened 2022-05-01
don't use CDN for semantic lato font

PadowYT2:
When switched to Fomantic, I didn't know about 6b34c4abe61fcd12863bf840c4cb6350f2125922 commit


Derkades:
maybe the font override could be placed in the custom.css file to prevent issues like this. Especially when we no longer include the fomantic css file in the repo at all after #2670

Aberdeener:
This is incompatible with #2670, either we have to revert this change to make it compatible, or find a different solution since we cannot edit files provided by yarn/npm

Derkades:
unfortunately I don't know how to make this compatible.. Does npm allow doing a regex replace in a file after `npm install`?

Aberdeener:
Heh, I guess so but that sounds wildly janky so I don’t want to do that

PR #2729, opened 2022-05-02
`InteractsWithDatabase` trait cleanup

Aberdeener:
- adds docblocks
- deprecates the `selectQuery` and `createQuery` (uses 1 `query` method instead, which only calls the `fetchAll` if it is a select statement)
- allows passing `true` or `false` to where clauses instead of `1` or `0` respectfully
- allows the `update(...)` method to take an array for the where clause instead of forcing an id
- removes the need for `'='` as the operator in a where clause array. arrays of length 2 will be inferred as =
- misc renames of methods & removal of unused/legacy ones


Derkades (approved):
get it in as soon as possible so it can get some final testing before the pr13 release?

Aberdeener:
> get it in as soon as possible so it can get some final testing before the pr13 release?

Agree, going to wait for @samerton for the final call though

samerton (approved):

PR #2730, opened 2022-05-02
use CDN for semantic lato font

PadowYT2:
It broke the whole website & #2670


PadowYT2:
Nevermind, I see the new commit

Issue #2660, opened 2022-04-02
Cookie design fail


PadowYT2:
### Describe the issue

On mobile the dark popup of the "Cookies" is pure dark and dark text on dark popup

### NamelessMC Version

From commit 76713ba

### To Reproduce

1. Go on mobile or turn on "Toggle device toolbar" in browser console
2. Reload the page
3. See the problem

### Expected Behaviour

Should be white on dark

### Screenshots

![https://cdn.padow.ru/PL5s](https://cdn.padow.ru/PL5s)

samerton:
Unable to reproduce on latest v2
<img width="140" alt="Screenshot 2022-04-02 at 13 51 51" src="https://user-images.githubusercontent.com/1812544/161384181-b1cff02b-8e3c-4b2d-bffd-ee5d9fa0f118.png">

PadowYT2:
What browser & resolution do you have?
I have
* 390x844 (example resolution, there are a lot)
* Google Chrome 99.0.4844.82 (Official Build) (64-bit)
https://user-images.githubusercontent.com/71085027/161384578-b4b71d32-800f-49f0-897b-d1f21d420087.mp4

samerton:
I've tried with Chrome + Safari on a variety of resolutions, including the one you mention.

Could you please right click -> inspect element on the cookie text to see what's determining the text colour? The following CSS should apply:
```css
.cc-revoke.cc-bottom.cc-right.cc-animate {
    color: #fff;
}
```
(from https://github.com/NamelessMC/Nameless/blob/4b4fe73b552f95d4b0e89e7b88acac23c4bc82c9/custom/templates/DefaultRevamp/css/custom.css#L601-L603)

PadowYT2:
> I've tried with Chrome + Safari on a variety of resolutions, including the one you mention.
> 
> Could you please right click -> inspect element on the cookie text to see what's determining the text colour? The following CSS should apply:
> 
> ```css
> .cc-revoke.cc-bottom.cc-right.cc-animate {
>     color: #fff;
> }
> ```
> 
> (from
> 
> https://github.com/NamelessMC/Nameless/blob/4b4fe73b552f95d4b0e89e7b88acac23c4bc82c9/custom/templates/DefaultRevamp/css/custom.css#L601-L603
> 
> )

By removing background-color, you can see it is dark
![https://cdn.padow.ru/flBn](https://cdn.padow.ru/flBn)

samerton:
[...]

PR #2740, opened 2022-05-08
Fixes dark text on Cookies popup

PadowYT2:
As on phones there's no animate and the animate class is removed, so the color is not declared for the phone. An actual fix for #2660


samerton:
Nice catch, thanks

PR #2743, opened 2022-05-11
Merge DB and DB_Custom to one class

Derkades:
~~Merge after #2729 , change base branch from `db-update` to `v2` before merging.~~ ~~Highly experimental, didn't test installation yet.~~

~~Checks are failing because of an issue in the other PR, not sure how to resolve it~~

The website works fine (no breaking changes) and other usages of DB_Custom have been updated. The web installer works fine, didn't test the other installers but the change is exactly the same so I expect them to work as well.


core/classes/Database/DB.php
```diff
@@ -8,29 +8,411 @@
...
+        }
+
+        if (!in_array($operator, $operators)) {
+            throw new InvalidArgumentException("Invalid SQL operator: '{$operator}'");
+        }
+
+        return [$field, $operator, $value];
+    }
+
+    public static function getCustomInstance(string $host, string $database, string $username,
```
Aberdeener, line 390: `connection` might be a more fluent word here... what do you think?

Maybe not actually, since then it wouldn't match with `getInstance` 🤔

Derkades: I do like getConnection, but that would be a breaking change (currently, only usage of the custom class changed)

PR #2747, opened 2022-05-12
(Optional) Auto language detection for guests

### LINKED ISSUE #2196: Automatically set user language
I was thinking that we could implement a dynamic language system in NamelessMC, let me explain:
When the user (Anonymous) enters the page, the language that is selected as the default in their browser is identified.
and when the user who is registered enters the page, the language that she has selected is selected, in case she has done so

🤔 I think I know how to do it

Aberdeener:
Will close #2196 if approved.

If a user is logged out, and has language detection enabled (via a cookie), or if the cookie doesn't exist, then we attempt to get the browser locale and see if we have a matching language.

Not totally happy with "Auto language" as the button, so ideas are appreciated.

### DIFF
--- a/core/init.php
+++ b/core/init.php
@@ -219,6 +219,15 @@ if ($page != 'install') {
     }
 
     if (!$user->isLoggedIn() || !($user->data()->language_id)) {
+        // Attempt to get the requested language from the browser if it exists
+        // and if the user has enabled auto language detection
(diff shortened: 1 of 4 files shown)


PadowYT2 (commented):
You could have the feature for enabling/disabling in StaffCP (if enabled next suggestion, if not, don't make the auto selection at all) & UserCP above language selection

samerton (approved):

zJerino:
ª

PR #2752, opened 2022-05-18
Manually Updated Lines 1~252 for Language zh_TW

ichenglin:
I manually read through lines 1 to 252 and updated some sentences to make them read more naturally, and also replaced some incorrectly used punctuations.

Example:
- "." is not a valid period punctuation, instead it should be replaced with "。"

Disclaimer:
- I only edited lines 1~252, this means that any lines after remains unchanged, which still includes incorrect punctuations.


Aberdeener:
@Dreta hi dreta, can you review this?

ichenglin:
Just realized I accidently removed the "r" from "server" in line 238, it was not part of my translation.

image

PadowYT2 (changes requested):
This should not be done at all, weblate could run into issues. You **must** translate https://translate.namelessmc.com

Aberdeener:
Manual changes to the files is fine Padow, but yep Weblate is a nicer
experience for the translator - not a requirement.

ichenglin:
I did checked the Weblate link before I begin translating, however I still prefer direct file edits over Weblate. Translated strings could easily be compared around to be written in the same style, plus the script editor I am using provide an efficient way of replacing the incorrect punctuations.

Dreta (approved):
I don’t think there’s a problem with this.

Derkades:
Weblate is configured to wait 48 hours before committing changes. As long as no changes have been made to the language in the last 48 hours, it should be fine

ichenglin:
Fixed the typo of "serve" to "server" in commit 38af18f.

Derkades:
Thanks! Even though you said you weren't finished yet I have already merged this for now, just to avoid any future conflicts.

PR #2754, opened 2022-05-18
Final Update for Language zh_TW

ichenglin:
Replaced all incorrect punctuations along with a slight update of translation. This should be the final update for now and no commits would be made anytime soon.


ichenglin:
Have to resolve conflict before pull request.

PR #2756, opened 2022-05-18
Event additions

supercrafter100:
Added 3 more events:

- userGroupAdded
- userGroupRemoved
- topicReply


core/classes/Core/User.php
```diff
@@ -718,6 +725,13 @@ public function removeGroup(?int $group_id): bool {
             ]
         );
 
+        EventHandler::executeEvent("userGroupRemoved", [
```
Aberdeener, line 728: Single quotes please

core/classes/Core/User.php
```diff
@@ -181,6 +181,13 @@ public function addGroup(int $group_id, int $expire = 0, $group_data = null): bo
             $this->_groups[$group_id] = new Group($group_data);
         }
 
+        EventHandler::executeEvent("userGroupAdded", [
```
Aberdeener, line 184: Single quotes please

Aberdeener (changes requested):

modules/Forum/module.php
```diff
@@ -59,7 +59,7 @@ public function __construct(Language $language, Language $forum_language, Pages
         EventHandler::registerEvent('newTopic',
             $this->_forum_language->get('forum', 'new_topic_hook_info'),
             [
-                'uuid' => $this->_language->get('admin', 'uuid'),
+                'user_id' => $this->_language->get('admin', 'user_id'),
```
Aberdeener, line 62: Why was this changed?

Aberdeener (approved):

modules/Forum/module.php
supercrafter100: Uuid wasn't provided when the event was triggered but the user id.

supercrafter100: As you can see on line 205 here https://github.com/NamelessMC/Nameless/blob/v2/modules/Forum/pages/forum/new_topic.php

modules/Forum/pages/forum/view_topic.php
```diff
@@ -293,6 +293,24 @@
                 'topic_reply_date' => date('U')
             ]);
 
+            // Fire event
+            EventHandler::executeEvent('topicReply', [
```
partydragen, line 297: Can you include topic id and post id

partydragen: Post id is not forum id xD

PR #2759, opened 2022-05-19
patch: small fixes

PadowYT2:
* Fix vulnerability
```
select2: 4.0.0
Known security vulnerability: 1
Security advisory: 0
Exploits: unavailable
Highest severity: medium
Recommendation: 4.1.0-rc.0
Dependency Analytics Plugin [Powered by Snyk]
```
* Fix formatting
* Fix wrong icons
* Fix color on tooltip background on graph

PR #2760, opened 2022-05-19
Remove vulnerability

PadowYT2:
select2: 4.0.0
Known security vulnerability: 1
Security advisory: 0
Exploits: unavailable
Highest severity: medium
Recommendation: 4.1.0-rc.0
Dependency Analytics Plugin [Powered by Snyk]

*From #2759*


samerton:
I was going to suggest that we don't use beta/release candidate versions, but looks like it's been a long while since that project received any sort of update so I'm happy to go with it in this case.

PR #2770, opened 2022-05-20
DB migration system

Aberdeener:
### What and Why
This PR converts our database table initialization/scaffolding to a database migration library called [Phinx](https://book.cakephp.org/phinx/0/en/intro.html), which is maintained by the CakePHP team.

"Database migrations" is a way to make changes to a database schema in a similar fashion to normal version control systems.

For example, this PR includes a file named `create_mc_servers_table.php`, which creates the `nl2_mc_servers` table. In the future we might want to make a change to this table - like add a new column.

Previously, we would have to open the Queries class and alter the raw SQL string which contains the table creation command, and make changes there. Then, we would have to add code to the upgrade script to reflect these changes again. Duplicating code like this gives room for error, and is just plain annoying.

With Phinx, all we have to do is run a command to create a new migration file (`vendor/bin/phinx create AddVersionColumnToMcServersTable`), and then programmatically make the change to the database in that file.
The new file will be added to the `core/migrations/` directory, and will be executed automatically during site installation as well as during updates (Phinx keeps a log of completed migrations, so only unexecuted ones will be ran during site upgrades).

This is beneficial for a couple reasons:
- No repeating SQL code anywhere
- Fluent table builder API makes modifying/creating tables super easy and not annoying

### Notes
- Migration files are stored in the `core/migrations/` directory, as well as the `phinx.php` configuration file.
- See the [Phinx docs](https://book.cakephp.org/phinx/0/en/migrations.html#creating-a-new-migration) for how to create migration files, but keep in mind since our `phinx.php` file is not in the root directory, we need to specify the `--configuration` option in any Phinx commands:
    - `vendor/bin/phinx migrate -c core/migrations/phinx.php`
    - `vendor/bin/phinx create CreateExampleTable -c core/migrations/phinx.php`
- Since it is bad practice to use the PHP `exec(...)` command, I've created the `PhinxAdapter` class to let us call Phinx methods without using the command line. This class currently has only a `migrate` method right now, but could be expanded in the future. Obviously, when making migrations you should use the command line. This class is only for NamelessMC to call when installing/upgrading a site.


core/migrations/phinx.php
```diff
@@ -0,0 +1,20 @@
...
+        'migrations' => __DIR__,
+    ],
+    'environments' => [
+        'nameless' => [
+            'adapter' => 'mysql',
+            'host' => $conf['host'],
+            'name' => $conf['db'],
+            'user' => $conf['username'],
+            'pass' => $conf['password'],
+            'port' => $conf['db'],
```
Derkades, line 17: do you mean `$conf['port']`

Aberdeener: Thank you

core/classes/Core/User.php
```diff
@@ -243,7 +243,7 @@ public function getIP(): string {
             return $_SERVER['HTTP_X_FORWARDED_FOR'];
         }
 
-        return $_SERVER['REMOTE_ADDR'];
+        return $_SERVER['REMOTE_ADDR'] ?? '';
```
mdcfe, line 246: ?? ''

mdcfe (approved):

core/classes/Core/User.php
Aberdeener: I agree.

partydragen (approved):

core/classes/Database/DatabaseInitializer.php
```diff
@@ -0,0 +1,631 @@
...
+
+    private function initializeIntegrations(): void {
+        $this->_queries->create('integrations', [
+            'name' => 'Minecraft',
+            'enabled' => true,
+            'can_unlink' => false,
+            'required' => false,
+        ]);
+
+        // TODO: should this be in the DiscordIntegration module...?
```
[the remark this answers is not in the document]
Aberdeener: Will come back to this soon:tm: since we don't have an eloquent way to handle this right now

Aberdeener:
If at least 1 other person could either do a fresh install on this branch and/or use the CLI install to make sure Phinx is working on their machine, that would be fantastic!
Don't want to merge this in the off chance it doesn't play nicely on different OS's etc

PR #2771, opened 2022-05-20
Replace links

PadowYT2:


modules/Core/pages/panel/api.php
```diff
@@ -233,7 +233,7 @@
             'API' => $language->get('admin', 'api'),
             'PAGE' => PANEL_PAGE,
             'API_INFO' => $language->get('admin', 'api_info', [
-                'pluginLinkStart' => '<a href="https://plugin.namelessmc.com" target="_blank">',
```
samerton, line 236: I think this should stay as `plugin.namelessmc.com`

README.md
```diff
@@ -17,7 +17,7 @@ The following list is a brief summary of the features available in v2 pre-releas
 - 🎮 Minecraft integration
   - Display your Java edition or Bedrock servers statuses
   - Account verification using MCAssoc or the NamelessMC plugin
-  - [In-game plugin](https://www.spigotmc.org/resources/nameless-plugin-for-v2.59032)
+  - [In-game plugin](https://www.spigotmc.org/resources/59032)
```
samerton, line 20: Can this change to `plugin.namelessmc.com`?

samerton (changes requested):

modules/Core/pages/panel/api.php
PadowYT2: But then, in README, should it be that?

README.md
Derkades: The plugin will probably need to move away from spigot for the next update so this is a good idea

PR #2758, opened 2022-05-19
Remove unused stuff

PadowYT2:
* Remove contact email fully, wasn't done in #2603
* Remove DefaultRevamp/img
https://canary.discord.com/channels/246705793066467328/434751012428054530/976745008076030002

PR #2774, opened 2022-05-21
patch: remove contact & img

PadowYT2:
*From #2758*


core/installation/steps/site_configuration.php
```diff
@@ -89,7 +89,6 @@
                 <div class="sixteen wide mobile twelve wide tablet ten wide computer column">
                     <div class="ui form">
                         <?php create_field('text', $language->get('installer', 'site_name'), 'sitename', 'inputSitename', getenv('NAMELESS_SITE_NAME') ?: ''); ?>
-                        <?php create_field('email', $language->get('installer', 'contact_email'), 'incoming', 'contact_email', getenv('NAMELESS_SITE_CONTACT_EMAIL') ?: ''); ?>
```
partydragen, line 92: Why remove this? does this still not need to be entered during installation

custom/panel_templates/Default/core/general_settings.tpl
```diff
@@ -47,10 +47,12 @@
                                                placeholder="{$SITE_NAME_LABEL}" value="{$SITE_NAME}">
                                     </div>
                                     <div class="col-md-6">
-                                        <label for="inputContactEmail">{$CONTACT_EMAIL_ADDRESS}</label>
```
partydragen, line 50: Same with this, they still need to be able to edit it?

modules/Core/pages/panel/general_settings.php
```diff
@@ -245,8 +234,8 @@
 }
 
 // Get form values
-$contact_email = $queries->getWhere('settings', ['name', 'incoming_email']);
-$contact_email = Output::getClean($contact_email[0]->value);
+$incoming_email = $queries->getWhere('settings', ['name', 'incoming_email']);
```
partydragen, line 237: Why you even keep this when you removed the setting? from my understanding the setting should still exist as incoming email

modules/Core/pages/user/settings.php
```diff
@@ -388,8 +388,7 @@
                                 Validate::REQUIRED => $language->get('user', 'password_required') . '<br />'
                             ],
                             'email' => [
-                                Validate::REQUIRED => $language->get('user', 'email_required') . '<br />',
-                                Validate::EMAIL => $language->get('general', 'contact_message_email') . '<br />'
```
partydragen, line 392: This validation error should still be there, might be wrong language term

PR #2788, opened 2022-05-27
Copy files using javascript to support non-Unix platforms

Derkades:


package.json
```diff
@@ -22,9 +22,10 @@
     "popper.js": "^1.16.1",
     "select2": "4.1.0-rc.0",
     "tinymce": "5.10.3",
-    "toastr": "^2.1.4"
+    "toastr": "^2.1.4",
+    "fs-extra": "^10.0.0"
```
samerton, line 26: Can this be listed alphabetically for easy diff comparison following a `yarn upgrade`?

PR #2795, opened 2022-05-31
Move Util::isModuleEnabled(...) before cache check

### LINKED ISSUE #2773: Discord hook broken
The core module checking if DiscordHook class is loaded before the discord module is loaded so the discord hook is not able to execute

**Another issue:**
The discord module enabled check is being cached in the hooks lists with wont be able to update if the module is either disabled or enabled with meand the cache value will be outdated
The module enabled check should be moved outside of the cache

supercrafter100:
Will close #2773

### DIFF
--- a/core/classes/Core/Util.php
+++ b/core/classes/Core/Util.php
@@ -643,12 +643,12 @@ class Util {
             return true;
         }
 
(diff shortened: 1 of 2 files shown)


Aberdeener:
Simply astounding work

PR #2803, opened 2022-06-01
Make a toast instead of an alert

### LINKED ISSUE #2800: Copying debug link
### Describe the issue

Sometimes copying debug link just doesn't work

### NamelessMC Version

Development version

### To Reproduce

_No response_

### Expected Behaviour

_No response_

### Screenshots

https://cdn.padow.ru/DcaW

### Additional Information

_No response_

PadowYT2:
Closes #2800

### DIFF
--- a/error.tpl
+++ b/error.tpl
@@ -154,9 +154,7 @@
 </body>
 
 <style>
-    {
-        literal
-    }
+    {literal}
 
     * {
         box-sizing: border-box
@@ -244,9 +242,7 @@
         left: 20px;
     }
 
-        {
-        /literal
-    }
+    {/literal}
 </style>
 
 <script src="{$JQUERY}"></script>


error.tpl
```diff
@@ -67,227 +66,254 @@
...
+    }
+
+    /* Move icon of the toast to right */
+    .toast .move-right {
+        position: absolute;
+        left: 20px;
+    }
+
+        {
+        /literal
```
Derkades, line 248: did your linter do this? `{literal}` and `{/literal}` would be more readable.

Derkades:
Seems to work, thanks!

custom/panel_templates/Default/template.php
```diff
@@ -155,11 +155,11 @@ public function onPageLoad() {
                             AssetTree::DATATABLES
                         ]);
                         $this->addCSSFiles([
-                            (defined('CONFIG_PATH') ? CONFIG_PATH : '') . '/custom/panel_templates/Default/assets/css/fomantic.toast.min.css' => [],
```
Aberdeener, line 158: Now that this is being used in the panel template and frontend template, we should probably use yarn for it

PR #2830, opened 2022-06-06
Discord webhook formatter

partydragen:
Allow modules to make their own format for their discord plain message or embed

The webhook execution should also not display text it should only contain data, the formater itself should make the text to send to discord


modules/Discord Integration/module.php
```diff
@@ -33,6 +33,17 @@ public function __construct(Language $language, Pages $pages, Endpoints $endpoin
         GroupSyncManager::getInstance()->registerInjector(new DiscordGroupSyncInjector);
 
         Integrations::getInstance()->registerIntegration(new DiscordIntegration($language));
+        
+        // Hooks
+        EventHandler::registerEvent('discordWebhookFormatter',
+            'Discord webhook formatter',
+            [
+                'data' => 'Event data',
+                'format' => 'The format witch being sent to the discord webhook'
```
Aberdeener, line 42: which*

modules/Core/hooks/DiscordFormatterHook.php
```diff
@@ -0,0 +1,47 @@
...
+ *  Discord formatter hook
+ */
+
+class DiscordFormatterHook extends HookBase {
+
+    public static function format(array $params = []): array {
+        $data = $params['data'];
+        $format = $params['format'];
+
+        if ($data['event'] == 'registerUser') {
```
Aberdeener, line 16: It would be great if these weren't hard coded here. 

Can we utilize the event you made in this PR within the core module to format them?

modules/Discord Integration/module.php
```diff
@@ -33,6 +33,17 @@ public function __construct(Language $language, Pages $pages, Endpoints $endpoin
...
+            ],
+            true,
+            true
```
Aberdeener, line 45: Weird spacing here, might just be GitHub

Aberdeener (changes requested):

modules/Core/module.php
```diff
@@ -321,7 +321,9 @@ public function __construct(Language $language, Pages $pages, User $user, Naviga
                 'user_id' => $language->get('admin', 'user_id'),
                 'username' => $language->get('user', 'username'),
                 'email_address' => $language->get('user', 'email_address')
-            ]
+            ],
+            false,
+            true
```
samerton, line 326: Why are these now marked as internal and therefore hidden from admins?

modules/Discord Integration/hooks/DiscordFormatterHook.php
```diff
@@ -0,0 +1,47 @@
...
+ *  Discord formatter hook
+ */
+
+class DiscordFormatterHook extends HookBase {
+
+    public static function format(array $params = []): array {
+        $data = $params['data'];
+        $format = $params['format'];
+
+        if ($data['event'] == 'registerUser') {
```
samerton, line 16: How does this work for events outside of registerUser or createAnnouncement? Can this be extended by a module to add new cases?

modules/Core/module.php
partydragen: Those are internal events, Those was discussed in discord to make they internal and aber told to include in this pr

modules/Discord Integration/hooks/DiscordFormatterHook.php
partydragen: Other modules can add their own event listeners like that ^^ all events have events names like that so they can format their own events or other events

i also tested this for my own modules to make my own format for my events and that worked fine just like that ^^

modules/Core/module.php
samerton: I don't think they should be internal. It might be useful for admins to post this type of info to a Discord channel for example for audit purposes, also for use in modules such as Websend to delete a rank from an in game user when their account gets deleted

modules/Core/module.php
```diff
@@ -364,7 +366,9 @@ public function __construct(Language $language, Pages $pages, User $user, Naviga
                 'punisher_id' => $language->get('admin', 'punisher_id'),
                 'reason' => $language->get('admin', 'reason'),
                 'ip_ban' => $language->get('admin', 'ip_ban'),
-            ]
+            ],
+            false,
+            true
```
samerton, line 371: This shouldn't be internal either

modules/Core/module.php
```diff
@@ -373,7 +377,9 @@ public function __construct(Language $language, Pages $pages, User $user, Naviga
                 'punished_id' => $language->get('admin', 'punished_id'),
                 'punisher_id' => $language->get('admin', 'punisher_id'),
                 'reason' => $language->get('admin', 'reason'),
-            ]
+            ],
+            false,
+            true
```
samerton, line 382: This shouldn't be internal either

modules/Core/module.php
```diff
@@ -429,7 +435,9 @@ public function __construct(Language $language, Pages $pages, User $user, Naviga
                 'group_name' => $language->get('admin', 'group_name'),
                 'username' => $language->get('user', 'username'),
                 'user_id' => $language->get('admin', 'user_id'),
-            ]
+            ],
+            false,
+            true
```
samerton, line 440: This shouldn't be internal either

modules/Core/module.php
```diff
@@ -439,7 +447,9 @@ public function __construct(Language $language, Pages $pages, User $user, Naviga
...
```
samerton, line 452: This shouldn't be internal either

samerton (changes requested):

samerton (approved):

PR #2850, opened 2022-06-11
Rewrite configuration file operations

Derkades:
Also many random other fixes. This should hopefully prevent or at least reduce weird "config unavailable" issues.

### DIFF
--- a/core/classes/Misc/DebugBarHelper.php
+++ b/core/classes/Misc/DebugBarHelper.php
@@ -31,7 +31,7 @@ class DebugBarHelper extends Instanceable {
         $debugbar->addCollector(new RequestDataCollector());
 
         $configCollector = new ConfigCollector();
-        $configCollector->setData(array_filter($GLOBALS['config'], static function ($key) {
+        $configCollector->setData(array_filter(Config::read(), static function ($key) {
             return $key !== 'mysql';
         }, ARRAY_FILTER_USE_KEY));
         $debugbar->addCollector($configCollector);
--- a/core/classes/Misc/DebugBarHelper.php
+++ b/core/classes/Misc/DebugBarHelper.php
@@ -31,7 +31,7 @@ class DebugBarHelper extends Instanceable {
         $debugbar->addCollector(new RequestDataCollector());
 
         $configCollector = new ConfigCollector();
-        $configCollector->setData(array_filter(Config::read(), static function ($key) {
+        $configCollector->setData(array_filter(Config::all(), static function ($key) {
             return $key !== 'mysql';
         }, ARRAY_FILTER_USE_KEY));
         $debugbar->addCollector($configCollector);
--- a/core/classes/Misc/ErrorHandler.php
+++ b/core/classes/Misc/ErrorHandler.php
@@ -110,14 +110,14 @@ class ErrorHandler {
             $language = new Language('core', 'en_UK');
         }
 
-        $detailed_error = defined('DEBUGGING');
-        $can_generate_debug = defined('DEBUGGING');
+        $detailed_error = defined('DEBUGGING') && DEBUGGING;
+        $can_generate_debug = defined('DEBUGGING') && DEBUGGING;
 
         try {
             $user = new User();
             $detailed_error |= $user->isLoggedIn() && $user->hasPermission('admincp.errors');
             $can_generate_debug |= $user->hasPermission('admincp.core.debugging');
-        } catch (Error $ignored) {
+        } catch (Error | RuntimeException $ignored) {
             // Getting user info might fail, for example if the website isn't
             // installed yet. Assume the user does not have permission.
         }
--- a/core/classes/Misc/ErrorHandler.php
+++ b/core/classes/Misc/ErrorHandler.php
@@ -117,7 +117,7 @@ class ErrorHandler {
             $user = new User();
             $detailed_error |= $user->isLoggedIn() && $user->hasPermission('admincp.errors');
             $can_generate_debug |= $user->hasPermission('admincp.core.debugging');
-        } catch (Error | RuntimeException $ignored) {
+        } catch (Exception $ignored) {
             // Getting user info might fail, for example if the website isn't
             // installed yet. Assume the user does not have permission.
         }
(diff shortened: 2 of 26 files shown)


core/classes/Misc/DebugBarHelper.php
```diff
@@ -31,7 +31,7 @@ public function enable(Smarty $smarty): void {
         $debugbar->addCollector(new RequestDataCollector());
 
         $configCollector = new ConfigCollector();
-        $configCollector->setData(array_filter($GLOBALS['config'], static function ($key) {
+        $configCollector->setData(array_filter(Config::read(), static function ($key) {
```
Aberdeener, line 34: `Config::all()` might read nicer?

core/classes/Misc/ErrorHandler.php
```diff
@@ -110,14 +110,14 @@ public static function catchException(?Throwable $exception, ?string $error_stri
...
-        $can_generate_debug = defined('DEBUGGING');
+        $detailed_error = defined('DEBUGGING') && DEBUGGING;
+        $can_generate_debug = defined('DEBUGGING') && DEBUGGING;
 
         try {
             $user = new User();
             $detailed_error |= $user->isLoggedIn() && $user->hasPermission('admincp.errors');
             $can_generate_debug |= $user->hasPermission('admincp.core.debugging');
-        } catch (Error $ignored) {
+        } catch (Error | RuntimeException $ignored) {
```
Aberdeener, line 120: Weird, thought this was here already. What's `RuntimeException` for exactly - totally forget

core/init.php
```diff
@@ -22,12 +22,8 @@
     die('$page variable is unset. Cannot continue.');
 }
 
-if (!file_exists(ROOT_PATH . '/core/config.php')) {
-    if (is_writable(ROOT_PATH . '/core')) {
-        fopen(ROOT_PATH . '/core/config.php', 'w');
```
Aberdeener, line 27: Nice, this whole bit is much cleaner

core/init.php
```diff
@@ -22,12 +22,8 @@
...
-    } else {
-        die('Your <strong>/core</strong> directory is not writable, please check your file permissions.');
-    }
+if (!is_writable(ROOT_PATH)) {
```
Aberdeener, line 25: Why the whole `ROOT_PATH` here? Shouldn't we be more specific for just the config path?

core/installation/steps/requirements_validation.php
```diff
@@ -31,11 +31,7 @@
             </div>
             <div class="sixteen wide mobile eight wide tablet eight wide computer column">
                 <?php
-                validate_requirement('Core Writable <span class="ui basic label">/core</span>', is_writable('core'));
-                validate_requirement('Core Config Writable <span class="ui basic label">/core/config.php</span>', is_writable('core/config.php'));
-                validate_requirement('Core Email Writable <span class="ui basic label">/core/email.php</span>', is_writable('core/email.php'));
-                validate_requirement('Cache Writable <span class="ui basic label">/cache</span>', is_writable('cache'));
-                validate_requirement('Template Cache Writable <span class="ui basic label">/cache/templates_c</span>', is_writable('cache/templates_c'));
+                validate_requirement('Website directory writable <span class="ui basic label">' . Output::getClean(ROOT_PATH) . '</span>', is_writable(ROOT_PATH));
```
Aberdeener, line 34: Same thing about `ROOT_PATH` and not specific paths

modules/Core/queries/debug_link.php
```diff
@@ -207,7 +207,7 @@
         'groups' => $groups,
         'config' => [
             'core' => array_filter(
-                $GLOBALS['config']['core'],
+                Config::read()['core'],
```
Aberdeener, line 210: Can't we do something like `Config::get('core');` here? Or is that not supported?

Would be easy enough to let `Config::get(...)` support returning `string|array`

scripts/cli_install.php
```diff
@@ -111,7 +114,7 @@ function getEnvVar(string $name, string $fallback = null, array $valid_values =
     'core' => [
         'hostname' => getEnvVar('NAMELESS_HOSTNAME', 'localhost'),
         'path' => getEnvVar('NAMELESS_PATH', ''),
-        'friendly' => getEnvVar('NAMELESS_FRIENDLY_URLS', 'false') === 'true' ? true : false,
```
Aberdeener, line 114: This was intentional, unless it magically works now for you? Gotta love php

scripts/cli_install.php
```diff
@@ -144,13 +140,19 @@ function getEnvVar(string $name, string $fallback = null, array $valid_values =
 }
 
 print('✍️  Creating tables...' . PHP_EOL);
-PhinxAdapter::migrate();
+
+$message = PhinxAdapter::migrate();
+
+if (!str_contains($message, 'All Done')) {
```
Aberdeener, line 146: Nice, didn't consider needing to see Phinx errors during CLI install

Aberdeener (changes requested):

core/classes/Misc/ErrorHandler.php
Derkades: This should be `Exception`, all exceptions inherit from `Exception` and all errors inherit from `Error`. We probably shouldn't be catching errors.

modules/Core/queries/debug_link.php
Derkades: Good idea, then instead of adding a Config::all() method, just `Config::get('')` could be used

scripts/cli_install.php
Derkades: The result of === returns boolean true/false so this must work

core/init.php
Derkades: I forgot is_writable obviously doesn't check recursively

Derkades:
Accidentally committed the proxies stuff to the wrong branch, please ignore it

modules/Core/queries/debug_link.php
Aberdeener: Honestly just to scratch an itch, I'd make `Config::all()` just an alias to `Config::get('')`

Aberdeener:
Whoops, I did not mean to commit that here :D

PR #2869, opened 2022-06-15
Use correct language codes

Derkades:
I think it makes more sense to use the standard codes for traditional and simplified Chinese rather than zh_TW (taiwan?) for traditional and zh_CN (china?) for simplified. I say that as someone who knows nothing about the Chinese language. Weblate seems to use Hans/Hant as well.

Could someone help with the migrations


Aberdeener:
This makes sense to do, but I think we should do it in 2.1.0.
I don't want any risks of further breaking changes for users, pre13 was quite rough 😬 
What do you think?

ichenglin:
It does make sense if you guys are not aiming for regional based translations.

The reason in which language codes zh_TW and zh_CN are usually referred to distinguish Traditional and Simplified is because they hold the greatest population. However, there are still some minor regional differences in the version of Chinese used by different countries.

Traditional Chinese:
- zh_TW (Taiwan)
- zh_HK (Hong Kong)
- zh-MO (Macau)

Simplified Chinese:
- zh_CN (Mainland China)
- zh_SG (Singapore)

ichenglin:
Hans and Hant would work out for most cases as the major difference is the characters used, while slight differences such as vocabulary shouldn't be hard to interpret. To be honest, most Chinese readers know how to read both Traditional and Simplified characters, so it's mostly all just about preference.

PikaMug:
Go with whatever the platform prefers. As a Crowdin user, zn-TW was the default (and is [what Minecraft uses](https://crowdin.com/project/minecraft/zh-TW)) and has wider adoptation, but zh-Hant does seem to have advantages. Neither is particularly stellar, though, at least in the eyes of the consortium https://www.w3.org/International/articles/bcp47/#script

Issue #2876, opened 2022-06-19
Add fourth module column to nl2_settings


Derkades:
and make Util::getSetting and setSetting work for modules

Derkades:
I have opened https://github.com/NamelessMC/Nameless/issues/2880 for increasing the length, I'm not sure how to do that

PR #2879, opened 2022-06-20
Add module column to settings table

Derkades:
Resolves https://github.com/NamelessMC/Nameless/issues/2876


Aberdeener:
Can we deprecate the `Configuration` class now?

core/classes/Core/Util.php
```diff
@@ -283,6 +283,22 @@ public static function replaceAnchorsWithText(string $data): string {
         return URL::replaceAnchorsWithText($data);
     }
 
+    private static function getSettingsCache(?string $module): ?array {
+        $cache_name = $module !== null ? $module : 'core';
```
Aberdeener, line 287: Here we default to module name of `'core'`, but in `Util::getSetting` we default to null as the module name for setting lookup.
I think we should use 1 convention

core/migrations/20220620183016_add_module_settings_column.php
```diff
@@ -0,0 +1,16 @@
+<?php
+declare(strict_types=1);
+
+use Phinx\Migration\AbstractMigration;
+
+final class AddModuleSettingsColumn extends AbstractMigration {
+
+    public function change(): void {
+        $table = $this->table('nl2_settings');
+        $table->addColumn('module', 'string', ['length' => 32, 'null' => true, 'default' => null]);
```
Aberdeener, line 10: Going off my previous comment, should we default to `'core'` here? Let me know your thoughts

Aberdeener (approved):

partydragen:
> Nice work, a couple thoughts:
> 
> * Can we deprecate the `Configuration` class now?
> * Do these methods belong in the `Util` class? Maybe we keep the `Configuration` class and move all related logic from `Util` to it...

Well it could been in Configuration but that will be a breaking change, believe its only my modules that use it thought ¯\_(ツ)_/¯

Aberdeener:
It doesn’t have to be a breaking change right now. We can mark them with
@deprecated and remove them in 2.1.0 (I’ve done this in many other places
already)

core/classes/Core/Util.php
Derkades: Does php support null keys for dictionaries? I just assumed it didn't

Derkades: Apparently it does support null keys but it has weird semantics (actually stored as an empty string). Probably best to avoid.

partydragen (approved):

partydragen:
So now this only missing https://github.com/NamelessMC/Nameless/issues/2880

core/classes/Core/Util.php
samerton, line 287: I take it there's no issue with this being lower case for `core` whereas the official module name is `Core`

core/classes/Core/Util.php
```diff
@@ -283,6 +283,22 @@ public static function replaceAnchorsWithText(string $data): string {
         return URL::replaceAnchorsWithText($data);
     }
 
+    private static function getSettingsCache(?string $module): ?array {
```
samerton, line 286: Could these parameters not just be
```php
private static function getSettingsCache(string $module = 'core'): ?array {
```
to save the assignment on 287 etc

samerton (changes requested):

core/classes/Core/Util.php
Derkades: I used null because that was consistent with how it's actually stored in the database, but I will change it.

core/classes/Core/Util.php
Derkades: there's nothing that enforces modules to use their settings, nothing prevents the store module from calling `setSetting('something', 'value', $module='suggestions')`, for example. I assumed lower case, but case matching module names might make more sense.

samerton (approved):

PR #2881, opened 2022-06-20
Allow server to send correct verify command

Derkades:
The plugin command is configurable, so the website should display the same command


partydragen:
What if server part is disabled, like no server id set then server data not being sendt, they might just use it for verify command

PR #2886, opened 2022-06-22
Handle non-ascii characters in subject, SITE_NAME and message

Derkades:
Opening a PR because I can't test these changes


Aberdeener:
What was the situation that required these changes? Is there a thread you can link on Discord?

Derkades:
This is a potential security issue, being able to insert line breaks into a string allows sending arbitrary headers or even an arbitrary message body

PR #2900, opened 2022-06-28
Use png instead of svg for initials avatar source

supercrafter100:


Derkades:
why?

Aberdeener:
Svg does not work in discord embeds (convo in party’s server)

PR #2915, opened 2022-07-07
Convert default panel and frontend templates to submodules

Aberdeener:
By splitting the templates into submodules, it will let our template developers track required changes much easier, as well as pull the upstream changes into their templates.

These templates are now located at:
- https://github.com/NamelessMC/DefaultRevamp
- https://github.com/NamelessMC/Default
Thanks to the power of git, all of the commits have been kept intact.

Making a change to a template requires committing that change to its repo separately (duh)
  - In PHPStorm, if you open the Nameless directory with the submodules properly setup, you can make changes to files in Nameless and the submodules and it will commit them properly for you!	
    - Screen Shot 2022-07-06 at 22 37 45
We will also need to make sure we update the ref to the submodule(s) in Nameless if we make a change to them (I'm not sure if this is done magically for us when the commit is made in PHPStorm?)

Lastly, we will have to let people know that they must include submodules when cloning the repo with `git clone --recurse-submodules`


Derkades:
> Thanks to the power of git, all of the commits have been kept intact

This is really cool, I did not know that was possible!

Aberdeener:
Will come back to this once we make a decision about whether or not to continue allowing custom panel templates

Issue #2865, opened 2022-06-15
Website->Discord group sync issues


MakeItReaal:
Dear Aberdeener,

The group sync on the discord roles currently has some issues when you change your rank in the game.
Your roles get set on the forums to the correct rank but on discord  it only removes rank and doesn't add them.

I had contact with Derkades he said i should create an issue for this on github.
Can you look in to this?

Kind regards!

Aberdeener:
Possible fix:tm: in https://github.com/NamelessMC/Nameless/commit/27fcb265a2140857357d3e1ef1d2b3550649e045

Waiting to hear back from users. If they are not successful I will setup an MC server and Discord server this weekend to do more extensive testing. I really want to banish all issues with group sync so we can forget about it for a while 😄

HarmHoog:
For me, these changes don't seem to fix it (pre-13 with these changes). However, I also have the issue that removing a role via Discord also doesn't get synced with the website whilst the other way around does.

DominikSLK:
> For me, these changes don't seem to fix it (pre-13 with these changes). However, I also have the issue that removing a role via Discord doesn't get synced with the website whilst the other way around does.

same

Gyztor:
> Possible fixtm in [27fcb26](https://github.com/NamelessMC/Nameless/commit/27fcb265a2140857357d3e1ef1d2b3550649e045)
> 
> Waiting to hear back from users. If they are not successful I will setup an MC server and Discord server this weekend to do more extensive testing. I really want to banish all issues with group sync so we can forget about it for a while smile

tested and seems to work stably when changing roles on discord, they reflect properly on the website now but changing on the website does not reflect properly to discord whether its removing or adding roles. so discord->website seems to be doing good website->discord is just ded this is also on latest pull from github at time of writing ( for reference: 5eb5db163ed59a5a5a2ac61bc8c2e3977b9f62c8 )

DominikSLK:
#2920 fixes it for me :D 🎉

Aberdeener:
> #2920 fixed it for me :D 🎉 

Thank you so much for letting us know! Great to hear!

Aberdeener:
> The website keeps sending /roleChange requests repeatedly: https://paste.rkslot.nl/asefipoyoc.txt
> 
> 
> 
> A few issues with this:
> 
> - The website keeps sending role remove requests, even though it has already send those requests before.
> 
[...]

PR #2920, opened 2022-07-10
group sync fixes:tm:

Aberdeener:
working for me as expected,  the trade-off is that `"discord_role_id -> <role id>"` will be logged every time a group sync broadcast is made to their site logs, even if they already had the group

this *could* be avoided by two ways:
- not log any discord injector changes, meh
	- group sync logging is a mess as it is. might be fine just removing the whole thing 🤔 
- making the bot return some nicer info when requests are made, such as whether the group was added/removed properly, or if they already had it/didn't have it
	- this is probably the better option, but very annoying to implement.

waiting for some more testing from someone else, will merge/edit/rage-quit when i hear back from them

will close https://github.com/NamelessMC/Nameless/issues/2865

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -59,6 +59,7 @@
 - Fix `AssetResolver` adding an asset more than once if several assets depend on it
 - Fix TinyMCE spoiler plugin not being added properly if a custom path is set
 - Fix error when creating group sync rules with only 1 external service
+- Fix website -> Discord group sync issues
 
 ### Deprecated
 *These will be removed in 2.1.0*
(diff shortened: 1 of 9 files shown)


Derkades:
New API proposal, v5: https://github.com/NamelessMC/Nameless-Link/wiki/API

Aberdeener:
> In testing, when a user linked their discord user, the discord role was just automatically set. Now that I'm running on the official release, it doesn't work anymore

Aberdeener:
> New API proposal, v5: [NamelessMC/Nameless-Link/wiki/API](https://github.com/NamelessMC/Nameless-Link/wiki/API)

let's go ahead and do this 🙏🏼 
another thing, is it possible to do something like a status of "invalid_role_id" when the website sends a role id that does not exist in the discord server? it would help to be able to log when this happens. currently "partsuccess" is returned

PR #2942, opened 2022-07-24
Block banned ips from registering

### LINKED ISSUE #2929: Block banned IPs from registering
We disallow users to login if their IP is banned, but not from registering. Should be an easy patch

supercrafter100:
Closes #2929

### DIFF
--- a/modules/Core/pages/register.php
+++ b/modules/Core/pages/register.php
@@ -220,6 +220,14 @@ if (Input::exists()) {
                 },
             ]);
 
+            // Check if the ip they are trying to register with is banned
+            $ip = HttpUtils::getRemoteAddress();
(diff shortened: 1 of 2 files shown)


modules/Core/pages/register.php
```diff
@@ -220,6 +220,13 @@
                 },
             ]);
 
+            // Check if the ip they are trying to register with is banned
+            $ip = HttpUtils::getRemoteAddress();
+            if (DB::getInstance()->get('ip_bans', ['ip', $ip])->count()) {
+                Session::flash('home_error', $language->get('user', 'you_have_been_banned'));
```
Aberdeener, line 226: "Your IP has been banned, you cannot register." would be a more explicit message here

Aberdeener (approved):

PR #2943, opened 2022-07-24
Support 2FA with OAuth2 logging in

### LINKED ISSUE #2939: OAuth login skips TFA
### Describe the issue

When logging in with an OAuth account, if the NamelessMC user has tfa enabled, it gets skipped.

### NamelessMC Version

Development version

### To Reproduce

_No response_

### Expected Behaviour

We should complete the OAuth flow in the `modules/Core/pages/oauth.php`, and then instead of logging in and redirecting to the homepage we should redirect to the tfa page.

### Screenshots

_No response_

### Additional Information

_No response_

supercrafter100:
Closes #2939

### DIFF
--- a/modules/Core/pages/oauth.php
+++ b/modules/Core/pages/oauth.php
@@ -1,5 +1,8 @@
 <?php
 
+const PAGE = 'oauth';
+require_once(ROOT_PATH . '/core/templates/frontend_init.php');
+
 if (!isset($_GET['provider'], $_GET['code'])) {
     if (!array_key_exists($_GET['provider'], NamelessOAuth::getInstance()->getProvidersAvailable())) {
         throw new RuntimeException("Invalid provider {$_GET['provider']}");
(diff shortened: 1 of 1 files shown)


modules/Core/pages/oauth.php
```diff
@@ -10,6 +13,31 @@
     throw new RuntimeException('No OAuth method set');
 }
 
+// If they are filling in 2FA. We've already retrieved their user. We can skip the other steps in this case
+if (isset($_SESSION['user_id']) && isset($_POST['tfa_code'])) {
+    $user = new User($_SESSION['userid']);
```
partydragen, line 18: if isset user_id but New User using userid????

partydragen (changes requested):

modules/Core/pages/oauth.php
supercrafter100: whooops

supercrafter100:
Code may need some cleanup but it works

modules/Core/pages/oauth.php
```diff
@@ -1,5 +1,8 @@
 <?php
 
+const PAGE = 'oauth';
+require_once(ROOT_PATH . '/core/templates/frontend_init.php');
```
Aberdeener, line 4: Why do we need frontend_init here?

supercrafter100: So the template variable is properly assigned when the 2FA page is rendered

Aberdeener (approved):

samerton (approved):

partydragen (approved):

PR #2944, opened 2022-07-24
Only show source code option in staffcp pages for TinyMCE

### LINKED ISSUE #2941: Make source code in editor for admin only
Make the source code to only be used when admin param is true, this will require all pages in staffcp that contain a editor to be updated to have admin param true

supercrafter100:
Closes #2941

### DIFF
--- a/core/classes/Core/Input.php
+++ b/core/classes/Core/Input.php
@@ -129,7 +129,7 @@ class Input {
               external_plugins: {
                 'spoiler': '" . URL::build('/core/assets/plugins/tinymce_spoiler/plugin.min.js') . "',
               },
-              toolbar: 'undo redo | bold italic underline strikethrough formatselect fontsizeselect forecolor backcolor ltr rtl emoticons | alignleft aligncenter alignright alignjustify | codesample code hr image link numlist bullist | spoiler-add spoiler-remove',
+              toolbar: 'undo redo | bold italic underline strikethrough formatselect fontsizeselect forecolor backcolor ltr rtl emoticons | alignleft aligncenter alignright alignjustify | codesample " . ($admin ? "code" : "") . " hr image link numlist bullist | spoiler-add spoiler-remove',
               spoiler_caption: '{$language->get('general', 'spoiler')}',
               default_link_target: '_blank',
               skin: '$skin'," .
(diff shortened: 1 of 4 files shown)


core/classes/Core/Input.php
```diff
@@ -129,7 +129,7 @@ public static function createTinyEditor(Language $language, string $name, ?strin
               external_plugins: {
                 'spoiler': '" . URL::build('/core/assets/plugins/tinymce_spoiler/plugin.min.js') . "',
               },
-              toolbar: 'undo redo | bold italic underline strikethrough formatselect fontsizeselect forecolor backcolor ltr rtl emoticons | alignleft aligncenter alignright alignjustify | codesample code hr image link numlist bullist | spoiler-add spoiler-remove',
+              toolbar: 'undo redo | bold italic underline strikethrough formatselect fontsizeselect forecolor backcolor ltr rtl emoticons | alignleft aligncenter alignright alignjustify | codesample " . ($admin ? "code" : "") . " hr image link numlist bullist | spoiler-add spoiler-remove',
```
samerton, line 132: I guess the extra space if admin is disabled (it'll render as `codesample  hr` won't matter?

supercrafter100: Didn't matter when I tested at least

samerton (approved):

custom/panel_templates/Default/template.php
```diff
@@ -232,7 +232,7 @@ public function onPageLoad() {
                             (defined('CONFIG_PATH') ? CONFIG_PATH : '') . '/core/assets/js/fomantic.toast.min.js' => [],
                         ]);
 
-                        $this->addJSScript(Input::createTinyEditor($this->_language, 'InputMaintenanceMessage'));
```
Derkades, line 235: You can do `$admin=true` instead of `null, false, true` so you don't have to set the earlier optional parameters

Aberdeener: Nope 😅
In php 8.0 you can use the named param feature but if we did $admin=true it would set the variable $admin to true in scope and then set whatever param is at that position to true

Derkades: huh.. too used to python I guess. thanks for the explanation

PR #2983, opened 2022-08-06
Fix non-string config values not saving properly

samerton:
Fixes issue where friendly URLs, force HTTPS, force WWW were saving into config as `'1'` and `'0'` instead of `true` and `false`


Derkades (approved):

partydragen:
Why are we switching back? 

`// Check if it is a string here so that `null` is not converted to `''`
            $loc = is_string($value) ? addslashes($value) : $value;` this was the way that not worked
but now we use same way just switched they around so true is last?

PR #2985, opened 2022-08-06
Remove mentions of "pre-release" from installer

Aberdeener:


Derkades:
Please note that you only have to remove terms in en_UK, weblate will remove it in other files automatically. This helps avoid conflicts if weblate happens to have edits queued for that file at the same time. You don't have to change the PR though, it'll most likely be fine

Derkades (approved):

Aberdeener:
Ahh good point @Derkades!
Old habits die hard

PR #2987, opened 2022-08-06
Make update check more resilient to API being down

tadhgboyle:


core/classes/DTO/UpdateCheck.php
```diff
@@ -9,18 +9,24 @@
...
 
-    private array $_response;
+    private ?string $_raw_response;
+    private ?array $_response;
 
-    public function __construct(array $response) {
-        $this->_response = $response;
+    public function __construct(HttpClient $update_check_response) {
+        $this->_raw_response = $update_check_response->contents();
+        $this->_response = json_decode($this->_raw_response, true);
```
samerton, line 17: Could json_decode throw due to invalid JSON response when the API is down?

Aberdeener: Only if we pass a third argument, it returns null if invalid input without the argument

samerton (approved):

partydragen (approved):

Aberdeener:
Closes https://github.com/NamelessMC/Nameless/issues/2979

PR #2995, opened 2022-08-07
:fire: Remove support for custom panel templates

tadhgboyle:
We have talked internally about this change, and decided unanimously that it is worth it.
- Very few (if any) users are using custom panel templates
- None of the 4 panel templates on namelessmc.com have been updated for even pre13

**The biggest reason is that removing custom panel template support means we can make changes to the panel template in a minor release, since we will not have to worry about breaking other custom panel templates. This will allow us to release updates to site configuration faster and more often!**

Requires https://github.com/NamelessMC/debug.namelessmc.com/pull/175


DominikSLK:
😶

samerton (dismissed):

PadowYT2 (dismissed):
Why should be this done by the context below
Noir - repo was deleted by Teemo
Argon - I'm lazy :smile:
Greendy - Nuke has a lot of problems irl
Dark_Blue - outdated & Viper left NMC

supercrafter100 (dismissed):

Aberdeener:
A good alternate proposal from @partydragen is to display a warning in StaffCP that panel templates have no compatibility promise.
I will change this PR to add this alert instead of fully removing the feature. This way it's a win-win!

DominikSLK:
> A good alternate proposal from @partydragen is to display a warning in StaffCP that panel templates have no compatibility promise.
> I will change this PR to add this alert instead of fully removing the feature. This way it's a win-win!

yes :tada:

PR #2998, opened 2022-08-07
Add ability to rate limit via Validate class

samerton:


modules/Core/pages/login.php
```diff
@@ -82,12 +86,14 @@
                 'email' => [
                     Validate::REQUIRED => $language->get('user', 'must_input_email'),
                     Validate::IS_BANNED => $language->get('user', 'account_banned'),
-                    Validate::IS_ACTIVE => $language->get('user', 'inactive_account')
+                    Validate::IS_ACTIVE => $language->get('user', 'inactive_account'),
+                    Validate::RATE_LIMIT => fn($meta) => $language->get('general', 'rate_limit', $meta),
```
supercrafter100, line 90: Wouldn’t this have to be $meta->expires ?

supercrafter100: Nvm

core/classes/Core/Validate.php
```diff
@@ -318,6 +324,50 @@ public static function check(array $source, array $items = []): Validate {
...
+                        } else if (is_int($rule_value)) {
+                            // If integer default seconds to 60
+                            [$limit, $seconds] = [$rule_value, 60];
+                        }
+
+                        if (!isset($limit) || !isset($seconds)) {
+                            throw new Exception('Invalid rate limit configuration');
+                        }
+
+                        $session = $_SESSION["rate_limit_{$item}"];
```
Aberdeener, line 341: We are repeating the "rate_limit_{$item}" a few times here, can we create a variable like $rate_limit_key to not need to repeat it?

Aberdeener (changes requested):

partydragen (approved):

Aberdeener (approved):
Great work, this is nice to have! Will try to remember to use it in places we have other cooldown systems right now

PR #3000, opened 2022-08-07
Rework user sessions system

partydragen:
Tasks:
- [x] Switch session to use hash
- [x] Logout all other hashes when user changing password and enables tfa
- [x] Update database
- [ ] Add new UserCP page to view sessions

### DIFF
--- a/core/classes/Core/User.php
+++ b/core/classes/Core/User.php
@@ -320,7 +320,6 @@ class User {
         } else if ($this->checkCredentials($username, $password, $method) === true) {
             // Valid credentials
             $hash = SecureRandom::alphanumeric();
-            Session::put($sessionName, $hash);
 
             $this->_db->insert('users_session', [
                 'user_id' => $this->data()->id,
@@ -329,6 +328,8 @@ class User {
                 'active' => 1,
                 'login_method' => $method
             ]);
+
+            Session::put($sessionName, $hash);
 
             if ($remember) {
                 $expiry = $is_admin ? 3600 : Config::get('remember.cookie_expiry');
--- a/core/classes/Core/User.php
+++ b/core/classes/Core/User.php
@@ -549,7 +549,9 @@ class User {
      * Process logout if user is admin
      */
     public function admLogout(): void {
-        $this->_db->delete('users_admin_session', ['user_id', $this->data()->id]);
+        $this->_db->update('users_session', ['user_id', $this->data()->id], [
+            'active' => 0
+        ]);
 
         Session::delete($this->_admSessionName);
         Cookie::delete($this->_cookieName . '_adm');
--- a/core/classes/Core/User.php
+++ b/core/classes/Core/User.php
@@ -6,7 +6,7 @@
  * @author Samerton
  * @author Partydragen
  * @author Aberdeener
- * @version 2.0.0-pr13
+ * @version 2.0.2
  * @license MIT
  */
 class User {
@@ -301,7 +301,7 @@ class User {
      * @param string|null $username Their username (or email, depending on $method).
      * @param string|null $password Their password.
      * @param bool $remember Whether to keep them logged in or not.
-     * @param string $method What column to check for their details in. Can be either `username` or `email`.
+     * @param string $method What column to check for their details in. Can be either `username` or `email` or `oauth`.
      *
      * @return bool True/false on success or failure respectfully.
      */
@@ -326,7 +326,7 @@ class User {
                 'hash' => $hash,
                 'remember_me' => $remember,
                 'active' => 1,
-                'login_method' => $method
+                'login_method' => $is_admin ? 'admin' : $method
             ]);
 
             Session::put($sessionName, $hash);
--- a/core/classes/Core/User.php
+++ b/core/classes/Core/User.php
@@ -73,14 +73,14 @@ class User {
 
         if ($user === null) {
             if (Session::exists($this->_sessionName)) {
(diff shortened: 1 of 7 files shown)


Aberdeener:
Two comments:
- "Add new UserCP page to view sessions"; this means we will require templates to update, meaning it will have to be in a major 2.x.0 update
- Is it possible to store the login method in the sessions table? Would be neat if we could display how they logged in on each of their active sessions!

partydragen:
> Two comments:
> 
> * "Add new UserCP page to view sessions"; this means we will require templates to update, meaning it will have to be in a major 2.x.0 update
> * Is it possible to store the login method in the sessions table? Would be neat if we could display how they logged in on each of their active sessions!

1. Not really a breaking change but people need to update their templates yea, at the same time this is a security feature with make it important, Lets see what Sam says

2. Should be possible yea by using the existing method param with show username/email/oauth atleast 
3.

core/classes/Core/User.php
```diff
@@ -307,29 +311,25 @@ public function login(?string $username = null, ?string $password = null, bool $
 
     private function _commonLogin(?string $username, ?string $password, bool $remember, string $method, bool $is_admin): bool {
         $sessionName = $is_admin ? $this->_admSessionName : $this->_sessionName;
-        if (!$username && !$password && $this->exists()) {
-            Session::put($sessionName, $this->data()->id);
+        if (!$username && $method == 'hash' && $this->exists()) {
+            // Logged in using hash from cookie
+            Session::put($sessionName, $password);
```
supercrafter100, line 316: Are we sure that this has to be the password and not the user id?

core/classes/Core/User.php
```diff
@@ -107,7 +107,11 @@ public function find(string $value = null, string $field = 'id'): bool {
                 return true;
             }
 
-            $data = $this->_db->get('users', [$field, $value]);
+            if ($field != 'hash') {
+                $data = $this->_db->get('users', [$field, $value]);
+            } else {
+                $data = $this->_db->query('SELECT nl2_users.* FROM nl2_users LEFT JOIN nl2_users_session ON nl2_users.id=user_id WHERE hash = ? AND nl2_users_session.active = 1', [$value]);
```
Aberdeener, line 113: Space between `nl2_users.id = user_id` please 😄

core/init.php
```diff
@@ -132,11 +132,11 @@
     // Do they need logging in (checked remember me)?
     if (Cookie::exists(Config::get('remember.cookie_name')) && !Session::exists(Config::get('session.session_name'))) {
         $hash = Cookie::get(Config::get('remember.cookie_name'));
-        $hashCheck = DB::getInstance()->get('users_session', ['hash', $hash]);
+        $hashCheck = DB::getInstance()->get('users_session', [['hash', $hash], ['active', 1]]);
```
Aberdeener, line 135: `['active', true]`

modules/Core/pages/user/settings.php
```diff
@@ -410,7 +417,7 @@
 
                                     // Update email
                                     $user->update([
-                                        'email' => Output::getClean($_POST['email'])
+                                        'email' => $_POST['email']
```
Aberdeener, line 420: Why this change?

core/migrations/20220807142446_add_user_session_activity_columns.php
```diff
@@ -0,0 +1,21 @@
+<?php
+declare(strict_types=1);
+
+use Phinx\Migration\AbstractMigration;
+
+final class AddUserSessionActivityColumns extends AbstractMigration
+{
+    public function change(): void
+    {
+        $table = $this->table('nl2_users_session');
```
Aberdeener, line 10: Can we add the "login method" column here?

tadhgboyle (changes requested):

modules/Core/pages/user/settings.php
partydragen: Well we no longer clean when saving to database ¯\_(ツ)_/¯

core/migrations/20220807142446_add_user_session_activity_columns.php
partydragen: Done

core/classes/Core/User.php
```diff
@@ -547,7 +560,9 @@ public function logout(): void {
      * Process logout if user is admin
      */
     public function admLogout(): void {
-        $this->_db->delete('users_admin_session', ['user_id', $this->data()->id]);
+        $this->_db->update('users_session', [['user_id', $this->data()->id], ['hash', Session::get($this->_admSessionName)]], [
```
samerton, line 564: This `admLogout` session should be deleted now

samerton (changes requested):

samerton (approved):

Aberdeener:
Is displaying the sessions in StaffCP delayed as well, or will that be added to this PR?

partydragen:
> Is displaying the sessions in StaffCP delayed as well, or will that be added to this PR?

Delayed so pr is done

supercrafter100 (approved):

supercrafter100:
Im guessing everything has been tested and confirmed working?

partydragen:
> Im guessing everything has been tested and confirmed working?

Have tested yea, was not really that major changes and is only installer, login, re-auth and oauth that using login and its all using the user login function so it worked everywhere by my testing, the changes itself is pretty simple

tadhgboyle (approved):

Issue #2968, opened 2022-08-05
Better "n migrations pending" page


Aberdeener:
- List all pending migrations
- Button to run them?

PR #3005, opened 2022-08-08
Better missing/extra migration exception page

Aberdeener:
Currently when a site is missing migrations, we [I] do three things that are not ideal:
- Throw an exception
	- This is bad because it requires debugging mode to be enabled, or the user to be admin logged in, which means they often have to check the logs to even see the message
- Missing migrations are determined by just comparing the *size* of the `core/migrations` table and the `nl2_phinxlog` table
    - This is not great because it does not consider the contents of them, and now that we have a new branch system, the likelihood of migrations being a little more messy in dev is much higher, so we should be stricter
- It does not say which migrations are missing

Screen Shot 2022-08-08 at 08 46 06

This change checks the file names and compares them to the phinxlog table contents, and prints the issues clearly in plain text so they do not need to check their logs.
The page is in plaintext since other fatal issues (wrong php version, missing vendor dir, etc) are also plaintext.

Finally, this is still much faster than forcing phinx to migrate on every page load:
- Force migrate ~18ms:
	- Screen Shot 2022-08-08 at 08 52 55
- This solution ~6ms:
    - Screen Shot 2022-08-08 at 08 53 22

### DIFF
--- a/core/classes/Database/PhinxAdapter.php
+++ b/core/classes/Database/PhinxAdapter.php
@@ -9,22 +9,45 @@ class PhinxAdapter {
      * @throws RuntimeException If these numbers don't match.
      */
     public static function ensureUpToDate(): void {
-        $migration_files = count(scandir(__DIR__ . '/../../migrations')) - 3; // -3 because of . and .. and phinx.php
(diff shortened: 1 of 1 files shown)


supercrafter100 (approved):

samerton (approved):

Aberdeener:
Closes https://github.com/NamelessMC/Nameless/issues/2968

PR #3006, opened 2022-08-08
Fix several profile images issues

partydragen:
Fix several profile images issues

### DIFF
--- a/core/includes/image_upload.php
+++ b/core/includes/image_upload.php
@@ -2,7 +2,7 @@
 /*
  *  Made by Samerton
  *  https://github.com/NamelessMC/Nameless/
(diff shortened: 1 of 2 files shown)


partydragen:
Ah right, This does not solve existing folders with incorrect permissions

core/includes/image_upload.php
```diff
@@ -83,6 +83,14 @@
             Redirect::to(URL::build('/profile/' . urlencode($user->data()->username)));
         }
 
+        if (
```
samerton, line 86: Do we not have language available here?

tadhgboyle (approved):

PR #3009, opened 2022-08-11
Remake tab initialization

PadowYT2:
Honestly, when updating to Fomantic UI, with it it is a problem, because "The menu items need to have a referencing `data-tab` property and all missing tabs also a corresponding `tab segment`"
e.g.
```html
<div class="ui top attached tabular menu">
  <a class="item active" data-tab="first">First</a>
  <a class="item" data-tab="second">Second</a>
  <a class="item" data-tab="third">Third</a>
</div>
<div class="ui bottom attached tab segment active" data-tab="first">
  First
</div>
<div class="ui bottom attached tab segment" data-tab="second">
  Second
</div>
<div class="ui bottom attached tab segment" data-tab="third">
  Third
</div>
```

But it is easier to just remove it.
"It makes more sense to fix it yeah but I doubt people will leave a bad review because of it" - @supercrafter100

### DIFF
--- a/custom/templates/DefaultRevamp/js/core/core.js
+++ b/custom/templates/DefaultRevamp/js/core/core.js
@@ -48,6 +48,8 @@ $(function () {
     $('.message .close').on('click', function () {
         $(this).closest('.message').transition('fade');
     });
+
+    $('.tabular.menu .item').tab();
 });
 
 $(function () {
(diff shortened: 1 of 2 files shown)


samerton:
If we remove this line then profiles tabs etc will also break

samerton (approved):

PR #3012, opened 2022-08-12
Fix exception when user id is null in email_error table

partydragen:
Fix exception when user id is null in email_error table

### DIFF
--- a/modules/Core/pages/panel/emails_errors.php
+++ b/modules/Core/pages/panel/emails_errors.php
@@ -2,7 +2,7 @@
 /*
  *  Made by Samerton
  *  https://github.com/NamelessMC/Nameless/
- *  NamelessMC version 2.0.0-pr9
+ *  NamelessMC version 2.0.2
  *
  *  License: MIT
  *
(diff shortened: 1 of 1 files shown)


modules/Core/pages/panel/emails_errors.php
```diff
@@ -78,7 +78,7 @@
             'BACK_LINK' => URL::build('/panel/core/emails/errors'),
             'VIEWING_ERROR' => $language->get('admin', 'viewing_email_error'),
             'USERNAME' => $language->get('user', 'username'),
-            'USERNAME_VALUE' => Output::getClean($user->idToName($error->user_id)),
+            'USERNAME_VALUE' => $error->user_id ? Output::getClean($user->idToName($error->user_id)) : '',
```
Aberdeener, line 81: Can we match the behavior of the logs page? It should display "Deleted User" instead of an empty string
https://github.com/NamelessMC/Nameless/blob/c1e179c6de0ed131f2d1201579813ac1edeaf4a6/modules/Core/pages/panel/security.php#L91-L95

Aberdeener (changes requested):

modules/Core/pages/panel/emails_errors.php
partydragen: Funny, i had even did that but then i removed it xD

anyways added back now :D

samerton (approved):

Issue #2910, opened 2022-07-03
Show OAuth URL when setting it up


supercrafter100:
### Describe the feature

The title pretty much explains it. Instead of forcing people to look up on the documentation on what to put for the redirect uri, just show it. It makes it so people can't make any mistakes when trying to copy it over as it will be the exact same as the one it puts as the redirect URI for the services.

PR #3014, opened 2022-08-13
Display each OAuth providers redirect URL & misc fixes to template

Aberdeener:
This PR removes a major pain point for users; finding their OAuth redirect URL.
It displays the Redirect URL under each provider, in a readonly input field so it is selectable.
This also makes the "Client ID" and "Client Secret" strings translatable, something I forgot to do in the original OAuth PR!

Screen Shot 2022-08-12 at 22 24 35


supercrafter100:
Closes #2910

supercrafter100 (approved):

modules/Core/pages/panel/registration.php
```diff
@@ -190,6 +190,10 @@
         'docLinkStart' => '<a href="https://docs.namelessmc.com/en/oauth" target="_blank">',
         'docLinkEnd' => '</a>'
     ]),
+    'REDIRECT_URL' => $language->get('admin', 'redirect_url'),
+    'CLIENT_ID' => $language->get('admin', 'client_id'),
+    'CLIENT_SECRET' => $language->get('admin', 'client_secret'),
+    'OAUTH_URL' => URL::getSelfURL() . 'oauth/&provider={{provider}}',
```
partydragen, line 196: This does not support un-friendly urls

`URL::getSelfURL() . ltrim(URL::build('/oauth/', 'provider={{provider}}'), '/'),`
^^ like we do on webhook events url

partydragen (changes requested):

modules/Core/pages/panel/registration.php
samerton: This should be resolved now

partydragen (approved):

PR #3016, opened 2022-08-13
Fix empty "No News" alert when Forum module is disabled

### LINKED ISSUE #3011: No NO_NEWS message without a Forum module
### Describe the issue

`NO_NEWS` variable is assigned in the Forum module, but when it is disabled, the message is blank

https://github.com/NamelessMC/Nameless/blob/c1e179c6de0ed131f2d1201579813ac1edeaf4a6/modules/Forum/front_page.php#L58

### NamelessMC Version

Development version

### To Reproduce

_No response_

### Expected Behaviour

_No response_

### Screenshots

_No response_

### Additional Information

_No response_

tadhgboyle:
Closes https://github.com/NamelessMC/Nameless/issues/3011


custom/templates/DefaultRevamp/index.tpl
```diff
@@ -65,9 +65,11 @@
                 </div>
             </div>
             {foreachelse}
-            <div class="ui segment red">
-                {$NO_NEWS}
-            </div>
+                {if $FORUM_ENABLED}
```
Aberdeener, line 68: This is pretty cursed - ideally the index file should have no idea about what modules are enabled. This can be fixed later:tm: with the module home page type revamp

supercrafter100: Why don’t you just check if the variable is set?

modules/Core/pages/home.php
```diff
@@ -32,10 +32,12 @@
 }
 
 $home_type = Util::getSetting('home_type');
+$forum_enabled = Util::isModuleEnabled('Forum');
 
 $smarty->assign('HOME_TYPE', $home_type);
+$smarty->assign('FORUM_ENABLED', $forum_enabled);
 
-if ($home_type === 'news') {
+if ($home_type === 'news' && $forum_enabled) {
```
partydragen, line 40: Can't use forum_enabled here, The front_page_modules  is for any modules to use, so its also wrong that you renamed it to news

Like my server list module using the front_page_modules

partydragen (changes requested):

modules/Core/pages/home.php
partydragen: the entire frontpage stuff need a rework anyways

PR #3028, opened 2022-08-14
Change copy function to Clipboard API

PadowYT2:
As the [`execCommand`](https://developer.mozilla.org/en-US/docs/Web/API/document/execCommand) is deprecated and not recommended, this is a remake of it. Simple, small, also used in

https://github.com/NamelessMC/Nameless/blob/848a2e33af959ce8cc59b70a5a0e3614706352d7/error.tpl#L384


tadhgboyle:
`navigator` only exists when HTTPS is enabled, so this will not work on HTTP only sites

PadowYT2:
> `navigator` only exists when HTTPS is enabled, so this will not work on HTTP only sites

But then even, should we remove it from error.tpl or what?

tadhgboyle:
No, as you can see in the error.tpl file we check that it [navigator] is
not undefined first, and then use it if it exists. Otherwise it falls back
to the other way

Derkades:
Should we be using deprecated functions at all? I have seen sites do something like this instead:

```js
if (window.isSecureContext) {
    navigator.clipboard.writeText(url);
    // Display "URL copied to clipboard" toast
} else {
    // Create popup or toast with the URL, with a message that it should be copied manually
}

tadhgboyle:
Looks like there are some other places in the codebase that still use `execCommand`, if you could fix all of those, and use Derks suggestion of `window.isSecureContext` instead of `navigator.clipboard !== undefined`, that would be great.

tadhgboyle:
Being done in new PR alongside other changes

PR #3029, opened 2022-08-14
Convert error page to Fomantic UI

### LINKED ISSUE #2887: Update exception page to Fomantic UI
Currently the exception page is made with bootstrap, primarily because it's from the era of the old default template. 

Updating it to fomantic ui will make it more consistent with defaultrevamp, and remove the need for the bootstrap dependency

tadhgboyle:
image

Screen Shot 2022-08-14 at 13 38 45


PadowYT2 (changes requested):
Looks nice, maybe make the debug link a little higher, so it doesn't take its own space?

core/classes/Misc/ErrorHandler.php
```diff
@@ -126,8 +128,8 @@ public static function catchException(?Throwable $exception, ?string $error_stri
             'LANG_CHARSET' => defined('LANG_CHARSET') ? LANG_CHARSET : 'utf-8',
             'TITLE' => $language->get('errors', 'fatal_error') . ' - ' . $site_name,
             'SITE_NAME' => $site_name,
-            'BOOTSTRAP' => $path . 'vendor/bootstrap/dist/css/bootstrap.min.css',
-            'BOOTSTRAP_JS' => $path . 'vendor/bootstrap/dist/js/bootstrap.min.js',
+            'SEMANTIC_JS' => '/custom/templates/DefaultRevamp/js/fomantic.min.js',
```
samerton: Might be worth considering adding fomantic elsewhere in case the default revamp template is deleted

core/includes/error.tpl
```diff
@@ -0,0 +1,351 @@
...
+
+        {/if}
+
+    </div>
+
+    {if $CAN_GENERATE_DEBUG}
+        <div class="ui modal tiny" id="debug_link_modal">
+            <i class="close icon"></i>
+            <div class="header">
+                Generate a debug link
```
samerton: Should we be looking to translate all terms in this file? Maybe in the future?

tadhgboyle: Good catch

PadowYT2 (dismissed):

partydragen (approved):

PR #3030, opened 2022-08-15
Actually handle OAuth errors

supercrafter100:
When previously accessing the oauth page directly, an error is still thrown. This PR will fix this.

### DIFF
--- a/modules/Core/pages/oauth.php
+++ b/modules/Core/pages/oauth.php
@@ -4,11 +4,15 @@ const PAGE = 'oauth';
 require_once(ROOT_PATH . '/core/templates/frontend_init.php');
 
 if (!isset($_GET['provider'], $_GET['code'])) {
-    if (!array_key_exists($_GET['provider'], NamelessOAuth::getInstance()->getProvidersAvailable())) {
-        ErrorHandler::logWarning("Invalid provider {$_GET['provider']}");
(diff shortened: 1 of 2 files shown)


DominikSLK (dismissed):

modules/Core/pages/oauth.php
```diff
@@ -4,11 +4,15 @@
 require_once(ROOT_PATH . '/core/templates/frontend_init.php');
 
 if (!isset($_GET['provider'], $_GET['code'])) {
-    if (!array_key_exists($_GET['provider'], NamelessOAuth::getInstance()->getProvidersAvailable())) {
-        ErrorHandler::logWarning("Invalid provider {$_GET['provider']}");
-        Session::flash('home_error', $language->get('general', 'oauth_failed'));
-        Redirect::to(URL::build('/'));
-    }
+    ErrorHandler::logWarning('No provider or code set when accessing OAuth');
+    Session::flash('home_error', $language->get('general', 'oauth_failed'));
```
tadhgboyle: I think a different message here like "No OAuth data detected, ensure you are accessing this page via an OAuth redirect" would be more applicable than an error message

tadhgboyle (changes requested):

tadhgboyle (approved):

PR #3033, opened 2022-08-16
Keep inaccessible labels in a topic when editing them as a unauthorized user

### LINKED ISSUE #2999: Default labels go away on edit if user doesnt have access to change labels.
### Describe the issue

If you've put a default label on a forum and a user makes a topic in that forum, the label will be applied to the post. However if that user then edits their own post the label will go away as the user doesnt have access to change labels.

### NamelessMC Version

2.0.0

### To Reproduce

1. Set default label on a forum
2. Have user without permission to change labels create a topic
3. Have that user then edit the topic afterwards, doesnt even need to change anything just click "edit" and "submit".
4. Label goes away

### Expected Behaviour

For the label to not go away.

### Screenshots

Created new post with default label:
image

Editing the post:
image

Label gone:
image

### Additional Information

_No response_

supercrafter100:
Currently when the user doesn't have access to any labels, existing ones get removed when they edit and re-submit their topic. With this, it checks if the topic has any existing labels that the user doesn't have access to. If this is the case, those inaccessible labels get added to the new labels array as the user cannot edit them.

Should fix #2999


modules/Forum/pages/forum/edit.php
```diff
@@ -77,6 +76,25 @@
     Redirect::to(URL::build('/forum/topic/' . urlencode($post_id)));
 }
 
+/**
+ * @param array $labels  of label ids
+ * @param array $user_groups array of a user their group ids
+ * @return array An array of the ids of the labels the user has access to
+ */
+function getAccessibleLabels(array $labels, array $user_groups): array {
```
supercrafter100: This doesn't feel like the right location for a function but I didn't know where else to put it... So I just put it here 😅

samerton: Probably the Forum class? modules/Forum/classes

modules/Forum/classes/Forum.php
```diff
@@ -668,4 +668,23 @@ public function getAnySubforums(int $forum_id, array $groups = [0], int $depth =
 
         return $ret;
     }
+
+    /**
+     * @param array $labels  of label ids
+     * @param array $user_groups array of a user their group ids
+     * @return array An array of the ids of the labels the user has access to
+     */
+    static function getAccessibleLabels(array $labels, array $user_groups): array {
```
samerton: Can we add the `public` access modifier here please - always good to be specific

tadhgboyle (approved):

PR #3051, opened 2022-08-23
Add option to use OAuth linking method for Discord Integration

partydragen:


custom/panel_templates/Default/integrations/discord/discord.tpl
```diff
@@ -84,6 +84,15 @@
                                         <input class="form-control" type="number" name="discord_guild_id"
                                             id="inputDiscordId" value="{$DISCORD_GUILD_ID_VALUE}">
                                     </div>
+                                    <div class="form-group">
```
tadhgboyle, line 87: We should either have a tool tip here to tell them where to setup discord oauth, or not display the `"oauth"` option unless it is enabled

modules/Discord Integration/classes/DiscordIntegration.php
```diff
@@ -20,12 +20,24 @@ public function __construct(Language $language) {
...
-        $token = uniqid('', true);
+        $link_method = Util::getSetting('integration_link_method', 'bot', 'Discord Integration');
+        if ($link_method == 'oauth') {
+            // Link with oauth
+            Session::put('oauth_method', 'link_integration');
 
-        $integrationUser = new IntegrationUser($this);
-        $integrationUser->linkIntegration($user, null, null, false, $token);
+            $providers = NamelessOAuth::getInstance()->getProvidersAvailable();
+            $provider = $providers['discord'];
```
tadhgboyle, line 29: If they have not setup the discord oauth provider this will throw an exception

partydragen: the allowLinking function checking if its enabled https://github.com/NamelessMC/Nameless/pull/3051/files#diff-0062210913b52e0bbdd8092ccf7057751329e958b53afef49165a937a9fab82aR139

custom/panel_templates/Default/integrations/discord/discord.tpl
partydragen: Well that setting is temporially until the integrations settings page gets a proper settings page with allows integrations to add fields and handle it

adding tool tip here being same as adding tool tip for Discord Bot aswell to see if its setup or not, default method is even bot with may not be setup

And we still have the option as you discussed earlier to remove the Bot linking, there is alot of people who copy/paste the command into discord, but the commands system not detect it as they write it as text instead so they're leaking the code so other people can fast link their account before they do as many don't understand they need to write /verify manually

tadhgboyle (dismissed):

tadhgboyle (approved):

tadhgboyle:
Should we possibly use this PR to remove the `/verify` Discord command and instead always use OAuth for linking Discord?

partydragen:
> Should we possibly use this PR to remove the `/verify` Discord command and instead always use OAuth for linking Discord?

Well idk i fell setting up oauth2 is a bit complicated for less advanced users 
but this do solve the issue of people posting the verify key in the chat channel etc, as currently copy/paste the command into discord chat is currently not supported it being posted as a normal chat message

thats a mistake ALOT of people don't understand and they leak the code  with allow alot of users to claim the code before they understand they cannot copy/paste the entire command

tadhgboyle:
Yeah there are certainly pros and cons either way

PR #2877, opened 2022-06-19
Add dark mode toggle switch

PikaMug:
Shown in #2826 not urgent / low priority

Debating whether to remove TOGGLE_DARK_MODE_TEXT or show it on hover

PR #3054, opened 2022-08-26
Place dark mode toggle inline

PikaMug:
On mobile devices (and other low-width screens), footer content is auto-centered, presumably for aesthetics. This PR allows that property to be applied to the new Dark Mode toggle introduced in #2877

Also removes newly unused TOGGLE_DARK_MODE_TEXT


tadhgboyle:
Could you attach a before/after screenshot comparison here?

PikaMug:
comparison

tadhgboyle (approved):

PR #3065, opened 2022-08-31
Delete reported user on deletion

partydragen:


tadhgboyle:
Line 28 here defines that when users are deleted from the users table, any
matching records will be deleted from the reports table. Is this not
currently working?

core/migrations/20220519153139_create_reports_table.php

partydragen:
> 

Might be old installation, without the foreign keys

samerton (approved):

PR #3067, opened 2022-09-01
add more versions to choose from in issue template

supercrafter100:


PadowYT2:
Why not just make 2.0.x?

Derkades:
It's useful to know what minor version was used to reproduce a bug, for example if we think we've already fixed that bug in a later minor version

Derkades (approved):

PR #3073, opened 2022-09-12
Add new webhook events for profile posts and profile post replies

Coookei:
Hi! This small pull request simply adds two new webhook events so that people can be notified of when someone posts on a profile or replies on a profile.

There are already most events such as registration, new forum topics and topic replies so I think this adds to the set and provides more insight into what is going on on the website from Discord.


modules/Core/module.php
```diff
@@ -478,6 +478,14 @@ public function __construct(Language $language, Pages $pages, User $user, Naviga
...
         );
 
+        EventHandler::registerEvent('userNewProfilePost',
+            $language->get('admin', 'user_new_profile_post_hook_info')
+        );
+
+        EventHandler::registerEvent('userProfilePostReply',
+            $language->get('admin', 'user_profile_post_reply_hook_info')
+        );
+
```
supercrafter100: You are missing the values that the event returns and their description. These should be 

- username
- content
- content_full
- avatar_url
- title
- url

modules/Core/pages/profile.php
```diff
@@ -108,6 +108,18 @@
                                 ]
                             );
 
+                            EventHandler::executeEvent('userNewProfilePost', [
+                                'username' => $user->getDisplayname(true),
+                                'content' => $language->get("user", "x_posted_on_y_profile", [
```
supercrafter100: This essentially doesn't contain the content though? Just more or less an alert text. I'd say the name of it doesn't fit in that case.

modules/Core/pages/profile.php
```diff
@@ -178,6 +190,18 @@
                                 ]
                             );
 
+                            EventHandler::executeEvent('userProfilePostReply', [
+                                'username' => $user->getDisplayname(true),
+                                'content' => $language->get("user", "x_replied_on_y_profile", [
```
supercrafter100: same as other comment

supercrafter100 (changes requested):

modules/Core/module.php
Coookei: Thanks - I've updated the events to show the return values and descriptions now.

modules/Core/pages/profile.php
Coookei: Sorry, I don't quite understand what you mean. The 'content' field is just so that it ends up in the embed footer: https://github.com/NamelessMC/Nameless/blob/develop/modules/Discord%20Integration/hooks/DiscordHook.php#L30 
Unless you meant something else? Thanks.

supercrafter100: Ah right. Forgot it was used for that. Nvm my comment then

supercrafter100 (dismissed):

modules/Core/pages/profile.php
```diff
@@ -108,6 +108,18 @@
...
+                            EventHandler::executeEvent('userNewProfilePost', [
+                                'username' => $user->getDisplayname(true),
+                                'content' => $language->get("user", "x_posted_on_y_profile", [
+                                    "poster" => $user->getDisplayname(),
+                                    "user" => $query->username
+                                ]),
+                                'content_full' => strip_tags(str_ireplace(['<br />', '<br>', '<br/>'], "\r\n", Input::get('post'))),
+                                'avatar_url' => $user->getAvatar(128, true),
+                                'title' => $language->get("user", "new_profile_post"),
+                                'url' => URL::getSelfURL() . 'profile/' . urlencode($profile_user->getDisplayname(true)) . '/#post-' . urlencode(DB::getInstance()->lastId())
```
partydragen: You need to use URL::Build after getSelfURL and remove / from it at the start, This does not support unfriendly urls

Use this `URL::getSelfURL() . ltrim(URL::build('/'profile/' . urlencode($profile_user->getDisplayname(true)) . '/#post-' . urlencode(DB::getInstance()->lastId())), '/'),`

modules/Core/pages/profile.php
```diff
@@ -178,6 +190,18 @@
...
+                            EventHandler::executeEvent('userProfilePostReply', [
+                                'username' => $user->getDisplayname(true),
+                                'content' => $language->get("user", "x_replied_on_y_profile", [
+                                    "replier" => $user->getDisplayname(),
+                                    "user" => $query->username
+                                ]),
+                                'content_full' => strip_tags(str_ireplace(['<br />', '<br>', '<br/>'], "\r\n", Input::get('reply'))),
+                                'avatar_url' => $user->getAvatar(128, true),
+                                'title' => $language->get("user", "profile_post_reply"),
+                                'url' => URL::getSelfURL() . 'profile/' . urlencode($profile_user->getDisplayname(true)) . '/#post-' . urlencode($_POST['post'])
```
partydragen: Same here

modules/Core/pages/profile.php
```diff
@@ -108,6 +108,18 @@
                                 ]
                             );
 
+                            EventHandler::executeEvent('userNewProfilePost', [
```
partydragen, line 112: Currently you use user language, you need to use default language

use `$default_language = new Language('core', DEFAULT_LANGUAGE);`

modules/Core/pages/profile.php
```diff
@@ -178,6 +190,18 @@
                                 ]
                             );
 
+                            EventHandler::executeEvent('userProfilePostReply', [
```
partydragen, line 195: Currently you use user language, you need to use default language

use `$default_language = new Language('core', DEFAULT_LANGUAGE);`

partydragen (changes requested):

modules/Core/pages/profile.php
Coookei: Thanks I've applied these changes.

modules/Core/pages/profile.php
Coookei: Sorry I don't understand what you mean :c

tadhgboyle: Resolved

tadhgboyle (approved):

modules/Core/pages/profile.php
Coookei: Thanks - I understand what partydragen meant now

partydragen (approved):

PR #3075, opened 2022-09-12
Show correct server id when it is updated

### LINKED ISSUE #3045: Bug on update information
### Describe the issue

When we do an update in the core and the page updates the information is not updated on the page at the first update we have to quit and then come back

### NamelessMC Version

2.0.0

### To Reproduce

1. Go to the admin panel
2. Now go to the integration discord
3. Update or add new id discord
4. Send form

### Expected Behaviour

_No response_

### Screenshots

_No response_

### Additional Information

_No response_

supercrafter100:
Because the in-memory cache has not refreshed when the discord id just got updated on the page, it is retrieved from the posted id when this is set so the user sees the actual updated discord id instead of the old one.

### DIFF
--- a/modules/Discord Integration/classes/Discord.php
+++ b/modules/Discord Integration/classes/Discord.php
@@ -15,11 +15,6 @@ class Discord {
     private static bool $_is_bot_setup;
 
     /**
-     * @var string|null The ID of this website's Discord server
-     */
-    private static ?string $_guild_id;
-
-    /**
      * @var Language Instance of Language class for translations
      */
     private static Language $_discord_integration_language;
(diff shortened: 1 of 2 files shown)


Derkades:
Huh? Doesn't it run `Util::setSetting('discord', Input::get('discord_guild_id'));`, which also updates the cache?

supercrafter100:
The method that gets the guild id is the following: 
```php

public static function getGuildId(): ?string {
        if (!isset(self::$_guild_id)) {
            self::$_guild_id = (string) Util::getSetting('discord');
        }

        return self::$_guild_id;
    }
```

As you can see once it's been requested once (before it's updated) it won't update it again during the request cycle so the old guild id is still stored in there if the discord id is updated. Until the page is refreshed at which point that cache is refreshed and the correct guild id is displayed

Derkades:
I see, this caching was probably implemented before Util::getSetting had it's own caching. Now, I think it's better to remove caching from getGuildId() to make the implementation simpler and avoid similar issues in the future

Derkades (approved):

PR #3076, opened 2022-09-12
Allow browser scripts to bypass force_2fa redirects

supercrafter100:
This is needed because scripts on page load to /queries and other places get automatically redirected to the enable_2fa page. This means a new secret is generated each time that page is requested by a script. As a result of this the 2fa secret that is displayed to the user is incorrect so when they proceed to enter their code, it is always incorrect as the secret got changed.

This PR solves this issue by checking if the `X_REQUESTED_WITH` is set. This header is not set by broswers but only by the request methods from scripts, this means if the user requests a page, they'll be redirected but if a script does it, it works fine. Thus the 2fa secret doesn't get regenerated.


tadhgboyle (dismissed):

tadhgboyle (approved):
🚢

PR #3077, opened 2022-09-14
Add missing $ to fix undefined constant error

supercrafter100:


tadhgboyle (approved):
Good catch, I should maybe see how to enable this check for phpstan

Derkades:
Do we need to create a release/2.0.3 branch and cherry-pick this commit to it?

tadhgboyle:
Yes I think this could call for a 2.0.3 release

PR #3080, opened 2022-09-19
Update TFA page to use a post request to disable TFA

PadowYT2:


Aberdeener:
Is there a discord conversation I missed that explains why this is needed?
Things seem to work already, and the settings PHP file looks for a `$_GET` action, not a post action.
Did you mean to also convert the backend to use a post request?

tadhgboyle (approved):

PR #3086, opened 2022-09-30
Cleanup some frontend JS

tadhgboyle:


custom/templates/DefaultRevamp/js/core/user.js
```diff
@@ -54,9 +46,9 @@ if (loggedIn == 1) {
     function updatePMs(data) {
         if (data.value > 0) {
             $("#button-pms").removeClass('default').addClass("red");
-            var pms_list = '';
-            for (var i in data.pms) {
-                pms_list += '<a class="item" href="' + URLBuild('user/messaging/?action=view&amp;message=' + data.pms[i].id) + '">' + data.pms[i].title + '</a>';
+            let pms_list = '';
+            for (const pm of data.pms) {
```
Derkades, line 50: are you sure it should be changed to `of` (for lists) instead of `in` (for object keys)?

samerton: This looks right considering how we're using `pm` on line 51 👍

samerton (approved):

custom/templates/DefaultRevamp/js/core/user.js
Derkades: I see now, thanks

PR #3090, opened 2022-09-30
Add link/unlink integration endpoints

partydragen:


tadhgboyle:
Would naming them link/unlink instead of add/remove make sense?

partydragen:
> Would naming them link/unlink instead of add/remove make sense?

Done

tadhgboyle (approved):

Derkades:
Any changes to the API must be made in a new minor release, since all patch releases are currently considered to be identical by API clients

partydragen:
2.0.x is mirror releases it can contain small things as sam also told

currently the entire develop can be in the 2.0.3 release

also this is not api changes, it just improvements by adding missing endpoints with i also need for my module it being a more bigger issue by including the files into core by modules with then end up being start override the newer versions files with older ones

Derkades:
2.0.x can indeed contain minor new features, but with the important exception of API changes. Or at least, any clients using the Nameless Java API will only be able to use the new endpoints after 2.1, because all 2.0.x versions are considered identical in API functionality. In addition, using these new APIs in your modules would break the promise of all modules for 2.0.x working on all 2.0.x versions. Or do we only provide backwards compatibility, not forwards compatibility for patch releases? 

Can't we just release 2.1.0 earlier?

PR #3092, opened 2022-09-30
Add NamelessMC OAuth & Integration

partydragen:
Add NamelessMC OAuth & Integration

Requires NamelessMC website to run OAuth2 Module, Currently users are not able to create applications self within the module ¯\_(ツ)_/¯


tadhgboyle:
Not sure if this fits into native Nameless, would be rarely used. Feels like a module to me

PR #3097, opened 2022-10-06
Add ability to change password of users via StaffCP

### LINKED ISSUE #3093: Add set password field to userCP
### Describe the feature

Because some people do not want to setup phpmailer, I think it would be useful to allow administrators to set a user their password manually to what they want. This way they can set it to something they can share with the person who forgot their password and then those people can change their password again. 

Currently when a user forgets their password, if phpmailer is not setup there is no way to reset the user their password as far as I'm aware.

supercrafter100:
Closed #3093 

Preview:
image


tadhgboyle:
What is the red box?

We also def want a confirmation password input too. Could these two inputs be on their own floating container? It looks a little wonky to just have them thrown at the bottom

supercrafter100:
> What is the red box?
> 
> We also def want a confirmation password input too. Could these two inputs be on their own floating container? It looks a little wonky to just have them thrown at the bottom

Red box is a browser password extension, nothing important. I'll make them on their own floating container

supercrafter100:
New workflow: image

partydragen:
Amazing good work :D

modules/Core/pages/panel/users_edit.php
```diff
@@ -210,19 +210,42 @@
...
+
+            if ($validation->passed()) {
+                $password = Input::get('password');
+                $encrypted_password = password_hash($password, PASSWORD_BCRYPT, ['cost' => 13]);
+                $view_user->update([
+                    'password' => $encrypted_password
+                ]);
+                Session::flash('edit_user_success', $language->get('admin', 'user_password_changed_successfully'));
+            } else {
+                Session::flash('edit_user_error', implode('\n', $validation->errors()));
```
tadhgboyle, line 248: Hmm, this feels really dirty. Not sure there is anything else we can do here though... Revamping how we handle success/error messages to simplify everything would be nice eventually:tm:

supercrafter100: I can't really think of a better way to do this currently

modules/Core/pages/panel/users_edit.php
```diff
@@ -210,19 +210,42 @@
...
-                if ($user_query->id > 1) {
-                    EventHandler::executeEvent('deleteUser', [
-                        'user_id' => $user_query->id,
-                        'username' => $user_query->username,
-                        'email_address' => $user_query->email
-                    ]);
 
-                    Session::flash('users_session', $language->get('admin', 'user_deleted'));
-                }
+        } else if ((Input::get('action') == 'delete') && $user_query->id > 1) {
```
Aberdeener, line 214:
```suggestion
        } else if (Input::get('action') == 'delete' && $user_query->id > 1) {
```

tadhgboyle (approved):

partydragen (approved):

PR #3098, opened 2022-10-07
Extract Minecraft features into module

tadhgboyle:


core/classes/Database/DatabaseInitialiser.php
```diff
@@ -117,6 +122,10 @@ private function initialiseModules(): void {
                 'name' => 'Cookie Consent',
                 'priority' => 10
             ],
+            [
+                'name' => 'Discord Integration',
```
PadowYT2, line 126: `discord`

PadowYT2 (changes requested):

modules/Minecraft Integration/module.php
```diff
@@ -0,0 +1,293 @@
...
+
+    private Language $_language;
+    private Language $_minecraft_language;
+
+    public function __construct(Language $language, Language $minecraft_language, Pages $pages, Endpoints $endpoints) {
+        $this->_language = $language;
+        $this->_minecraft_language = $minecraft_language;
+
+        $name = 'Minecraft Integration';
+        $author = '<a href="https://tadhg.sh" target="_blank" rel="nofollow noopener">Aberdeener</a>';
```
partydragen, line 13: Well being wrong to call it made by you, When its made by Samerton well NamelessMC, Same with the Discord Module

You did many changes to Discord but like this Minecraft it all about moving existing code but that does not mean you made it

I think it better that all Modules & Templates being author NamelessMC, as it made from all and the link could be namelessmc website or github contributors page

modules/Minecraft Integration/language/en_UK.json
```diff
@@ -0,0 +1,3 @@
+{
+    "minecraft_integration/test": "Test"
```
partydragen, line 2: not a fan of the long "minecraft_integration" should just be general or something, or modify language file to accept 1 string with then mean its no file its legacy ¯\_(ツ)_/¯

modules/Minecraft Integration/module.php
tadhgboyle: Yes, was going to change this :p

PR #3100, opened 2022-10-07
Allow members to reset their avatar

### LINKED ISSUE #3042: Add option for user to reset their avatar
### Describe the feature

If a user decide to upload a custom avatar he can no longer go back to the default one (that one generated by mc-heads or other providers), and only the staff could do that.
I therefore propose to add the option to reset the personal avatar.

supercrafter100:
This adds the option for users to reset their custom avatar. Closes #3042 though this should probably not be merged before a major release as this will contain breaking changes for front-end templates (the addition of the button)


supercrafter100:
I also cleaned up the file a bit to get rid of so many nested if () {} else { if() {} } things. Else if exits

custom/templates/DefaultRevamp/user/settings.tpl
```diff
@@ -194,6 +194,14 @@
                     <input type="submit" class="ui primary button" value="{$SUBMIT}">
                 </form>
             </div>
+            <div class="ui segment">
```
samerton: Might be worth only showing this if the user actually has an avatar?

samerton (changes requested):

custom/templates/DefaultRevamp/user/settings.tpl
supercrafter100: Yep that makes sense

samerton (approved):

Issue #2690, opened 2022-04-22
Mechanism to forcefully log a user out


samerton:
### Describe the feature

Add a mechanism to forcefully log a user out, potential use cases:

- Logging a user out after flagging them as spammer
- Logging all sessions out for a user in case their account is compromised
- More?

PR #3102, opened 2022-10-07
Ability to force logout all sessions of user via staffcp

supercrafter100:
Add the ability to logout all sessions of a user via staffcp. Only if the target user does not have access to staffcp

### DIFF
--- a/custom/languages/en_UK.json
+++ b/custom/languages/en_UK.json
@@ -650,6 +650,8 @@
     "admin/user": "User",
     "admin/user_deleted": "User deleted successfully.",
     "admin/user_id": "User ID",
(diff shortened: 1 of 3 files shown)


supercrafter100:
would close #2690 I guess? Though technically this functionality was added with the session rewrite a while ago (the ability to force logout all sessions)

tadhgboyle:
Hmm... I think this would be nicer to have once we have the page which lists all their sessions with device names and login times, rather than just throwing a new button onto this page.

partydragen:
> Hmm... I think this would be nicer to have once we have the page which lists all their sessions with device names and login times, rather than just throwing a new button onto this page.

agree

PR #3106, opened 2022-10-11
Create Members module with member list page

tadhgboyle:
image
image
image

- Allows modules to register custom member lists to display on the member list page
- Allows site admins to toggle display of each member list
- Allows modules to add some metadata about each member list user to display underneath their name

TODO:
- make nav item a dropdown:
  - links for active member lists (under main header)
  - header for "groups"
  - all active groups under this header


core/installation/includes/upgrade_perform.php
```diff
@@ -173,7 +173,7 @@
...
...admincp.core.reactions":1,"admincp.core.registration":1,"admincp.core.social_media":1,"admincp.core.terms":1,"admincp.errors":1,"admincp.members":1,"admincp.integrations":1,"admincp.discord":1,"admincp.minecraft":1,"admincp.minecraft.authme":1,"admincp.minecraft.servers":1,"admincp.minecraft.query_errors":1,"admincp.minecraft.banners":1,"admincp.modules":1,"admincp.pages":1,"admincp.security":1,"admincp.security.acp_logins":1,"admincp.security.template":1,"admincp.styles":1,"admincp.styles.panel_templates":1,"admincp.styles.templates":1,"admincp.styles.templates.edit":1,"admincp.styles.images":1,"admincp.update":1,"admincp.users":1,"admincp.users.edit":1,"admincp.groups":1,"admincp.groups.self":1,"admincp.widgets":1,"modcp.ip_lookup":1,"modcp.punishments":1,"modcp.punishments.warn":1,"modcp.punishments.ban":1,"modcp.punishments.banip":1,"modcp.punishments.revoke":1,"modcp.reports":1,"modcp.profile_banner_reset":1,"usercp.messaging":1,"usercp.signature":1,"admincp.forums":1,"usercp.private_profile":1,"usercp.nickname":1,"usercp.profile_banner":1,"profile.private.bypass":1, "admincp.security.all":1,"admincp.core.hooks":1,"admincp.core.emails_mass_message":1,"usercp.gif_avatar":1}']);
```
samerton, line 176: Unrelated to this PR but I wonder if we should stop supporting upgrades from v1 -> v2.1.0 to remove this overhead (i.e. v1 sites should go to v2.0.x first)

custom/panel_templates/Default/members/member_lists.tpl
```diff
@@ -0,0 +1,116 @@
...
+                        <!-- Success and Error Alerts -->
+                        {include file='includes/alerts.tpl'}
+
+                        <form id="toggleHideBannedUsers" action="" method="post">
+                            <input type="hidden" name="token" value="{$TOKEN}">
+                            <div class="form-group custom-control custom-switch">
+                                <input type="hidden" name="action" value="toggle_hide_banned_users">
+                                <input id="inputToggleHideBannedUsers" name="hide_banned_users" type="checkbox" class="custom-control-input js-check-change" value="1" {if $HIDE_BANNED_USERS_VALUE eq 1} checked{/if} />
+                                <label for="inputToggleHideBannedUsers" class="custom-control-label">
+                                    {$HIDE_BANNED_USERS}
```
samerton: Probably not a day one request but maybe we could make this configurable per member list?

custom/templates/DefaultRevamp/members/members.tpl
```diff
@@ -0,0 +1,213 @@
...
+                                <i class="search icon"></i>
+                                <input type="text" name="search" minlength="3" placeholder="Name..." autocomplete="off">
+                            </div>
+                        </form>
+                    </div>
+                </div>
+            </div>
+            <div class="ui fluid card">
+                <div class="content">
+                    <h4 class="ui header">View by Group</h4>
```
samerton: Can this be translated?

custom/templates/DefaultRevamp/members/members.tpl
```diff
@@ -0,0 +1,213 @@
...
+                    <div class="description">
+                        <select class="ui selection fluid dropdown" onchange="viewGroup(this)">
+                            <option value="">Group...</option>
```
samerton: And this 😄

modules/Core/module.php
```diff
@@ -635,6 +637,14 @@ public function __construct(Language $language, Pages $pages, User $user, Naviga
         Email::addPlaceholder('[Greeting]', static fn(Language $viewing_language) => $viewing_language->get('emails', 'greeting'));
         Email::addPlaceholder('[Message]', static fn(Language $viewing_language, string $email) => $viewing_language->get('emails', $email . '_message'));
         Email::addPlaceholder('[Thanks]', static fn(Language $viewing_language) => $viewing_language->get('emails', 'thanks'));
+
+        MemberList::getInstance()->registerListProvider(new StaffMembersListProvider($language));
+
+        MemberList::getInstance()->registerMemberMetadataProvider(function (User $member) use ($language) {
+            return [
+                $language->get('general', 'joined') => date('d M Y', $member->data()->joined),
```
samerton: Should we be using the date format constant here?

modules/Forum/classes/HighestReactionScoresMemberListProvider.php
```diff
@@ -0,0 +1,22 @@
...
+        $this->_name = 'highest_reaction_scores';
+        $this->_friendly_name = $language->get('forum', 'highest_reaction_scores');
+        $this->_module = 'Forum';
+        $this->_icon = 'star icon';
+    }
+
+    protected function generateMembers(): array {
+        $members = [];
+
+        $query = DB::getInstance()->query('SELECT COUNT(user_received) AS `count`, user_received FROM nl2_forums_reactions WHERE reaction_id = 1 GROUP BY user_received ORDER BY `count` DESC');
```
samerton: Wonder if we should be joining on the reactions table to get all positive reaction types instead of just reaction_id = 1

modules/Forum/classes/HighestReactionScoresMemberListProvider.php
```diff
@@ -0,0 +1,22 @@
...
+        foreach ($query->results() as $result) {
+            $members[] = [new User($result->user_received), $result->count];
```
samerton: Is there a limit to the number of users we'll be getting? Creating a new instance of User each time will be very resource heavy

modules/Forum/classes/MostPostsMemberListProvider.php
```diff
@@ -0,0 +1,22 @@
...
+        $this->_module = 'Forum';
+        $this->_icon = 'sort numeric up icon';
+    }
+
+    protected function generateMembers(): array {
+        $members = [];
+
+        $query = DB::getInstance()->query('SELECT post_creator, COUNT(post_content) AS `count` FROM nl2_posts GROUP BY post_creator ORDER BY `count` DESC');
+        foreach ($query->results() as $result) {
+            $members[] = [new User($result->post_creator), $result->count];
```
samerton: Same question regarding limiting this as we are creating a new instance of User each time

modules/Forum/module.php
```diff
@@ -177,6 +177,29 @@ public function __construct(Language $language, Language $forum_language, Pages
...
+                        [$member->data()->id]
+                    )->first()->count,
+            ];
+        });
+
+        MemberList::getInstance()->registerMemberMetadataProvider(function (User $member) use ($forum_language) {
+            return [
+                $forum_language->get('forum', 'reaction_score') =>
+                    DB::getInstance()->query(
+                        'SELECT COUNT(user_received) AS `count` FROM nl2_forums_reactions WHERE reaction_id = 1 AND user_received = ?',
```
samerton: Same Q as above RE joining on reactions table to get all positive reactions instead of just reaction id 1

modules/Members/classes/MemberList.php
```diff
@@ -0,0 +1,73 @@
...
+            throw new RuntimeException("Provider '$name' does not exist");
+        }
+
+        return $this->_lists[$name];
+    }
+
+    public function getMembersInGroup(int $group_id): array {
+        return MemberListProvider::parseMembers(
+            array_map(
+                static fn ($row) => new User($row->id),
```
samerton: Same Q RE instantiating new User class for a potentially large dataset

modules/Members/language/en_UK.json
```diff
@@ -0,0 +1,11 @@
+{
+  "members/members": "Members",
+  "members/new_members": "New members",
+  "members/no_members": "No members found.",
+  "members/member_lists": "Member Lists",
+  "members/member_list_hide_banned_users": "Hide banned users?",
+  "members/member_list_search_min_3_chars": "Please ensure your search query is at least 3 characters long.",
+  "members/member_list_search_no_results": "No user found for '{{search}}'.",
+  "members/member_list_toggled": "Successfully {{value}} {{list}} list.",
```
samerton: Might this make sense to split into two terms (i.e. one for the positive case and one for the negative case), in case two different languages have different sentence structures in each case?

modules/Members/pages/members.php
```diff
@@ -0,0 +1,92 @@
...
+ */
+
+if (count(MemberList::getInstance()->allEnabledLists()) == 0) {
+    Redirect::to(URL::build('/'));
+}
+
+if (Input::exists()) {
+    if (Token::check()) {
+        $search = Input::get('search');
+        if (strlen($search) < 3) {
```
samerton: Was there something which came up recently about OG Minecraft usernames having 2 characters? Maybe we could decrease this to 2 if so

samerton: Also maybe we could rate limit this since this is a public facing query

modules/Members/queries/member_list.php
```diff
@@ -0,0 +1,17 @@
+<?php
+
+header('Content-type: application/json;charset=utf-8');
+
+$list = $_GET['list'];
+
+if (str_starts_with($list, 'group_')) {
```
samerton, line 19: Is this response cached anywhere?

samerton (dismissed):

modules/Core/module.php
tadhgboyle: Great idea

modules/Forum/classes/HighestReactionScoresMemberListProvider.php
tadhgboyle: This is 100% a potential issue...
Originally I had it so we passed a `limit` int down to the `generateMembers` method on each list provider, but that caused an issue when we then filtered the returned members, for example:
- View the overview lists page (only 5 members displayed per list)
- The `StaffMemberListProvider` returns 5 users
- They have enabled the "hide banned users", so we filter out any banned users returned from the `StaffMemberListProvider`, which could make it < 5

The alternative would be to pass another param to the `generateMembers` method to delegate the filtering of members to each of the list providers but repeating that logic and relying on modules to properly handle it seems wrong... any ideas?

custom/templates/DefaultRevamp/members/members.tpl
```diff
@@ -0,0 +1,213 @@
...
+    const renderList = (name) => {
+        return function () {
+            const xhr = new XMLHttpRequest();
+            xhr.withCredentials = false;
+            xhr.open('GET', '{$QUERIES_URL}'.replace(
+                {literal}
+                '{{list}}',
+                {/literal}
+                name
+            ));
```
supercrafter100: Can we please stop using these ancient methods of requesting endpoints? Seeing as we don't support IE anymore we can use the fetch api, which results in way cleaner code, especially if we use async/await

supercrafter100 (dismissed):

modules/Members/language/en_UK.json
tadhgboyle: Great point

custom/templates/DefaultRevamp/members/members.tpl
tadhgboyle: Thank you! Always forget that is in the browser 🤦🏼‍♂️

modules/Forum/classes/HighestReactionScoresMemberListProvider.php
tadhgboyle: With the new system of list providers returning just a SQL string, lists load in < 100ms on a site seeded with 5000 users... hopefully this is good!

partydragen (dismissed):

partydragen (dismissed):

custom/panel_templates/Default/members/member_lists.tpl
```diff
@@ -0,0 +1,158 @@
...
+                            </div>
+                        </form>
+
+                        <form action="" method="post">
+                            <input type="hidden" name="token" value="{$TOKEN}">
+                            <div class="form-group">
+                                <label for="groups">{$GROUPS}</label>
+                                <select name="groups[]" id="groups" class="form-control" multiple>
+                                    {foreach from=$GROUPS_ARRAY item=group}
+                                        <option value="{$group->id}" {if in_array($group->id, $GROUPS_VALUE)} selected{/if}>{$group->name}</option>
```
samerton: Should group name be purified before printing out here?

custom/templates/DefaultRevamp/members/members.tpl
```diff
@@ -0,0 +1,228 @@
...
+            </div>
+            {if $GROUPS|count}
+                <div class="ui fluid card">
+                    <div class="content">
+                        <h4 class="ui header">{$VIEW_GROUP}</h4>
+                        <div class="description">
+                            <select class="ui selection fluid dropdown" onchange="viewGroup(this)">
+                                <option value="">{$GROUP}</option>
+                                {foreach from=$GROUPS item=group}
+                                    <option value="{$group->id}" {if $VIEWING_GROUP->id == $group->id} selected {/if}>{$group->name}</option>
```
samerton: Same Q RE sanitising group name

custom/templates/DefaultRevamp/members/members.tpl
```diff
@@ -0,0 +1,228 @@
...
+                    </div>
+                </div>
+            </div>
+        </div>
+        <div class="ui column">
+            {if $VIEWING_LIST == "group" || $MEMBER_LISTS_VIEWING|count}
+                <div class="ui stackable equal width left aligned three column grid segment" style="margin-top: 0">
+                    {if $VIEWING_LIST == "group"}
+                        <div class="ui column">
+                            <h3>{$VIEWING_GROUP->name}</h3>
```
samerton: Same Q as above

modules/Core/queries/users.php
```diff
@@ -1,18 +1,20 @@
 <?php
 // Searchable user list
-if (!$user->isLoggedIn()) {
-    die(json_encode(['error' => 'Unauthenticated']));
-}
-
-if (!isset($_GET['search']) || strlen($_GET['search']) < 3) {
-    die(json_encode(['error' => 'Please enter a search query of at least 3 characters']));
+if (!isset($_GET['search']) || strlen($_GET['search']) < 2) {
```
samerton, line 16: I wonder if we should look at rate limiting this query

modules/Members/classes/MemberListProvider.php
```diff
@@ -0,0 +1,179 @@
...
+            if (count($list_members) === $limit) {
+                break;
+            }
+
+            $user_id = $row->{$id_column};
+            $member = new User($user_id);
+
+            $list_members[] = array_merge(
+                [
+                    'username' => $member->data()->username,
```
samerton: Do you think this needs escaping?

modules/Members/classes/MemberListProvider.php
```diff
@@ -0,0 +1,179 @@
...
+                    'avatar_url' => $member->getAvatar(),
+                    'group_style' => $member->getGroupStyle(),
+                    'profile_url' => $member->getProfileURL(),
+                    'count' => $count_column ? $row->{$count_column} : null,
+                ],
+                $overview
+                    ? []
+                    : [
+                        'group_html' => implode('', $member->getAllGroupHtml()),
```
samerton: Assuming this gets displayed as a list of all users' groups, should this be passed through to the template as an array so the template can separate the group tags however it wants

tadhgboyle: Great idea

modules/Core/queries/users.php
tadhgboyle: Added - 60 rq/m.
Also added a default `limit 5` clause to the sql query.

core/installation/includes/upgrade_perform.php
tadhgboyle: Totally agree about removing overhead, annoying needing to duplicate a bunch of logic here and in `DatabaseInitialiser`.
We could probably find a way to continue supporting upgrades but clean up the code...

modules/Members/module.php
```diff
@@ -0,0 +1,104 @@
...
+                $members_order = $cache->retrieve('members_order');
+            }
+
+            $cache->setCache('navbar_icons');
+            if (!$cache->isCached('members_icon')) {
+                $members_icon = '';
+            } else {
+                $members_icon = $cache->retrieve('members_icon');
+            }
+            $navs[0]->add('members', $this->_member_language->get('members', 'members'), URL::build('/members'), 'top', null, $members_order, $members_icon);
```
partydragen: Setting to change link location

modules/Members/pages/members.php
```diff
@@ -0,0 +1,122 @@
...
+if (isset($error)) {
+    $smarty->assign([
+        'ERROR_TITLE' => $language->get('general', 'error'),
+        'ERROR' => $error,
+    ]);
+}
+
+$groups = [];
+foreach (json_decode(Util::getSetting('member_list_viewable_groups', '{}', 'Members'), true) as $group_id) {
+    $group = Group::find($group_id);
```
partydragen, line 56: Group may be deleted

modules/Core/queries/users.php
```diff
@@ -1,18 +1,30 @@
...
-    die(json_encode(['error' => 'Please enter a search query of at least 3 characters']));
+// Searchable user list
+if (!isset($_GET['search']) || strlen($_GET['search']) < 2) {
+    die(json_encode(['error' => 'Please enter a search query of at least 2 characters']));
 }
 
 $query = '%' . $_GET['search'] . '%';
 
-$users = DB::getInstance()->query('SELECT id, username, nickname FROM nl2_users WHERE username LIKE ? OR nickname LIKE ?', [
+$users = DB::getInstance()->query('SELECT id, username, nickname, gravatar, email, has_avatar, avatar_updated FROM nl2_users WHERE username LIKE ? OR nickname LIKE ? LIMIT 5', [
```
partydragen: Not sure if it should be limited to 5 what else used this queries page to get users private message suggestions? should maybe be get param instead &limit=5 if isset

partydragen (changes requested):

modules/Members/module.php
tadhgboyle: Thanks

tadhgboyle: Adding settings page in StaffCP for this and other config options

core/classes/Avatars/AvatarSource.php
```diff
@@ -89,7 +89,7 @@ public static function getAvatarFromUserData(object $data, bool $allow_gifs = fa
             }
         }
 
-        return "https://avatars.dicebear.com/api/initials/{$data->username}.png?size={$size}";
+        return "https://api.dicebear.com/5.x/initials/svg?seed={$data->username}&size={$size}";
```
partydragen: It was switched to .png for discord reasons?`now its back to svg?

tadhgboyle: Accidentally changed when I copied new URL from their docs 🤦🏼‍♂️

partydragen (approved):

PR #3107, opened 2022-10-13
Add limit setting to latest posts widget

### LINKED ISSUE #2862: Latest Posts Widget should have limit options, not be forced to use 5
### Describe the feature

The Latest Posts Widget has a limit of 5 hardcoded into the template so that only 5 posts can ever be shown. It would be great if I could choose from more options, such as 5, 10, 15. Forums with higher activity will need to show more.

partydragen:
Closes https://github.com/NamelessMC/Nameless/issues/2862

### DIFF
--- a/modules/Forum/widgets/admin/latest_posts.php
+++ b/modules/Forum/widgets/admin/latest_posts.php
@@ -1,0 +1,32 @@
+<?php
+/*
+ *  Made by Partydragen
+ *  https://github.com/NamelessMC/Nameless/
+ *  NamelessMC version 2.1.0
+ *
+ *  License: MIT
+ *
+ *  Latest posts widget settings
+ */
+
+if (Input::exists()) {
+    $errors = [];
+
+    if (Token::check(Input::get('token'))) {
+        if (isset($_POST['limit']) && $_POST['limit'] > 0) {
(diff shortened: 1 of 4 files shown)


modules/Forum/widgets/admin/latest_posts.php
```diff
@@ -0,0 +1,32 @@
...
+ *  https://github.com/NamelessMC/Nameless/
+ *  NamelessMC version 2.1.0
+ *
+ *  License: MIT
+ *
+ *  Latest posts widget settings
+ */
+
+if (Input::exists()) {
+    $errors = [];
```
supercrafter100, line 13: You are never actually assigning the errors to smarty?

supercrafter100: same with the success message

partydragen: > You are never actually assigning the errors to smarty?

It being done by widgets page, this is not a actuall page but a sub file with add the data to the widgets settings page

supercrafter100: ah that explains

supercrafter100 (approved):

samerton (approved):

PR #3109, opened 2022-10-14
Add Integrations settings system & Setting to turn off mc username requirement

### LINKED ISSUE #3043: Disable Minecraft username requirement without disabling Minecraft integration
### Describe the feature

I don't know if this is already a feature or if there is another way to do it but when Minecraft Integration is enabled it makes the user sign up with a Minecraft Username. Though you can make it so it doesn't check the validity of the account, wouldn't it be better to have an option where they can create a regular account and link there Minecraft account to it? This way they have an option to link or not and it's easier to verify.

I have three ideas for this:
1. When they create an account, just like how they would link there Discord to the account maybe they can request a code from the website and use it to verify on the server. 

2. When they type /link on the server, a link would popup and when they click on said link it would redirect them to the website, where if they are signed in it would verify there account, or give them an option to sign in.

3. I dont know if this one is possible but maybe use the Microsoft account popup(The way people sign into there Minecraft accounts), and link/verify that way.

partydragen:
Add integrations settings system like how widgets & templates settings work also added the requested feature to turn off minecraft username requirement on registration while keeping the minecraft integration enabled

Closes https://github.com/NamelessMC/Nameless/issues/3043

minecraft
discord

### DIFF
--- a/modules/Core/pages/panel/integrations.php
+++ b/modules/Core/pages/panel/integrations.php
@@ -71,12 +71,18 @@ if (!isset($_GET['integration'])) {
         }
     }
 
-    if ($integration->getSettings() !== null && file_exists($integration->getSettings())) {
-        require_once($integration->getSettings());
+    if ($integration->getSettings() !== null) {
+        if (file_exists($integration->getSettings())) {
+            require_once($integration->getSettings());
+        } else {
+            $errors[] = $language->get('admin', 'integration_settings_does_not_exist', [
+                'integration' => Output::getClean($integration->getName())
+            ]);
+        }
     }
 
     $smarty->assign([
-        'EDITING_INTEGRATION' => $language->get('admin', 'editing_integration_x', ['integration' => $integration->getName()]),
+        'EDITING_INTEGRATION' => $language->get('admin', 'editing_integration_x', ['integration' => Output::getClean($integration->getName())]),
         'BACK' => $language->get('general', 'back'),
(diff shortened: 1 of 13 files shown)


samerton (dismissed):

tadhgboyle (dismissed):

modules/Core/pages/panel/integrations.php
```diff
@@ -56,19 +56,25 @@
...
+
+                Session::flash('integrations_success', $language->get('admin', 'integration_updated_successfully'));
+                Redirect::to(URL::build('/panel/core/integrations/', 'integration=' . $integration->getName()));
+            }
         } else {
             $errors[] = $language->get('general', 'invalid_token');
         }
     }
 
+    if ($integration->getSettings() !== null && file_exists($integration->getSettings())) {
```
supercrafter100: Wouldn't it make more sense to throw an error if the integration settings file doesn't exist? Rather than just silently ignoring it. I feel like this can cause some frustration for developers that forget it.  And the settings file should never be configured anyways if the file doesn't exist. This can result in unwanted behavior in case a file was not uploaded correctly

supercrafter100 (changes requested):

modules/Core/pages/panel/integrations.php
partydragen: Done

supercrafter100 (approved):

PR #3111, opened 2022-10-17
Add semantic config and change node_modules copy method

### LINKED ISSUE #3110: Add semantic config file so it can install without interaction
Prevent this:
```
? Set-up Semantic UI (Use arrow keys)
❯ Automatic (Use default locations and all components) 
  Express (Set components and output folder) 
  Custom (Customize all src/dist values)
```

For context, currently the CI is broken because of this (and also the development docker images, and as a result named hosting development installations)

samerton:
Closes #3110


Derkades:
No reviews but I'll just assume this is fine, since I would really like named hosting to not be broken anymore

Derkades (approved):

PR #3112, opened 2022-10-19
Allow HTML in announcements

TeemoCell:


Derkades:
Instead of allowing all HTML, this should probably use HTMLPurifier like other instances where HTML is configured in StaffCP

partydragen:
Should also be switched to use the tinymc editor instead of people write the html themself

TeemoCell:
Well that is the frontend only.

PR #3125, opened 2022-11-30
Update

Justman10000:


supercrafter100:
I feel like this is unnecessary? It gives info about the permission and its **always** a dangerous permission to give. Even if the user is trustworthy it does give them essentially full power over the website.

Justman10000:
> I feel like this is unnecessary? It gives info about the permission and its **always** a dangerous permission to give. Even if the user is trustworthy it does give them essentially full power over the website.

Yes, but if the user is trustworthy, is he abusing his power?

partydragen:
no reason to break all languages term just for a term keywork

supercrafter100:
> Yes, but if the user is trustworthy, is he abusing his power?
I never said they weren't. I'm just saying that the permission is and will always be dangerous to give. No matter if they're trustworthy or not. If they ever end up turning against you they could nuke your entire website in minutes when they've got administration. You should be very aware of this so "it can be dangrous" isn't clear enough if you ask me. It **IS** dangerous!

Justman10000:
> no reason to break all languages term just for a term keywork

?

Justman10000:
> > Yes, but if the user is trustworthy, is he abusing his power?
> 
> I never said they were abusing powers. I'm just saying that the permission is and will always be dangerous to give. No matter if they're trustworthy or not. If they ever end up turning against you they could nuke your entire website in minutes when they've got administration. You should be very aware of this so "it can be dangrous" isn't clear enough if you ask me. It **IS** dangerous!

Hmm, probably true again

Justman10000:
> no reason to break all languages term just for a term keywork

What difference does it make? As long as it is already done and works...

Justman10000:
8f8e5a6 - Better?

supercrafter100:
Well what's wrong with the original one? 😅 I don't really see the point in changing it at all

Justman10000:
> Well what's wrong with the original one? 😅 I don't really see the point in changing it at all

Mine sounds a bit better...

Derkades:
Thank you for contributing. As others have said I agree that this rewording is not any better and not worth forcing re-translation for all languages

PR #3128, opened 2022-12-04
Misc fixes

tadhgboyle:
Fixes:
- Sorting in StaffCP user list
- Loading forum page
- CLI installer
- Twemoji base URL

Adds:
- Check during installer that does not continue installation if the DB contains `nl2_modules` table already

### DIFF
--- a/core/classes/Minecraft/MCQuery.php
+++ b/core/classes/Minecraft/MCQuery.php
@@ -395,7 +395,7 @@ class MCQuery {
      * @return string The closest Minecraft colour code to the given RGB value
      */
     private static function getColor(string $rgb): string {
-        if (!str_contains($rgb, '#')) {
+        if (strpos($rgb, '#') === 0) {
             $rgb = substr($rgb, 1);
         }
 
--- a/core/classes/Minecraft/MCQuery.php
+++ b/core/classes/Minecraft/MCQuery.php
@@ -395,7 +395,7 @@ class MCQuery {
      * @return string The closest Minecraft colour code to the given RGB value
      */
     private static function getColor(string $rgb): string {
-        if (strpos($rgb, '#') === 0) {
+        if (!str_contains($rgb, '#')) {
             $rgb = substr($rgb, 1);
         }
 
--- a/core/classes/Misc/Text.php
+++ b/core/classes/Misc/Text.php
@@ -154,7 +154,8 @@ class Text {
      * @return string Text with emojis replaced with URLs to their Twemoji equivalent.
      */
     public static function renderEmojis(string $text): string {
-        return Twemoji::text($text)->toHtml(null, [
+        // We can remove this ->base(...) call when https://github.com/Astrotomic/php-twemoji/pull/13 is merged
+        return Twemoji::text($text)->base('https://cdn.jsdelivr.net/gh/twitter/twemoji@latest/assets')->toHtml(null, [
             'width' => 20,
             'height' => 20,
             'style' => 'vertical-align: middle;'
--- a/core/installation/steps/database_configuration.php
+++ b/core/installation/steps/database_configuration.php
@@ -37,7 +37,12 @@ if ($_SERVER['REQUEST_METHOD'] == 'POST') {
 
         try {
             // This throws a PDOException if the connection fails
-            DB::getCustomInstance($db_address, $db_name, $db_username, $db_password, $db_port, $force_charset=$db_charset);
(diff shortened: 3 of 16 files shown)


Derkades:
Not all changes can/should be backported, for example the path change in the db seeder only applies to 2.1.0

tadhgboyle:
> Not all changes can/should be backported, for example the path change in the db seeder only applies to 2.1.0

Removed that patch from this PR

modules/Forum/classes/Forum.php
```diff
@@ -450,6 +449,7 @@ public function updateTopicLatestPosts(int $topic_id): void {
                 'topic_last_user' => $latest_post->post_creator,
             ]);
         } else {
+            // Undefined variable $forum_id, what should we do?
             $this->_db->update('forums', $forum_id, [
```
tadhgboyle, line 453: cc @samerton this seems to be from the performance improvement stuff about a month ago. What should be done?

core/classes/Minecraft/MCQuery.php
```diff
@@ -395,7 +395,7 @@ private static function getMotd(string $text, array $modern_format): string {
      * @return string The closest Minecraft colour code to the given RGB value
      */
     private static function getColor(string $rgb): string {
-        if (strpos($rgb, '#') === 0) {
+        if (!str_contains($rgb, '#')) {
```
samerton: The logic is inverted here (ie should be `if (str_contains(...`, but I think this is probably best to stay as it was before, since we're checking for a hex code starting with # and taking the remaining characters if so.

modules/Forum/classes/Forum.php
samerton: Oops, I think we'll either need to perform an additional SQL query here to get the forum_id column from nl2_topics based off the $topic_id parameter, _or_ update everywhere we call this function and pass $forum_id in as a second parameter

samerton (changes requested):

core/classes/Minecraft/MCQuery.php
tadhgboyle: Good catch

modules/Forum/classes/Forum.php
tadhgboyle: It looks like `nl2_posts` has a `forum_id` column, I think if we used that column this would work?

samerton: Problem is that this is the case where the query returns no results, so I think it'd have to be passed into this function as a parameter

tadhgboyle: Ohh haha, facepalm, sure that makes sense. I will update the pr to do so

core/installation/steps/database_configuration.php
```diff
@@ -37,7 +37,12 @@
 
         try {
             // This throws a PDOException if the connection fails
-            DB::getCustomInstance($db_address, $db_name, $db_username, $db_password, $db_port, $force_charset=$db_charset);
+            $db = DB::getCustomInstance($db_address, $db_name, $db_username, $db_password, $db_port, $force_charset=$db_charset);
```
tadhgboyle, line 40: `$force_charset=$db_charset`
Neither variable here exists, and the `=` is clearly not the correct syntax here.
What is intended behavior?

I think we should always force charset of `utf8mb4` and this param can probably be removed.

modules/Forum/classes/Forum.php
tadhgboyle: @samerton can you take another look here when you have a chance?
Everywhere except `spam.php` had access to the `forum_id` of the `topic`, so I allowed `forum_id` to be null in the case of `spam.php`. Seems like a code smell and can be investigated and improved later on:tm:

core/classes/Misc/Text.php
```diff
@@ -154,7 +154,8 @@ public static function bold(string $text): string {
      * @return string Text with emojis replaced with URLs to their Twemoji equivalent.
      */
     public static function renderEmojis(string $text): string {
-        return Twemoji::text($text)->toHtml(null, [
+        // We can remove this ->base(...) call when https://github.com/Astrotomic/php-twemoji/pull/13 is merged
```
tadhgboyle, line 157: PR for reference https://github.com/Astrotomic/php-twemoji/pull/13

modules/Forum/classes/Forum.php
samerton: Sounds good, we'll revisit to query for forum_id if not provided based off topic_id

samerton (approved):

PR #3130, opened 2022-12-06
Apply rate limiting to forgot password

supercrafter100:
this will move the email check to the validate class and apply a rate limit. I haven't been able to test these changes though to my knowledge, this should work.

### DIFF
--- a/modules/Core/pages/forgot_password.php
+++ b/modules/Core/pages/forgot_password.php
@@ -28,7 +28,7 @@ if (!isset($_GET['c'])) {
             $validation = Validate::check([
                 'email' => [
                     Validate::REQUIRED => true,
-                    Validate::RATE_LIMIT => $rate_limit
+                    Validate::RATE_LIMIT => [1, 60] // 1 attempt every 60 seconds
                 ]
             ])->messages([
                 'email' => [
--- a/modules/Core/pages/forgot_password.php
+++ b/modules/Core/pages/forgot_password.php
@@ -23,9 +23,21 @@ if (!isset($_GET['c'])) {
     // Enter email address form
     if (Input::exists()) {
         if (Token::check()) {
-            if (!isset($_POST['email']) || empty($_POST['email'])) {
-                $error = $language->get('user', 'email_required');
-            } else {
+
+            $rate_limit = [1, 60]; // 1 attempt every 60 seconds
+            $validation = Validate::check([
+                'email' => [
+                    Validate::REQUIRED => true,
+                    Validate::RATE_LIMIT => $rate_limit
+                ]
+            ])->messages([
+                'email' => [
+                    Validate::REQUIRED => $language->get('user', 'email_required'),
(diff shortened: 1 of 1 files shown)


modules/Core/pages/forgot_password.php
```diff
@@ -23,9 +23,21 @@
...
         if (Token::check()) {
-            if (!isset($_POST['email']) || empty($_POST['email'])) {
-                $error = $language->get('user', 'email_required');
-            } else {
+
+            $rate_limit = [1, 60]; // 1 attempt every 60 seconds
+            $validation = Validate::check([
+                'email' => [
+                    Validate::REQUIRED => true,
+                    Validate::RATE_LIMIT => $rate_limit
```
tadhgboyle:
```suggestion
                    Validate::RATE_LIMIT => [1, 60] // 1 attempt every 60 seconds
```

partydragen:
only 1? should be 2 atleast as users can often test several emails or writed wrong

modules/Core/pages/forgot_password.php
supercrafter100: Cant have a ; at the end as we're in an array. I'll see if I can do that tomorrow

tadhgboyle: Ah typo

tadhgboyle (dismissed):

modules/Core/pages/forgot_password.php
```diff
@@ -61,6 +72,8 @@
                 }
 
                 $success = $language->get('user', 'forgot_password_email_sent');
+            } else {
+                $error = join('<br />', $validation->errors());
```
samerton, line 76: Note to self, change this to an array in the future

modules/Core/pages/forgot_password.php
```diff
@@ -23,9 +23,20 @@
...
+
+            $validation = Validate::check([
+                'email' => [
+                    Validate::REQUIRED => true,
+                    Validate::RATE_LIMIT => [1, 60] // 1 attempt every 60 seconds
+                ]
+            ])->messages([
+                'email' => [
+                    Validate::REQUIRED => $language->get('user', 'email_required'),
+                    Validate::RATE_LIMIT => fn($meta) => $language->get('general', 'rate_limit', $meta)
```
samerton: This could be a static function (see aber's changes to login.php in his PR: https://github.com/NamelessMC/Nameless/pull/3128/files#diff-4be83c7d87065b9b6e0404325a3c8d6a33f61b96904d4686c0280bdeaa61c5fa)

supercrafter100:
> only 1? should be 2 atleast as users can often test several emails or writed wrong

The whole  point is to not allow spamming though. "often test several emails" is obviously what we are trying to prevent here. I'll make it two though to at least give the option to correct yourself

tadhgboyle (approved):

samerton (approved):

PR #3136, opened 2022-12-12
StaffCP performance improvements

samerton:


modules/Core/collections/panel/RecentPunishments.php
```diff
@@ -109,10 +108,6 @@ public function getContent(): string {
                         'revoked_by_profile' => ($revoked_by_user ? URL::build('/panel/user/' . urlencode($revoked_by_user->data()->id) . '-' . urlencode($revoked_by_user->data()->username)) : ''),
                         'revoked_at' => $timeago->inWords($item->revoked_at, $this->_language)
                     ];
-
-                    if (++$i == 5) {
-                        break;
-                    }
```
tadhgboyle, line 115: Good catch 👏🏼

modules/Core/module.php
```diff
@@ -1618,6 +1608,30 @@ public static function addDataToDashboardGraph($title, $data): void {
...
     }
 
+    /**
+     * Fill missing days in an array containing graph data
+     *
+     * @param array $data Initial data to display on graph
+     * @param string $key Key to output data under
+     * @return array Formatted data under $key
+     */
+    public static function fillMissingGraphDays(array $data, string $key): array {
```
tadhgboyle, line 1618: Not a huge deal, but I think this would better fit in the main `Collection` class, rather than requiring other modules to use the `Core_Module` class

tadhgboyle (approved):
Some non blocking comments

modules/Core/module.php
samerton: I was debating this but in the end decided against moving it out - currently we have other graph related code in here which could also do with moving out, which was a bit out of scope for this performance improvements piece.

Would be nice to have a class specifically for graphs in the future 😄

PR #3137, opened 2022-12-13
Update

Justman10000:


core/classes/Core/TimeAgo.php
```diff
@@ -59,7 +59,7 @@ public function inWords($past, Language $language): string {
             $timeDifference < (($this->_secondsPerMinute * 89) + 29)
         ) {
             // between 44mins30secs and 1hour29mins29secs
```
Derkades, line 61: Between roughly 45 minutes and 90 minutes should say "about 1 hour", not "more than an hour". The previous strings were correct. Same goes for other strings you changed.

core/classes/Core/TimeAgo.php
```diff
@@ -128,7 +128,7 @@ public function inWords($past, Language $language): string {
             )
         ) {
             // between 29days23hours59mins30secs and 59days23hours59mins29secs
-            $key = 'about_1_month';
+            $key = 'more_then_1_hour';
```
Derkades, line 131: Accidentally copied `more_then_1_hour` the string?

Derkades:
In my opinion, the previous strings were more accurate.

Justman10000:
Since when does About Etwa mean in German? That's something completely new... Then are the German translations incorrectly

Justman10000:
Have the translations fixed

PR #3138, opened 2022-12-14
Show message if there was an error loading a widget

samerton:


tadhgboyle (approved):
This is really great, but could it go in for 2.0.x?
I guess the main BC here is the change to Widgets constructor, but this is an internal class so I don't think we risk breaking any modules

partydragen (approved):

Derkades:
> This is really great, but could it go in for 2.0.x? I guess the main BC here is the change to Widgets constructor, but this is an internal class so I don't think we risk breaking any modules

Does it break if a third party template is missing the `widgets/widget_error.tpl` file?

tadhgboyle:
> > This is really great, but could it go in for 2.0.x? I guess the main BC here is the change to Widgets constructor, but this is an internal class so I don't think we risk breaking any modules
> 
> Does it break if a third party template is missing the `widgets/widget_error.tpl` file?

Ah good point

PR #3143, opened 2022-12-27
Catch exception when installing modules

samerton:
Also add custom term for Translation file not found

### DIFF
--- a/modules/Core/pages/panel/modules.php
+++ b/modules/Core/pages/panel/modules.php
@@ -28,6 +28,7 @@ if (!isset($_GET['action'])) {
     $modules = DB::getInstance()->get('modules', ['id', '<>', 0])->results();
     $enabled_modules = Module::getModules();
 
+    $errors = [];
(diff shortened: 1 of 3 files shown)


modules/Core/pages/panel/modules.php
```diff
@@ -45,7 +46,21 @@
                 continue;
             }
 
-            require_once(ROOT_PATH . '/modules/' . $item->name . '/init.php');
+            try {
+                require_once ROOT_PATH . '/modules/' . $item->name . '/init.php';
+            } catch (Exception $e) {
+                $term = 'unable_to_load_module';
+
+                if ($e->getMessage() == 'Translation file not found') {
```
tadhgboyle:
```suggestion
                if ($e->getMessage() === 'Translation file not found') {
```

tadhgboyle (dismissed):

tadhgboyle (approved):

PR #3146, opened 2022-12-28
Code cleanup

XYZHolder:


supercrafter100:
I feel like we don’t need a pr that is the code run through an auto formatter. We already have phpstan to enforce some code styling

supercrafter100:
As seen by phpstan failing, it doesn’t agree 😅

supercrafter100:
Code style check I mean*

Derkades:
We unfortunately can't accept large reformatting PRs, because they are impossible to review (way more work than to just do it ourselves).

PR #3154, opened 2023-01-04
Change to new /configure command

Derkades:
Nameless-Link has a new /configure command to replace all previous administrative commands

image


tadhgboyle (approved):
Looks good - do any auto responses also need updating?

Derkades:
Good point, I will update those when the hosted bot is running the new version.

PR #3159, opened 2023-01-08
Integrity checking system

Derkades:
## The problem

Many user reported issues are caused by missing files, extra files, modified files, files from different Nameless versions, or corrupt in some other way.

## The solution

Ship a checksums file with NamelessMC, so the integrity of files can be verified. Files that are expected to change or be edited (e.g. cache, uploads, custom templates and modules) are ignored.

## Workflow

Run `php -f dev/scripts/generate_checksums.php` to generate the checksums file to be included in a release

Run `php -f dev/scripts/verify_checksums.php` to verify checksums using the checksums file (after preparing the release, so removing directories like `node_modules`, `.git`, etc.)

Ideally StaffCP would also run this with a cache, and display a warning in the "Server compatibility" section if there are issues. I did not implement that for now.


core/classes/Misc/IntegrityChecker.php
```diff
@@ -0,0 +1,131 @@
+<?php
+
+// str_starts_with() is PHP 8, polyfill for PHP 7.4 compatibility
+// source: Laravel Framework
+// https://github.com/laravel/framework/blob/8.x/src/Illuminate/Support/Str.php
+if (!function_exists('str_starts_with')) {
```
samerton: This should already be polyfilled via https://github.com/NamelessMC/Nameless/blob/develop/composer.json#L26

Derkades: Thanks

Derkades:
I don't know how to fix (or ignore?) the "undefined constant ROOT_PATH" issue phpstan is reporting

core/classes/Misc/IntegrityChecker.php
```diff
@@ -0,0 +1,147 @@
...
+        'modules/Cookie Consent',
+        'modules/Core',
+        'modules/Discord Integration',
+        'modules/Forum',
+    ];
+
+    /**
+     * @return bool If the file should be ignored from integrity checking, according to IGNORED_PATHS and INCLUDED_PATHS.
+     */
+    private static function is_ignored($path): bool {
```
tadhgboyle: Can we swap these to camel case; isIgnored?

tadhgboyle (dismissed):

core/classes/Misc/IntegrityChecker.php
Derkades: right, I forgot functions where snake case but methods camel case

Derkades:
I have no idea why phpstan is failing, it works locally...
image

Derkades:
I think phpstan itself is crashing. It does not say problems were found with the code, it says that it *itself* crashed with a link to create a bug report. That also explains why ignoring the line doesn't help.

tadhgboyle:
I really like this idea, and it would be cool to push it out eventually.
Since we want to display this information to users in StaffCP, I think it would take up a lot of space so it would not work well in the Server Compatibility widget - could we consider making a more general "Website Health" page (name can be hashed out)?

samerton (approved):

PR #3164, opened 2023-01-11
Remove FollowSymLinks rule from htaccess

Derkades:
doesn't seem needed, and causes issues with some hosting providers

### DIFF
--- a/.htaccess
+++ b/.htaccess
@@ -1,33 +1,33 @@
-  <IfModule mod_negotiation.c>
-    Options -MultiViews
-  </IfModule>
-  Options -Indexes
-  RewriteEngine on
-  IndexIgnore *
+<IfModule mod_negotiation.c>
+  Options -MultiViews
+</IfModule>
+Options -Indexes
+RewriteEngine on
+IndexIgnore *
 
-  <IfModule mod_headers.c>
-    Header append X-FRAME-OPTIONS "SAMEORIGIN"
-  </IfModule>
+<IfModule mod_headers.c>
+  Header append X-FRAME-OPTIONS "SAMEORIGIN"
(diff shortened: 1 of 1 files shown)


.htaccess
```diff
@@ -1,34 +1,33 @@
...
+  Header append X-FRAME-OPTIONS "SAMEORIGIN"
+</IfModule>
 
-  RewriteCond %{REQUEST_URI}::$1 ^(.*?/)(.*)::\2$
-  RewriteRule ^(.*)$ - [E=BASE:%1]
+RewriteCond %{REQUEST_URI}::$1 ^(.*?/)(.*)::\2$
+RewriteRule ^(.*)$ - [E=BASE:%1]
 
-  RewriteCond %{REQUEST_FILENAME} !-d
+RewriteCond %{REQUEST_FILENAME} !-d
```
tadhgboyle, line 15:
```suggestion
```

samerton:
As per discussion on Discord, `FollowSymLinks` was added to solve the following error
```
[Sat Jan 15 21:53:10.492773 2022] [rewrite:error] [pid 68711] [client <ip>] AH00670: Options FollowSymLinks and SymLinksIfOwnerMatch are both off, so the RewriteRule directive is also forbidden due to its similar ability to circumvent directory restrictions : /var/www/html/index.php, referer: http://<url>/index.php?route=/
```

Possible solution for Plesk users: https://support.plesk.com/hc/en-us/articles/115001125289-A-website-hosted-in-Plesk-shows-500-Internal-Server-Error-Option-FollowSymLinks-not-allowed-here

Derkades:
Thanks for the link, I have added it to the FAQ

PR #3165, opened 2023-01-11
Remove code scheduled for removal in 2.1.0

Derkades:


core/classes/Core/Pages.php
```diff
@@ -150,16 +150,6 @@ public function getPageByURL(string $url): ?array {
         return null;
     }
 
-    /**
-     * Get the page details the user currently viewing.
-     * @deprecated  Not used internally. Will be removed in 2.1.0
-     *
-     * @return array Details of current page.
-     */
-    public function getActivePage(): array {
```
partydragen: This must stay! Also used by all my modules

core/classes/Core/Paginator.php
```diff
@@ -179,17 +179,4 @@ public function generate(int $links, string $href = '?'): string {
...
 
-    /**
-     * Set values of instance variables, alternative function (as they are set in getLimited()).
-     * @deprecated Not used internally. Will be removed in 2.1.0
-     *
-     * @param int $total
-     * @param int $limit
-     * @param int $page
-     */
-    public function setValues(int $total, int $limit, int $page): void {
```
partydragen, line 190: Also believe @samerton added this for the infractions module?

partydragen: hm not sure, i did a fast look on infractions module but could not find it, but pretty sure Sam added it for a reason somewhere unless it got removed again

but sam can answer when he see this

samerton: Not able to find a commit from the same time for Infractions, Resources, Tebex or Voting Plugin, and I can't remember why this was added. I reckon we go with removing it for now and any usages of the Paginator class can amend their implementation

tadhgboyle (approved):
Are we good to merge this now, given we are confident 2.1.0 is the next release?

samerton:
> Are we good to merge this now, given we are confident 2.1.0 is the next release?

Yep I'd say so 👍

PR #3170, opened 2023-01-12
Fix checkbox aligning vertically

supercrafter100:
This pr finally makes the checkbox properly align instead of being aligned lower. 
image

### DIFF
--- a/custom/templates/DefaultRevamp/css/custom.css
+++ b/custom/templates/DefaultRevamp/css/custom.css
@@ -947,13 +947,11 @@ select {
     height: auto;
 }
 
-.ui.checkbox .box:before,
-.ui.checkbox label:before {
+.ui.checkbox .box:before {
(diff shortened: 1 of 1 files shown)


Derkades:
Firefox on the left, Chromium on the right. With your PR, the checkbox looks fine on Chromium based browsers, but is now no longer aligned properly in Firefox.

image

supercrafter100:
> Firefox on the left, Chromium on the right. With your PR, the checkbox looks fine on Chromium based browsers, but is now no longer aligned properly in Firefox.
> 
> image

That's very weird. Because the additional css would push down the checkbox even more. So it would be even more broken on firefox then (even further pushed down). Are you sure its not some cached css in firefox?

Derkades:
Very smart, yes indeed it was just the cache!

Derkades (approved):

PR #3171, opened 2023-01-12
Don't decode before encoding for new installations

Derkades:
NamelessMC used to HTML-encode posts before saving them to the database. It no longer does. When this change was made, there was never a database migration to decode all encoded content. The reasoning for this was that this migration would be too slow.

Currently, before being encoded, content is always decoded first. This improves compatibility with content that was stored in encoded form, preventing it from being encoded twice. However, it also causes various issues for new content that is stored in non-encoded form.

Code snippets and more complex usage of special characters do not work properly.

With this change, the compatibility decoding has been made configurable. It is enabled by default for websites that upgrade from a previous version, and disabled (as it should always should be) for new installations.

Ideally it would only be enabled for websites that have upgraded from pr12, but I don't think there's any way to detect that?


modules/Core/hooks/ContentHook.php
```diff
@@ -10,7 +10,7 @@
 class ContentHook extends HookBase {
 
     public static function codeTransform(array $params = []): array {
-        if (parent::validateParams($params, ['content'])) {
+        if (Config::get('core.encoded_content_compat') && parent::validateParams($params, ['content'])) {
```
partydragen: Should not be disabled here, As its a function any module can use
Disable the listener for the forum content instead

modules/Core/hooks/ContentHook.php
```diff
@@ -39,7 +39,7 @@ public static function codeTransform(array $params = []): array {
     }
 
     public static function decode(array $params = []): array {
-        if (parent::validateParams($params, ['content'])) {
+        if (Config::get('core.encoded_content_compat') && parent::validateParams($params, ['content'])) {
```
partydragen: same here

partydragen (changes requested):

modules/Core/hooks/ContentHook.php
Derkades: I'd argue that codeTransform *should* be disabled for all modules using it. Otherwise, forum content will be fixed but other content will still be broken.

partydragen: Yet these 2 functions are functions modules may wanna use for other things

But yea at the same time it suck that modules must self disable it then but no other way to do it i guess other than making 2 new codeTransform and decode functions

partydragen: What you think @samerton

samerton: This function actually needs to be removed as per the following Discord conversation: https://discord.com/channels/246705793066467328/246705793066467328/1071149195165175950

However this will require the queue system to be implemented so we can decode all post content

samerton (approved):

partydragen (approved):

PR #3175, opened 2023-01-13
Fix captcha validation

supercrafter100:
This is a temporary solution as the validation for captcha's has changed and I currently don't have the time to properly dig into the validation of the captchas. So for now, disabling the validation should be fine.


samerton (approved):

samerton:
@supercrafter100 would you mind creating an issue so we revisit this

PR #3176, opened 2023-01-13
Fix post editing

supercrafter100:
Previously, if no existing topic labels existed, the site would crash due to the fact that it wouldn't return an empty array but null. This defaults the existing labels array to an empty array.

### DIFF
--- a/modules/Forum/pages/forum/edit.php
+++ b/modules/Forum/pages/forum/edit.php
@@ -150,12 +150,12 @@ if (Input::exists()) {
                     }
                     return $prev;
                 });
-                $accessible_labels = Forum::getAccessibleLabels($forum_labels, $user_groups);
+                $accessible_labels = Forum::getAccessibleLabels($forum_labels ?? [], $user_groups);
                 $existing_inaccessible_labels = array_diff($existing_labels, $accessible_labels);
 
                 // Get all the posted labels and see which ones the user can actually edit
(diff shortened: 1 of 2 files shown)


modules/Forum/pages/forum/edit.php
```diff
@@ -150,12 +150,12 @@
                     }
                     return $prev;
                 });
-                $accessible_labels = Forum::getAccessibleLabels($forum_labels, $user_groups);
+                $accessible_labels = Forum::getAccessibleLabels($forum_labels ?? [], $user_groups);
```
samerton: Instead of adding `?? []` to here and line 158, it should be possible to specify `[]`  as the third parameter to array_reduce

tadhgboyle:
This does not solve the problem at hand unfortunately, as the `getAccessibleLabels` method signature requires an `array`, and does not allow `null`.
Adding the default value to the `array_reduce` is better practice but does not solve the problem of `null` being passed where an `array` is expected

supercrafter100:
> This does not solve the problem at hand unfortunately, as the `getAccessibleLabels` method signature requires an `array`, and does not allow `null`. Adding the default value to the `array_reduce` is better practice but does not solve the problem of `null` being passed where an `array` is expected

Wouldn't the variable always have an array though? Seeing as the default value is now an array and not null

tadhgboyle:
Ah yes sorry was looking in wrong area

tadhgboyle (approved):

samerton (approved):

PR #3181, opened 2023-01-13
Make sure integration verify code is not null

partydragen:


tadhgboyle:
In what circumstance would the code be null?

tadhgboyle (approved):

partydragen:
> In what circumstance would the code be null?

If they manually link the integration from StaffCP or if they unverify it after it got verified in StaffCP
This is the issue many users have had, all releated to they messed with it from StaffCP

PR #3183, opened 2023-01-14
Fix footer with CSS to remove flicker on each page load

tadhgboyle:
Before:

https://user-images.githubusercontent.com/26070412/212441260-e2329a0b-98a2-4b72-ba83-388f1f00d3a9.mov

After:

https://user-images.githubusercontent.com/26070412/212441343-0428f0dd-a265-4b70-b160-a6b556222137.mov

Mobile display is the exact same still:
Screenshot 2023-01-13 at 16 32 04


Derkades:
Nice!

samerton (approved):

PR #3184, opened 2023-01-14
Simplify forum post quoting system

tadhgboyle:
Previously, to quote a post a user would click the "Quote" button, and then a "Insert Quotes" button would appear by their editor, which they would have to click again.

This let users insert multiple post quotes into their editor at the same time, but made the underlying javascript and php endpoint slightly more complex for no gain.

This PR makes the system more immediate, adding the quoted post content to the editor as soon as the button is clicked, as well as updating the `custom.css` file to display a border on the left beside the `<blockquote>` to make it more clear they are quoted.


custom/templates/DefaultRevamp/css/custom.css
```diff
@@ -295,6 +295,12 @@ body.pushable>.pusher {
     padding-bottom: 1em;
 }
 
+.blockquote {
+    border-left: 2px solid #ccc;
```
tadhgboyle, line 299: This colour and width is taken from the tinymce preview so that they match exactly when rendered later

modules/Forum/pages/forum/get_quotes.php
```diff
@@ -20,28 +20,18 @@
...
 
-$posts = [];
+$post = $forum->getIndividualPost($_GET['post']);
 
-foreach ($_POST['posts'] as $item) {
-    $post = $forum->getIndividualPost($item);
+$content = $post['content'];
 
-    $content = $post['content'];
-    $content = preg_replace('~<blockquote(.*?)>(.*)</blockquote>~si', '', $content);
```
tadhgboyle, line 33: This line caused issues when quoting a post which contained another quote. Removing it allows nested quotes like so:
Screenshot 2023-01-14 at 00 43 26

If there is a reason this has to stay, let me know

samerton (commented):
Looks good - I see the template change is minimal but could we get confirmation this wouldn't cause any breaking changes as a patch release?

tadhgboyle:
> Looks good - I see the template change is minimal but could we get confirmation this wouldn't cause any breaking changes as a patch release?

It should be okay - the quote(post_id) function is called the same, and the "insert quotes" button is hidden by default so it can't be clicked

samerton (approved):

PR #3186, opened 2023-01-14
Fix fallback to default template dark mode option

### LINKED ISSUE #3177: Night mode bug
### Describe the issue

dark mode column in nl2_users is no longer default null, it should allow null and be default null

NULL means the user will listen on website default mode, currently if the website is dark mode the user will get light mode when logging in due of this bug

### NamelessMC Version

Development version

### To Reproduce

_No response_

### Expected Behaviour

_No response_

### Screenshots

_No response_

### Additional Information

_No response_

samerton:
Closes #3177

### DIFF
--- a/core/migrations/20230114124352_users_night_mode_nullable.php
+++ b/core/migrations/20230114124352_users_night_mode_nullable.php
@@ -1,0 +1,15 @@
+<?php
+declare(strict_types=1);
+
+use Phinx\Migration\AbstractMigration;
+
+final class UsersNightModeNullable extends AbstractMigration
+{
+    public function change(): void
+    {
+        $table = $this->table('nl2_users');
+        $table->changeColumn('night_mode', 'boolean', ['length' => 1, 'null' => true, 'default' => null]);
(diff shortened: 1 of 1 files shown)


core/migrations/20230114124352_users_night_mode_nullable.php
```diff
@@ -0,0 +1,15 @@
...
+use Phinx\Migration\AbstractMigration;
+
+final class UsersNightModeNullable extends AbstractMigration
+{
+    public function change(): void
+    {
+        $table = $this->table('nl2_users');
+        $table->changeColumn('night_mode', 'boolean', ['length' => 1, 'null' => true, 'default' => null]);
+        $table->update();
+        $this->execute('UPDATE nl2_users SET night_mode = NULL WHERE night_mode = 0');
```
samerton, line 13: Decided to add this in so that users are correctly reset to default - no harm in a user disabling dark mode again if they explicitly turned it off previously

partydragen (approved):

PR #3191, opened 2023-01-15
Create release script

Derkades:
This script produces release zip files just like the ci. The advantage of running it locally is that you can do it on any branch (which will be needed for the upcoming 2.0.3 which is not based on the `develop` branch).

Thanks to Docker, it does not require a local installation of php, composer, or npm.

TODO: Generate checksum file, when #3159 is merged
TODO: Finish instructions

### DIFF
--- a/dev/scripts/release.py
+++ b/dev/scripts/release.py
@@ -1,0 +1,92 @@
+import subprocess
+from subprocess import PIPE
+import itertools
+import shutil
+
+
+EXCLUDE_DIRS = [
+    '.git',
+    '.github',
+    '.idea',
+    '.node-cache',
+    '.vscode',
+    'node_modules',
+    'release',
+]
+EXCLUDE_FILES = [
+    '.gitignore',
+    '.phpcs.xml',
+    'CHANGELOG.md',
+    'composer.json',
+    'composer.lock',
+    'CONTRIBUTING.md',
+    'docker-compose.yaml',
+    'Dockerfile.phpdoc',
+    'nginx.example',
+    'package.json',
+    'package-lock.json',
+    'phpstan.neon',
+    'postinstall.js',
+    'README.md',
+    'release.sh',
+    'yarn-postinstall.js',
+    'yarn.lock',
+    'SECURITY.md',
+    'semantic.json',
+    'web.config.example',
+]
+TARGET_DIR = 'release'
+
+
+def create_archives(archive_name: str):
+    zip_command = [
+        'zip',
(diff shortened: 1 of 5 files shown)


tadhgboyle:
Is it possible to extend this to add support for creating just the upgrade zip?
The upgrade zip contains only files that have changed since the last given version - including vendor files

Derkades:
Yes, I want to integrate your upgrade script with this so it can produce upgrade archives as well

dev/scripts/release.py
```diff
@@ -0,0 +1,121 @@
...
+
+EXCLUDE_DIRS = [
+    '.git',
+    '.github',
+    '.idea',
+    '.node-cache',
+    '.vscode',
+    'node_modules',
+    'release',
+]
```
tadhgboyle, line 17: I think we should add `cache` and `core/avatar/cache` here. Currently cache folder contents are copied to the generated zips which we don't want

dev/scripts/release.py
```diff
@@ -0,0 +1,121 @@
...
+
+    upgrade_temp = Path('release', 'upgrade_temp')
+
+    # Find previous tag
+    previous_tag_command = ['git', 'describe', '--abbrev=0', '--tags']
+    previous_tag = subprocess.check_output(previous_tag_command, shell=False)[:-1].decode()
+
+    # Find all files changed between previous tag and HEAD (current state)
+    changed_command = ['git', 'diff', previous_tag, 'HEAD', '--name-only', '--diff-filter=d']
+    changed_files = subprocess.check_output(changed_command, shell=False)[:-1].decode()
```
tadhgboyle, line 110: When I tested this locally it seemed to take all of the files... for this to work properly does it have to be run on the `develop` branch right after a new tag (ie `2.0.3`) was made?

dev/scripts/release.py
Derkades: `cache` contains an important `.htaccess` file, so I don't think it should be excluded

dev/scripts/release.py
Derkades: The upgrade package should include all files changed since the previous tag (so the tag for the upcoming release should not be created yet). I tested it on the `release/2.0.3` branch and it generated all changed files since the `2.0.2` tag.

dev/scripts/release.py
tadhgboyle: Makes sense, thanks!

tadhgboyle (approved):

PR #3192, opened 2023-01-15
Show user debug details in debug info

partydragen:
This allow us to validate if the user has the groups example if we're investigating their permissions problems to view form submissions example

### DIFF
--- a/modules/Core/queries/debug_link.php
+++ b/modules/Core/queries/debug_link.php
@@ -191,7 +191,7 @@ if ($user->isLoggedIn()) {
         $user_integrations[$integrationUser->getIntegration()->getName()] = [
             'username' => $integrationUser->data()->username,
             'identifier' => $integrationUser->data()->identifier,
-            'verified' => $integrationUser->data()->verified ? true : false
+            'verified' => $integrationUser->data()->verified
         ];
     }
 
@@ -208,7 +208,7 @@ $data = [
     'generated_at' => time(),
     'generated_by_name' => $user->data()->username,
     'generated_by_uuid' => $uuid,
-    "user" => $user_data,
+    'user' => $user_data,
     'namelessmc' => [
         'version' => $namelessmc_version,
         'update_available' => Util::getSetting('version_update') === 'urgent' || Util::getSetting('version_update') === 'true',
--- a/modules/Core/queries/debug_link.php
+++ b/modules/Core/queries/debug_link.php
@@ -179,10 +179,36 @@ foreach (['fatal', 'warning', 'notice', 'other', 'custom'] as $type) {
     $logs[$type] = file_exists($file_path) ? Util::readFileEnd($file_path, $max_bytes = 10_000) : '';
 }
 
+$user_data = [];
+if ($user->isLoggedIn()) {
+    $user_groups = [];
+    foreach ($user->getAllGroupIds() as $group_id) {
+        $user_groups[] = $group_id;
+    }
+
+    $user_integrations = [];
+    foreach ($user->getIntegrations() as $integrationUser) {
+        $user_integrations[$integrationUser->getIntegration()->getName()] = [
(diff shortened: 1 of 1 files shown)


partydragen:
```
    "user": {
        "id": 1,
        "username": "Partydragen",
        "nickname": "Partydragen",
        "groups": [
            2,
            8,
            7
        ],
        "integrations": {
            "Patreon": {
                "username": "Partydragen",
                "identifier": "5206074",
                "verified": true
            },
            "MC-Server-List": {
                "username": "Partydragen",
                "identifier": "1",
                "verified": true
            },
            "CubedCraft": {
                "username": "Partydragen",
                "identifier": "1",
                "verified": true
            },
            "MCStatistics": {
                "username": "Partydragen",
                "identifier": "1",
                "verified": true
            },
            "Discord": {
                "username": "Partydragen#0001",
                "identifier": "117749257569566723",
                "verified": true
            }
        }
    },
```

samerton:
Would it be more useful to return
```json
"groups": [
  {
    "id": 2,
    "name": "Admin"
  }
]
```
or similar?

partydragen:
> Would it be more useful to return
> 
> ```json
> "groups": [
>   {
>     "id": 2,
>     "name": "Admin"
>   }
> ]
> ```
> 
> or similar?

The debug info already have the group list, so the fancy debug page can already collect the group info from that list like several other things work there

modules/Core/queries/debug_link.php
```diff
@@ -179,10 +179,36 @@
     $logs[$type] = file_exists($file_path) ? Util::readFileEnd($file_path, $max_bytes = 10_000) : '';
 }
 
+$user_data = [];
+if ($user->isLoggedIn()) {
+    $user_groups = [];
+    foreach ($user->getAllGroupIds() as $group_id) {
+        $user_groups[] = $group_id;
```
tadhgboyle: Is this any different than doing "$user_groups = $user->getAllGroupIds();"?

modules/Core/queries/debug_link.php
```diff
@@ -179,10 +179,36 @@
...
+    }
+
+    $user_integrations = [];
+    foreach ($user->getIntegrations() as $integrationUser) {
+        $user_integrations[$integrationUser->getIntegration()->getName()] = [
+            'username' => $integrationUser->data()->username,
+            'identifier' => $integrationUser->data()->identifier,
+            'verified' => $integrationUser->data()->verified ? true : false
```
tadhgboyle: We probably don't need this ternary if "verified" returns a bool already

modules/Core/queries/debug_link.php
```diff
@@ -179,10 +179,36 @@
...
+        'groups' => $user_groups,
+        'integrations' => $user_integrations
+    ];
+}
+
 $data = [
     'generated_at' => time(),
     'generated_by_name' => $user->data()->username,
     'generated_by_uuid' => $uuid,
+    "user" => $user_data,
```
tadhgboyle: Single quotes!

tadhgboyle (changes requested):

modules/Core/queries/debug_link.php
partydragen: that returned stupid 8 => 8, 3 => 3 etc

tadhgboyle: array_values($user->getAllGroupIds());

partydragen: That worked :D

modules/Core/queries/debug_link.php
partydragen: Solved

modules/Core/queries/debug_link.php
partydragen: Solved

tadhgboyle (approved):

PR #3193, opened 2023-01-15
Add groups parameter to list users endpoint

### LINKED ISSUE #3052: ListUser end point does not return groups
### Describe the issue

If you make a request /users to get a list of multiple users (not UserInfo) it will not return the player's group.

From what I can see in ListUsersEndPoint.php there is a query for groups, but no groups are returned.

This currently does not affect much, but using Java-API with this function causes NullPointerException when requesting anything related to groups.

### NamelessMC Version

2.0.0

### To Reproduce

_No response_

### Expected Behaviour

_No response_

### Screenshots

_No response_

### Additional Information

_No response_

samerton:
Allows specifying `groups` GET parameter ie `/api/v2/users&groups`

Closes #3052

### DIFF
--- a/modules/Core/includes/endpoints/ListUsersEndpoint.php
+++ b/modules/Core/includes/endpoints/ListUsersEndpoint.php
@@ -108,6 +108,23 @@ class ListUsersEndpoint extends KeyAuthEndpoint {
                 'integrations' => $integrations
             ];
 
+            if (isset($_GET['groups'])) {
+                $groups = $api->getDb()->query(
+                    <<<SQL
+                    SELECT g.id, g.name
+                    FROM nl2_users_groups ug
+                        RIGHT JOIN nl2_groups g
+                            ON g.id = ug.group_id
+                    WHERE ug.user_id = ?
+                    SQL,
+                    [$user->id]
+                );
+
+                if ($groups->count()) {
+                    $user_json['groups'] = $groups->results();
+                }
+            }
+
             $users_json[] = $user_json;
         }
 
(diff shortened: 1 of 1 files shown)


Derkades:
Is it possible to return the same info in the same format as the user info endpoint, or would that cause too many queries?

image

Personally I don't care about `discord_role_id` and `ingame_rank_name`. In my opinion they can even be removed from the user info endpoint, as they are not currently used by the plugin (and won't ever be). So just id, name, staff, order in object format.

partydragen:
discord_role_id is not removed? i still believe it returns in correct conditions (YOU EDITED THE MESSAGE)

and still being alot of queries thought, also then UserInfo endpoint pretty being pointless as then everything can be done from UserList endpoint itself for searching

Derkades:
Yeah I edited the message when I looked at the source code and realised I was wrong :P

Returning `staff` and `order` should be possible with the same query, I believe they are columns in `nl2_groups`?

samerton:
> Yeah I edited the message when I looked at the source code and realised I was wrong :P
> 
> Returning `staff` and `order` should be possible with the same query, I believe they are columns in `nl2_groups`?

Added these now - it's not much effort to add `discord_role_id` and `ingame_rank_name` too

Derkades:
Thanks. A final question, does a user always have at least one group? I.e, should the plugin expect `groups` to not be present at all times?

tadhgboyle:
> should the plugin expect `groups` to not be present at all times?

Do you mean **to be present**? 🤔

samerton:
> Thanks. A final question, does a user always have at least one group? I.e, should the plugin expect `groups` to not be present at all times?

I wouldn't necessarily rely on there always being at least one group - but it will always return at least an empty array.

partydragen:
> > Thanks. A final question, does a user always have at least one group? I.e, should the plugin expect `groups` to not be present at all times?
> 
> I wouldn't necessarily rely on there always being at least one group - but it will always return at least an empty array.

Atleast using user class will always return group, as if a loaded user missing groups it will give the default group

Derkades (approved):

PR #3194, opened 2023-01-15
Remove group sync info from user info endpoint

Derkades:
This info is not currently used by the plugin or discord bot. Nor is there any reason they would ever use it in the future. Group sync logic is done by the website, not API clients, so API clients need not be aware of group sync rules.

Especially useful to reduce queries for #3193 while still sharing the same format

Curious to hear what you think. In my opinion, if we do need this data in the future, we should add a dedicated group sync info endpoint to request it.

### DIFF
--- a/core/classes/Core/Util.php
+++ b/core/classes/Core/Util.php
@@ -380,23 +380,6 @@ class Util {
     }
 
     /**
-     * Get in-game rank name from a website group ID, uses Group Sync rules.
-     *
-     * @param int $website_group_id ID of website group to search for.
-     * @return string|null Name of in-game rank or null if rule is not set up.
-     */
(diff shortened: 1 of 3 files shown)


tadhgboyle (dismissed):
Makes sense to me, we might also be able to remove the getIngameRankName method from Util

Derkades:
Util::getIngameRankName() is now removed

tadhgboyle:
Reminder to self we need to update docs here when merged https://docs.namelessmc.com/en/api-documentation

tadhgboyle (approved):

Derkades:
> Reminder to self we need to update docs here when merged https://docs.namelessmc.com/en/api-documentation

done

PR #3196, opened 2023-01-16
Remove panel template "Find Templates" section

tadhgboyle:
Since we no longer provide backwards compatibility to panel templates, and very few (if any) people use them, we are removing this section. There are none posted on the namelessmc marketplace at the moment anyways.


Derkades:
Backporting this to 2.0.x will break panel templates, but I assume that is intentional given the content of this PR :-)

tadhgboyle:
> Backporting this to 2.0.x will break panel templates, but I assume that is intentional given the content of this PR :-)

Exactly!

samerton:
I think if that's the idea, we should add a migration to reset the panel template back to default, otherwise people won't be able to access the panel templates tab to disable their custom template.

partydragen:
>

partydragen:
> I think if that's the idea, we should add a migration to reset the panel template back to default, otherwise people won't be able to access the panel templates tab to disable their custom template.

Panel templates page and functions will still be there, this is only about removing "Find Templates" section

samerton:
Will the removal of the variables in the page PHP file not break any template files referring to them?

partydragen:
> Will the removal of the variables in the page PHP file not break any template files referring to them?

No, All of this is just for the resource listing, the list of template suggestions from NamelessMC website

samerton:
This does seem to be a breaking change for users using custom panel templates - tested and the attached screenshot is the result (white screen).

We'll need the migration or to push this back to 2.1.0

Screenshot 2023-01-17 at 16 56 03

tadhgboyle:
Ah in that case I think pushing to 2.1.0 is fine! Good catch

partydragen (approved):

PR #3198, opened 2023-01-17
AuthMe improvements and fixes

tadhgboyle:
This PR brings some changes to the AuthMe system:
- Add check authme is setup properly before allowing users to enter username/password
- Add support for making users enter profile fields when they enter their email on the second step of authme registration
- Use `DB` class instead of `mysqli` for connecting to authme db
- Allow users to configure if they get their authme password synced upon login or not


custom/templates/DefaultRevamp/authme_email.tpl
```diff
@@ -31,17 +31,64 @@
                     </div>
                 </div>
                 <form class="ui form" action="" method="post" id="form-authme-email">
-                    <div class="field">
-                        <label for="inputEmail">{$EMAIL}</label>
-                        <input type="email" id="inputEmail" name="email" placeholder="{$EMAIL}">
-                    </div>
+                    {assign var=counter value=1}
```
tadhgboyle, line 34: We should look into reusable tpl files 🤔 😅

partydragen (commented):
I believe this caused duplication issue, As module column NULL being always as unique same, it dont care if key already exist when module is null

custom/templates/DefaultRevamp/authme_email.tpl
partydragen: make the fields stuff into a include tpl? Yea i thinked about doing that when i made it, same as few of my modules will also use this

partydragen:
> I believe this caused duplication issue, As module column NULL being always as unique same, it dont care if key already exist when module is null

modules/Core/pages/authme_connector.php
```diff
@@ -28,139 +28,153 @@
...
-                    ],
-                    'email' => [
-                        Validate::REQUIRED => true,
-                        Validate::MIN => 4,
-                        Validate::MAX => 64,
-                        Validate::UNIQUE => 'users'
-                    ]
-                ]);
-            } else {
-                $validation = Validate::check($_POST, [
```
partydragen, line 46: Hm we still need username validation check, I see it never was here but we still need to make sure username is unique, same rules as nickname

modules/Core/pages/authme_connector.php
```diff
@@ -28,139 +28,153 @@
...
-                ]);
+                        Validate::MAX => (is_null($field->length) ? 1024 : $field->length)
+                    ];
+                }
             }
 
+            $validation = Validate::check($_POST, $to_validation);
+
             $validation->messages([
-                'username' => [
```
partydragen, line 57: Then you need to keep this aswell

modules/Core/pages/authme_connector.php
tadhgboyle: Great catch

custom/panel_templates/Default/integrations/minecraft/minecraft_authme.tpl
```diff
@@ -65,54 +65,49 @@
...
-                                    <label for="inputAuthmeSync" class="custom-control-label">
-                                        {$AUTHME_PASSWORD_SYNC} <span class="badge badge-info"><i class="fa fa-question-circle" data-container="body" data-toggle="popover" data-placement="top" title="{$INFO}" data-content="{$AUTHME_PASSWORD_SYNC_HELP}"></i></span>
-                                    </label>
+                                    <input type="text" class="form-control" name="db_table" value="{if isset($AUTHME_DB_DETAILS['table'])}{$AUTHME_DB_DETAILS['table']|escape}{else}authme{/if}">
                                 </div>
                                 <div class="form-group">
                                     <input type="hidden" name="token" value="{$TOKEN}">
                                     <input type="submit" value="{$SUBMIT}" class="btn btn-primary">
+                                    <button class="btn btn-info" id="testConnection">Test connection</button>
```
partydragen: Missing translation

modules/Core/pages/authme_connector.php
```diff
@@ -26,141 +27,179 @@
...
+                        'nickname' => $nickname,
+                        'password' => $_SESSION['authme']['pass'],
+                        'pass_method' => $_SESSION['authme']['hash'],
+                        'joined' => date('U'),
+                        'email' => $email,
+                        'reset_code' => $code,
+                        'lastip' => $ip,
+                        'last_online' => date('U'),
+                        'language_id' => $language_id,
+                        'authme_registered' => true,
```
partydragen: switch to register_method instead or something then we could also use it for oauth_{service} etc

modules/Core/pages/authme_connector.php
```diff
@@ -26,141 +27,179 @@
...
+                        'authme_sync_password' => Input::get('authme_sync_password') === 'on',
+                    ]);
+
```
partydragen, line 153: Missing to fire new user event

partydragen (changes requested):

partydragen (approved):

PR #3199, opened 2023-01-18
Allow selecting timezone during install

tadhgboyle:
Will automatically determine timezone from browser, but fall back to `Europe/London` if nothing was found.
CLI installer also updated and tested

### DIFF
--- a/core/classes/Database/DatabaseInitialiser.php
+++ b/core/classes/Database/DatabaseInitialiser.php
@@ -192,7 +192,7 @@ class DatabaseInitialiser {
         Util::setSetting('external_query', '0');
         Util::setSetting('followers', '0');
         Util::setSetting('language', '1');
-        Util::setSetting('timezone', 'Europe/London');
+        Util::setSetting('timezone', $_SESSION['install_timezone']);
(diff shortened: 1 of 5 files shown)


partydragen:
Is it not better with server timezone

tadhgboyle:
> Is it not better with server timezone

Unsure... if they are using shared host it is extremely unlikely the timezone is correct. 
Otherwise, they may have not configured the timezone on the server and they will likely want it to match their local time

Derkades:
I agree, guest users are more likely to be in the same country as the server owner for some servers. Not for others, but in that case the server owner's timezone is not any worse than hardcoded Europe/London.

Ideally guests would get their own local timezone instead of the site-wide timezone, but I don't think there's a good way to get the user's timezone from PHP. You'd need to set a cookie from javascript, and read that cookie value in PHP (not present in the first request)
https://github.com/NamelessMC/Nameless/blob/e7f2e2ad38559f15ae7b154cfbf39a6aba00d8e2/core/init.php#L175-L177

partydragen (approved):

PR #3200, opened 2023-01-18
Allow running release script without TTY

Derkades:
Turns out jenkins couldn't run the release script because I added `-it` out of habit


tadhgboyle (approved):
Masterful work as always my good sir

PR #3202, opened 2023-01-18
Check for pdo_mysql extension, mysqli is no longer used

Derkades:
With #3198, the last usages of mysqli will be removed. Now that `mysqli` is no longer required by NamelessMC, check for presence of the `pdo_mysql` extension instead.

### DIFF
--- a/core/installation/steps/requirements_validation.php
+++ b/core/installation/steps/requirements_validation.php
@@ -19,8 +19,8 @@ unset($_SESSION['requirements_validated']);
             <div class="sixteen wide mobile eight wide tablet seven wide computer column">
                 <?php
                 validate_requirement('PHP 7.4+', PHP_VERSION_ID >= 70400);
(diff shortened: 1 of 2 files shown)


tadhgboyle (commented):
Is `pdo_mysql` going to be as common as `mysqli`?
As in, does this make it even harder to install for people on bad hosts?

Derkades:
`pdo_mysql` is an adapter for PDO, `mysqli` is something else entirely. NamelessMC uses `pdo_mysql`, but never checked if it was installed. This is not really an issue, virtually every host that has `PDO` installed also has the `pdo_mysql` extension installed.

`mysqli` is no longer used by NamelessMC so the check for it can be removed.

TL;DR: This does not worsen compatibility, instead it improves compatibility by not requiring extensions that are not used

tadhgboyle (approved):

PR #3219, opened 2023-01-21
Suggest using Nameles-Plugin as an alternative to mcassoc

Derkades:
image


Derkades:
I heard mcassoc is going to be :boom:

PR #3220, opened 2023-01-21
Remove MCAssoc 🔥 

tadhgboyle:
Microsoft removed the ability to apply a skin automatically via a URL, so MCAssoc has become much more annoying to use for the end user.
Additionally our implementation of it has been broken for a few years, but only a handful of people have ever mentioned it.
The Nameless Plugin is a better option for handling account verification!


core/classes/Core/Util.php
```diff
@@ -377,6 +377,8 @@ public static function setSetting(string $setting, ?string $new_value, string $m
         } else if ($new_value !== null) {
             $cache[$setting] = $new_value;
         }
+
+        self::setSettingsCache($module, $cache);
```
Aberdeener, line 381: Important line here, not worth it's own PR

modules/Core/pages/register.php
```diff
@@ -245,152 +210,123 @@
...
+                    }
 
-                        // Hash password first
-                        $password = password_hash($_POST['password'], PASSWORD_BCRYPT, ['cost' => 13]);
-                        $_SESSION['password'] = $password;
-                        unset($_POST['password']);
+                    $password = password_hash(Input::get('password'), PASSWORD_BCRYPT, ['cost' => 13]);
+                    // Get current unix time
+                    $date = new DateTime();
+                    $date = $date->getTimestamp();
```
supercrafter100:
```suggestion
                    $date = date('U');
```

partydragen (dismissed):

supercrafter100 (changes requested):

modules/Core/pages/register.php
supercrafter100: I guess you could technically just move this definition straight to the `$user->create()` part

tadhgboyle: Done

supercrafter100 (dismissed):

partydragen (dismissed):

supercrafter100 (approved):

partydragen (approved):

PR #3221, opened 2023-01-21
Fix `Util::setSetting` cache not being updated

tadhgboyle:


Derkades:
Are you sure this is an issue? The reference hasn't changed, so unless PHP is different than any other programming language, the line you added does nothing

for example:
```php
function something($example_array) {
    $example_array[] = 'Content';
}

$example_array = [];
something($example_array);
echo $example_array[0]; // should print "Content", while the $example_array variable was never re-assigned
```

your are proposing something like this:
```php
function something($example_array) {
    $example_array[] = 'Content';
    return $example_array;
}

$example_array = [];
$example_array = something($example_array);
echo $example_array[0];
```

tadhgboyle:
Yes it makes a difference, not sure the technicalities but since $cache is set to the value returned by getSettingCache, it is not updated just by updating the value, we have to reassign using the setSettingsCache

This fixes the issue I mentioned the other day about needing to reload page for cache value to fully update. I can do a demo video later

tadhgboyle:
For your first example to work, you'd need to pass by reference, using the & symbol before the variable name - PHP is different than other languages haha

Derkades:
The documentation on this is scarce, but I could find [this](https://stackoverflow.com/a/9740541) which does say PHP makes a full array copy when it is modified inside a function... That's insane.

Derkades:
Perhaps a better solution is modifying getSettingsCache to return a reference (just prefix the function name with `&`)? Avoids the overhead of copying the entire settings array whenever a setting is modified

tadhgboyle:
> Perhaps a better solution is modifying getSettingsCache to return a reference (just prefix the function name with `&`)? Avoids the overhead of copying the entire settings array whenever a setting is modified

I don't think we can store results of functions as references:
Screenshot 2023-01-21 at 16 09 58
Plus even if we could, I'm not sure PHP would know "where" to save the modified variable.

Derkades:
All good now

Derkades:
I am going to open a separate PR to update all remaining manual database queries to use Util::getSetting

Derkades (approved):

PR #3222, opened 2023-01-22
Rewrite Discord group sync to use v5 Nameless-Link API

tadhgboyle:


modules/Core/includes/endpoints/InfoEndpoint.php
```diff
@@ -24,7 +24,7 @@ public function execute(Nameless2API $api): void {
 
         $ret = [];
 
-        $ret['nameless_version'] = Util::getSetting('nameless_version');
+        $ret['nameless_version'] = '2.1.0';
```
tadhgboyle: Undo before merge

Derkades (dismissed):
You've requested my review but I don't know enough about NamelessMC to review this code.

I will say I am happy with the current API, from the perspective of Nameless-Plugin and Nameless-Link. And you tested it and it worked, so that's a good sign (although all previous group sync system iterations also appeared to work)

You will need a second approval anyway, after undoing `2.1.0` line :P

supercrafter100 (dismissed):

Derkades (approved):

PR #3224, opened 2023-01-22
Replace manual queries with util setting functions

Derkades:
Labeled 'bug' because using manual queries means Util::setSetting will return old data (cache is only updated when using setSetting)

For 2.1.0, not 2.0.3 because large changes like this tend to accidentally introduce bugs as well


Derkades:
please do not merge yet; I would like to check again after the mcassoc and authme pull requests are merged

Derkades:
Now ready for merge

modules/Core/pages/panel/api.php
```diff
@@ -30,17 +30,8 @@
...
-
                 // Update key
-                DB::getInstance()->update(
-                    'settings',
-                    $plugin_api,
-                    [
-                        'value' => $new_api_key
-                    ]
-                );
+                Util::setSetting('mc_api_key', $new_api_key);
```
tadhgboyle, line 34: This diff is awesome to see

modules/Core/pages/panel/privacy_and_terms.php
```diff
@@ -100,15 +100,17 @@
...
 }
-$site_terms = $site_terms[0]->value;
 
 $site_privacy = DB::getInstance()->get('privacy_terms', ['name', 'privacy'])->results();
 if (!count($site_privacy)) {
-    $site_privacy = DB::getInstance()->get('settings', ['name', 'privacy_policy'])->results();
+    $site_privacy = Util::getSetting('privacy_policy');
+} else {
+    $site_privacy = $site_privacy[0]->value;
 }
```
tadhgboyle, line 113: I feel like these are likely for backwards compatibility which would have been initially made for years ago... probably safe to always assume `privacy_policy` and `t_and_c_site` (or whichever of the two we are currently setting during install for each of them)?

modules/Core/pages/register.php
```diff
@@ -23,18 +23,13 @@
...
     // Registration is disabled, display a message
     // Get registration disabled message and assign to Smarty variable
-    $registration_disabled_message = DB::getInstance()->get('settings', ['name', 'registration_disabled_message'])->results();
-    if (count($registration_disabled_message)) {
-        $message = Output::getPurified($registration_disabled_message[0]->value);
-    } else {
-        $message = 'Registration is currently disabled.';
-    }
+    $fallback_message = 'Registration is currently disabled.';
+    $message = Output::getPurified(Util::getSetting('registration_disabled_message', $fallback_message));
```
tadhgboyle:
```suggestion
    $message = Output::getPurified(Util::getSetting('registration_disabled_message', 'Registration is currently disabled.'));
```

modules/Core/pages/terms.php
```diff
@@ -17,12 +17,13 @@
...
+    $site_terms = Util::getSetting('t_and_c_site');
+} else {
+    $site_terms = $site_terms[0]->value;
 }
-$site_terms = Output::getPurified($site_terms[0]->value);
+$site_terms = Output::getPurified($site_terms);
 
-$nameless_terms = DB::getInstance()->get('settings', ['name', 't_and_c'])->results();
-$nameless_terms = Output::getPurified($nameless_terms[0]->value);
+$nameless_terms = Util::getSetting('t_and_c');
```
tadhgboyle: Same note here about the multiple settings names

tadhgboyle (dismissed):
Looks great, really nice to see these areas simplified

modules/Core/pages/register.php
Derkades: Decided to make it translatable instead :D

modules/Core/pages/minecraft/banner.php
```diff
@@ -49,16 +49,7 @@
...
-            if (count($query_type)) {
-                if ($query_type[0]->value == '1') {
-                    $query_type = 'external';
-                } else {
-                    $query_type = 'internal';
-                }
-            } else {
-                $query_type = 'internal';
-            }
+            $query_type = Util::getSetting('external_query') ? 'external' : 'internal';
```
samerton: Was this changed in a recent PR so that we now have `query_type` instead of `external_query` or am I imagining things?

modules/Core/pages/panel/minecraft_servers.php
```diff
@@ -393,8 +392,7 @@
             $available_parent_servers = DB::getInstance()->get('mc_servers', ['parent_server', 0])->results();
 
             // Display query information alert only if external query is selected
-            $external_query = DB::getInstance()->get('settings', ['name', 'external_query'])->results();
-            $external_query = $external_query[0]->value;
+            $external_query = Util::getSetting('external_query');
```
samerton: Same question RE query type

modules/Core/pages/panel/privacy_and_terms.php
samerton: Yes let's get rid of this read from the settings table since we now use the separate `privacy_terms` table

samerton (changes requested):

modules/Core/pages/minecraft/banner.php
Derkades: You're right: https://github.com/NamelessMC/Nameless/pull/3244/files#diff-ddc3a26d3f3aee9e751be4668fd9c35c5e593001adc6ebc7d5c1259713cdc116

I wonder why it didn't produce conflicts. Or perhaps I made a mistake when merging

Derkades: > I wonder why it didn't produce conflicts

Mystery solved, this branch was just very outdated

modules/Core/pages/terms.php
samerton: Could maybe do with keeping `Output::getPurified` here

Derkades: I don't know if I had a good reason for removing it or if I did it by accident.. in any case I'll restore getPurified

samerton (approved):

PR #3225, opened 2023-01-22
Remove MINECRAFT constant

Derkades:
Now that getting the setting value no longer requires a query, the value of the mc_integration setting does not need to be saved.

Having one place for settings prevents issues like this https://github.com/NamelessMC/Nameless/pull/3221/commits/6792a808ffa48ffd24941932ac40c48cf24b1883


tadhgboyle:
This is a good change, however I think the downside is now the possibility of making typos while doing `getSetting('mc_integration');`
I wonder what it would be like to create some sort of `Settings` class which houses the `get`/`set` methods as well as constants for common settings, in a different PR of course?

It would also be great to remove some other settings stored as constants, such as `BOT_URL`, in another PR though

Derkades:
Good idea, having a class with setting constants would be nice to reduce bugs caused by typos

tadhgboyle (dismissed):

tadhgboyle (dismissed):

Derkades:
it had conflicts again :(

tadhgboyle (approved):

PR #3226, opened 2023-01-23
Event system DX improvements

### LINKED ISSUE #3212: Event system DX improvements
*[Inspiration](https://laravel.com/docs/9.x/events)*

# Problems
Currently our `EventHandler` system works well, but has one main caveat; it is very hard to know what params to pass for each event.
For example, when executing an event:
```php
EventHandler::executeEvent('userBanned', [
	'banned_user_id' => $user->id,
	'banned_by_id' => $staff_user->id,
	// Anything else??
]);
```
When listening to an event:
```php
class UserBannedListener {
	public static function execute(array $params = []) {
		$banned_user = $params['banned_user']; // Typo
		$banned_by = $params['banned_by_id'];
		$reason = $params['reason']; // Doesn't exist!
	}
}
```

To make matters worse, most event hooks will silently ignore missing params:
https://github.com/NamelessMC/Nameless/blob/4cc529be04be6816a00e49546711061756a89bdd/core/classes/Events/HookBase.php#L20-L31
https://github.com/NamelessMC/Nameless/blob/4cc529be04be6816a00e49546711061756a89bdd/modules/Core/hooks/DeleteUserHook.php#L13

# Solutions
One way of solving these problems is to introduce classes for each type of event.
Define the event class:
```php
class UserBannedEvent extends AbstractEvent {
	public User $banned;
	public User $banned_by;
	public string $reason;	

	public function __construct(User $banned, User $banned_by, string $reason) {
		$this->banned = $banned;
		$this->banned_by = $banned;
		$this->reason = $reason; // oh, so we did forget a param in the first example! :p
	}
}
```

Register the event
```php
EventHandler::registerEvent(UserBannedEvent::class);
```
The event class will contain `description()`, `return()` and `internal()` methods, instead of needing to pass them here.

Register some listeners for the `UserBannedEvent`
```php
EventHandler::registerListener(UserBannedEvent::class, UserBannedListener::class);
EventHandler::registerListener(UserBannedEvent::class, UserBannedForumListener::class);
```

Execute the `UserBannedEvent`
```php
EventHandler::executeEvent(new UserBannedEvent(
	$user,
	$staff_user,
	$reason,
));
```

Handle the `UserBannedEvent` within a listener
```php
class UserBannedListener extends AbstractListener {
	public function handle(UserBannedEvent $event) {
		// Do stuff with all of the params, and get type safety!
	}
}

class UserBannedForumListener extends AbstractListener {
	public function handle(UserBannedEvent $event) {
		// Do stuff with all of the params, and get type safety!
	}
}
```

# Backwards compatibility
Since these changes would be quite drastic for event heavy modules, we should provide backwards compatibility for at least 1 major release.
Thanks to the power of PHP reflection:tm:, we can implement a fairly simple backwards compatibility strategy (assuming we keep the behind the scenes the same system until we stop supporting the old way of doing things):

## Registering events
When registering an event, there are two things that can happen for the first parameter (event name (or class name)):
1. They pass some string; `’user_banned’`
	- No conversion has to be done here, we will assume this is what event they want to listen to.
2. They pass a class name; `UserBannedEvent::class`
	- If they pass a class name, no other params will be required, as they will be accessible via the class (`description()`, `internal()`, `return()` methods)
	- We have two options on how to convert it to the equivalent string:
		1. Allow `AbstractEvent`s to define a `name()` method
			```php
		  	class UserWasBannedForSomeReasonEvent extends AbstractEvent {
				public function name() {
					return 'user_banned';
				}
			
				// ...
			}
			```
		2. If `name()` method was not defined, we can attempt to convert the class name - something like this:
			```php
			$class_name = UserBannedEvent::class;
			$parts = preg_split('/(?=[A-Z])/', $class_name); // ['User', 'Banned', 'Event']
			if (end($parts) === 'Event') {
				array_pop($parts); // ['User', 'Banned']
			}

			$string = lcfirst(implode('', $parts)); // 'userBanned'
			```
*These same conversion rules will apply to executing events, so `EventHandler::registerListener($event, ...)` can take in a string or class name as well.*
We can store the event with the string or converted class name so it can be executed using either.

## Registering listeners
Minimal changes will need to occur here.
We will convert the first parameter (string or class name) to determine which event to execute, but the second param will remain the same.

## Executing events
When executing an event, there are two possibilities of how they call `EventHandler::executeEvent`:
1. They pass 2 parameters: `$event` (name of event to execute) + `$params` (array of params to pass to listeners)
	- No changes needed
2. They pass 1 parameter: `$event` (an instance of `AbstractEvent`)
	- If this is the case, we can pull out the `$params` from the `$event` instance, doing something like this:
	```php
	abstract class AbstractEvent {
		public function params(): array {
			return get_object_vars($this); // ['banned' => User, 'banned_by' => User, 'reason' => string]
		}
	}

	class UserBannedEvent extends AbstractEvent {
		public User $banned;
		public User $banned_by;
		public string $reason;	
	
		public function __construct(User $banned, User $banned_by, string $reason) {
			$this->banned = $banned;
			$this->banned_by = $banned;
			$this->reason = $reason;
		}
	}
	```
Actually passing the `$params` array to the listeners may vary, depending on how their `handle` method is defined. See the next section on how we can do it.

## Handling events within listeners
If a listener is defined like so:
```php
class UserBannedListener extends AbstractListener {
	public function handle(UserBannedEvent $event) {
		// Do cool stuff
	}
}
```
When iterating the listeners and we come across a listener which has a param of type `AbstractEvent`, then we can pass an `$event` object (created from `$params` if not passed as an object to `executeEvent`)

Otherwise if a listener is wanting an `array` for its param, we can pass the `$params` array we made in the step above directly.
```php
class UserBannedListener {
	public function handle(array $params) {
		// Do cool stuff, but not as cool as using the new event syntax... :(
	}
}
```

# Bonuses
- Allowing events to be represented as classes could let us do stuff like this:
	```php
	abstract class AbstractEvent {
		public static function dispatch() {
			EventHander::executeEvent(new static(...func_get_args()));
		}
	}

	UserBannedEvent::dispatch($banned, $banned_by, $reason);
	```
- Once we remove the backwards compatibility layer, I think we can remove the idea of "registering events"
- We can create our own [PHP Debugbar collector](http://phpdebugbar.com/docs/data-collectors.html#creating-collectors) to display events executed during page load

# Conclusion
Overall, I am looking to add support for strongly typed and easier to work with events.
By keeping backwards compatibility using the strategy I explained above, I think there are very few (if any) downsides to this nicer system.
If you made it this far, thank you! Please let me know what you think, or if you have suggestions! ❤️

tadhgboyle:
Implementation of https://github.com/NamelessMC/Nameless/issues/3212


core/classes/Events/EventHandler.php
```diff
@@ -5,14 +5,21 @@
...
- * @version 2.0.0-pr13
+ * @version 2.1.0
  * @license MIT
  */
 class EventHandler {
 
     private static array $_events = [];
     private static array $_webhooks = [];
 
+    public static function discoverEvents(): void {
```
partydragen: Believe we should keep that modules self register the event to prevent outdated files, and if event require certain modules etc so better to keep so modules self register it

modules/Core/classes/Events/UserIntegrationLinkedEvent.php
```diff
@@ -0,0 +1,38 @@
+<?php
+
+class UserIntegrationLinkedEvent extends AbstractEvent implements DiscordDispatchable {
```
partydragen, line 3: Well most event class names sounds like the event is complete, adding the ability to setCancelled later will make the event file names sounds weird So i guess the event names should had stayed as they were

Example this is linkIntegrationUser with can still be cancellable but the name UserIntegrationLinkedEvent does not sound like that

partydragen (changes requested):

core/classes/Events/EventHandler.php
partydragen: Also with the autoDiscover er don't know if the event is internal or not anymore

modules/Forum/hooks/ForumDiscordWebhookListener.php
```diff
@@ -0,0 +1,14 @@
+<?php
```
Aberdeener, line 1: Delete before merge, for testing

modules/Forum/module.php
```diff
@@ -55,21 +55,11 @@ public function __construct(Language $language, Language $forum_language, Pages
...
-                'content_full' => $this->_language->get('general', 'full_content'),
-                'avatar_url' => $this->_language->get('user', 'avatar'),
-                'title' => $this->_forum_language->get('forum', 'topic_title'),
-                'url' => $this->_language->get('general', 'url'),
-                'available_hooks' => $this->_forum_language->get('forum', 'available_hooks')
-            ]
-        );
+        EventHandler::registerListener(UserDeletedEvent::class, [DeleteUserForumHook::class, 'execute']);
+        EventHandler::registerListener(GroupClonedEvent::class, [CloneGroupForumHook::class, 'execute']);
+        EventHandler::registerListener(DiscordWebhookFormatterEvent::class, [ForumDiscordWebhookListener::class, 'execute']);
```
Aberdeener, line 60: Figure out a way for `ForumDiscordWebhookListener` to check the underlying event name that called the `DiscordWebhookFormatterEvent`

partydragen (commented):
looks good so far

core/classes/Events/EventHandler.php
```diff
@@ -139,18 +202,24 @@ public static function executeEvent(string $event, array $params = []) {
...
-    public static function getEvents(bool $internal = false): array {
+    public static function getEvents(Language $language, bool $internal = false): array {
         $return = [];
 
         foreach (self::$_events as $name => $meta) {
             if (!$meta['internal'] || $internal) {
-                $return[$name] = $meta['description'];
+                // The description might be an array if the event is a class based event
+                $description = is_array($meta['description'])
+                    ? $language->get(...$meta['description'])
```
partydragen: Event description need to be reworked, this does not support modules module languages

Example the forum events:
forum/new_topic_hook_info
forum/topic_reply

partydragen (changes requested):

partydragen (commented):
Forum posting is broken 

`Uncaught ReflectionException
Function MentionsHook::preCreate() does not exist
/var/www/dev/core/classes/Events/EventHandler.php`

Skjermbilde

partydragen (changes requested):
Registration broken [2023-02-17, 14:37:16] /var/www/dev/core/classes/Events/DiscordWebhookBuilder.php(64): Embed closure must return a DiscordEmbed instance

partydragen:
> Registration broken [2023-02-17, 14:37:16] /var/www/dev/core/classes/Events/DiscordWebhookBuilder.php(64): Embed closure must return a DiscordEmbed instance

same with userIntegrationLinked

modules/Core/classes/Events/UserProfilePostCreatedEvent.php
```diff
@@ -0,0 +1,39 @@
...
+            ->setUsername($this->poster->data()->username . ' | ' . SITE_NAME)
+            ->setAvatarUrl($this->poster->getAvatar(128, true))
+            ->addEmbed(function (DiscordEmbed $embed) use ($language) {
+                return $embed
+                    ->setDescription($language->get('user', 'x_posted_on_y_profile', [
+                        'poster' => $this->poster->getDisplayname(),
+                        'user' => $this->profile_user->getDisplayname(),
+                    ]))
+                    ->setUrl(URL::getSelfURL() . ltrim(URL::build('/profile/' . urlencode($this->profile_user->getDisplayname(true)) . '/#post-' . urlencode(DB::getInstance()->lastId())), '/'))
+                    ->addField($language->get('user', 'content'), strip_tags(str_ireplace(['<br />', '<br>', '<br/>'], "\r\n", $this->content)));
```
partydragen: user/content term is invalid, its general/content

modules/Core/classes/Events/UserProfilePostReplyCreatedEvent.php
```diff
@@ -0,0 +1,39 @@
...
+            ->setUsername($this->poster->data()->username . ' | ' . SITE_NAME)
+            ->setAvatarUrl($this->poster->getAvatar(128, true))
+            ->addEmbed(function (DiscordEmbed $embed) use ($language) {
+                return $embed
+                    ->setDescription($language->get('user', 'x_replied_on_y_profile', [
+                        'replier' => $this->poster->getDisplayname(),
+                        'user' => $this->profile_user->getDisplayname(),
+                    ]))
+                    ->setUrl(URL::getSelfURL() . ltrim(URL::build('/profile/' . urlencode($this->profile_user->getDisplayname(true)) . '/#post-' . urlencode(DB::getInstance()->lastId())), '/'))
+                    ->addField($language->get('user', 'content'), strip_tags(str_ireplace(['<br />', '<br>', '<br/>'], "\r\n", $this->content)));
```
partydragen: user/content term is invalid, its general/content

partydragen (changes requested):

modules/Core/classes/Events/AnnouncementCreatedEvent.php
```diff
@@ -0,0 +1,38 @@
+<?php
+
+class AnnouncementCreatedEvent extends AbstractEvent implements DiscordDispatchable {
+
+    public User $creator;
+    public string $header;
+    public string $message;
+
+    public function __construct(User $creator, string $header, string $message) {
+        $this->creator = $creator;
```
partydragen, line 10: Uncaught TypeError
AnnouncementCreatedEvent::__construct(): Argument #1 ($creator) must be of type User, string given, called in /var/www/dev/core/classes/Misc/Announcements.php on line 187

partydragen (changes requested):

partydragen:
Another legacy support error?
Uncaught RuntimeException
Could not find class to hydrate for event 'paymentCompleted'
/var/www/dev/core/classes/Events/EventHandler.php
https://dev2.partydragen.com/panel/store/payments/?action=create

tadhgboyle:
> Another legacy support error?
> 
> Uncaught RuntimeException
> 
> Could not find class to hydrate for event 'paymentCompleted'
> 
> /var/www/dev/core/classes/Events/EventHandler.php
> 
> https://dev2.partydragen.com/panel/store/payments/?action=create

What does the registerEvent and listener for this event look like?

partydragen:
> > Another legacy support error?
> > Uncaught RuntimeException
> > Could not find class to hydrate for event 'paymentCompleted'
> > /var/www/dev/core/classes/Events/EventHandler.php
> > https://dev2.partydragen.com/panel/store/payments/?action=create
> 
> What does the registerEvent and listener for this event look like?

https://github.com/partydragen/Nameless-Store/blob/a1ad2fcee01d62c748eb282d7c3f9f0cf3aa9f4a/upload/modules/Store/classes/Payment.php#L139

`        EventHandler::registerEvent('paymentPending',  $store_language->get('admin', 'payment_pending'));
        EventHandler::registerEvent('paymentCompleted', $store_language->get('admin', 'payment_completed'));
        EventHandler::registerEvent('paymentRefunded', $store_language->get('admin', 'payment_refunded'));
        EventHandler::registerEvent('paymentReversed', $store_language->get('admin', 'payment_reversed'));
        EventHandler::registerEvent('paymentDenied', $store_language->get('admin', 'payment_denied'));
        EventHandler::registerEvent('storeCheckoutAddProduct', 'storeCheckoutAddProduct', [], true, true);
        EventHandler::registerEvent('renderStoreProduct', 'renderStoreProduct', [], true, true);
        EventHandler::registerEvent('storeCheckoutAddProduct', 'storeCheckoutAddProduct', [], true, true);
        EventHandler::registerEvent('storeCheckoutFieldsValidation', 'storeCheckoutFieldsValidation', [], true, true);

        require_once(ROOT_PATH . '/modules/Store/hooks/CheckoutAddProductHook.php');
        EventHandler::registerListener('storeCheckoutAddProduct', 'CheckoutAddProductHook::globalLimit');
        EventHandler::registerListener('storeCheckoutAddProduct', 'CheckoutAddProductHook::userLimit');
        EventHandler::registerListener('storeCheckoutAddProduct', 'CheckoutAddProductHook::requiredProducts');
        EventHandler::registerListener('storeCheckoutAddProduct', 'CheckoutAddProductHook::requiredGroups');
        EventHandler::registerListener('storeCheckoutAddProduct', 'CheckoutAddProductHook::requiredIntegrations');
        EventHandler::registerListener('renderStoreCategory', 'ContentHook::purify');
        EventHandler::registerListener('renderStoreCategory', 'ContentHook::codeTransform', 15);
        EventHandler::registerListener('renderStoreCategory', 'ContentHook::decode', 20);
        EventHandler::registerListener('renderStoreCategory', 'ContentHook::renderEmojis', 10);
        EventHandler::registerListener('renderStoreCategory', 'ContentHook::replaceAnchors', 15);
        EventHandler::registerListener('renderStoreProduct', 'ContentHook::purify');
        EventHandler::registerListener('renderStoreProduct', 'ContentHook::codeTransform', 15);
        EventHandler::registerListener('renderStoreProduct', 'ContentHook::decode', 20);
        EventHandler::registerListener('renderStoreProduct', 'ContentHook::renderEmojis', 10);
        EventHandler::registerListener('renderStoreProduct', 'ContentHook::replaceAnchors', 15);`

partydragen (approved):

PR #3227, opened 2023-01-23
Properly encode content for javascript

Derkades:
`json_encode` is a convenient way to escape backslashes, quotes, and other characters for a javascript string.

Fixes javascript parsing error when route contains `\` or `"`, for example.

image


tadhgboyle (approved):
Seems to work good for me locally

PR #3231, opened 2023-01-26
Fix `register_method` column not being filled when accounts are created

tadhgboyle:
Column added here https://github.com/NamelessMC/Nameless/pull/3198, but not updated around the codebase 🤦🏼‍♂️


partydragen (changes requested):
API endpoint is missing

supercrafter100 (approved):

partydragen (approved):

PR #3238, opened 2023-01-28
Use correct source directory

Derkades:
This fixes the issue where the upgrade zip was identical to the release zip


tadhgboyle (approved):

Derkades:
Wow, this backported without manual conflict resolution, even though the script is in an entirely different location in the `release/2.0.3` branch! (`dev/scripts/` vs `scripts/`)

PR #3244, opened 2023-01-29
Add ability to query servers via plugin

### LINKED ISSUE #2017: Plugin Based Status Page
I really want my status page to be like this
https://tensa.co.ua/status/ (also see attached image)
But I cannot do that the way my network is setup

Is it possible that a feature could be added that makes the ping work with the nameless plugin?
I am thinking of a system similar to push notifications.
When the status page is visited it sends a request to the plugin, and the plugin responds with online players. If no response is received then the site says the server is offline.

Link to other github issue: https://github.com/NamelessMC/Nameless-Plugin/issues/101

image

supercrafter100:
Currently we are limited to the query port of a server being open for getting information to display on the status page. This is quite limiting in a few ways. 

This pull request adds the ability to use the plugin as a method of querying. The plugin sends information about the server anyways so we can use this information to get an idea of the server being online and use information about the server which gets sent anyways like the online player count, player list, etc.


supercrafter100:
May add the server version as well if this gets implemented in the plugin data sender. Currently works as is though.

modules/Core/includes/endpoints/ServerInfoEndpoint.php
```diff
@@ -81,6 +81,24 @@ public function execute(Nameless2API $api): void {
             $api->throwError(CoreApiErrors::ERROR_UNABLE_TO_UPDATE_SERVER_INFO, $e->getMessage(), 500);
         }
 
+        // Server query
+        $query_type = Util::getSetting('external_query');
```
tadhgboyle: Does not make sense to keep using `'external_query'` as the setting name now that there are 3 possible values. Could we create a new setting named something like `'server_query_type'`, and use that instead?

modules/Core/includes/endpoints/ServerInfoEndpoint.php
```diff
@@ -81,6 +81,24 @@ public function execute(Nameless2API $api): void {
...
+            }
+
+            $cache = new Cache(['name' => 'nameless', 'extension' => '.cache', 'path' => ROOT_PATH . '/cache/']);
+            $cache->setCache('latest_query');
+            $cache->store($server_id, [
+                'player_count' => count($_POST['players']),
+                'player_count_max' => $_POST['max_players'],
+                'player_list' => $players_list,
+                'motd' => $_POST['motd']
+            ], intval($_POST['interval_seconds'] ?? 10) * 2);
```
tadhgboyle: Will we have to update the call to `validateParams` on line 13 to check for these as well?

modules/Core/module.php
```diff
@@ -923,21 +923,29 @@ public function onPageLoad(User $user, Pages $pages, Cache $cache, Smarty $smart
                         $full_ip = ['ip' => $default->ip . (is_null($default->port) ? '' : ':' . $default->port), 'pre' => $default->pre, 'name' => $default->name];
 
                         // Get query type
-                        $query_type = Util::getSetting('external_query') === '1' ? 'external' : 'internal';
+                        $query_type = Util::getSetting('external_query');
+                        if ($query_type == QueryType::INTERNAL) {
```
tadhgboyle: Rather than storing these as numbers and then converting to/from strings a bunch, can we just store them in the `QueryType` class as string constants?

modules/Core/pages/panel/minecraft_servers.php
```diff
@@ -393,10 +391,9 @@
             $available_parent_servers = DB::getInstance()->get('mc_servers', ['parent_server', 0])->results();
 
             // Display query information alert only if external query is selected
-            $external_query = DB::getInstance()->get('settings', ['name', 'external_query'])->results();
-            $external_query = $external_query[0]->value;
+            $external_query = Util::getSetting('external_query');
```
tadhgboyle: Same opinion here about the setting name (and `$external_query` variable name). These should be something like `$query_type`

modules/Core/pages/panel/minecraft_servers.php
```diff
@@ -502,10 +499,19 @@
...
+                    case 'internal':
+                        $external = QueryType::INTERNAL;
+                        break;
+                    case 'external':
+                        $external = QueryType::EXTERNAL;
+                        break;
+                    case 'plugin':
+                        $external = QueryType::PLUGIN;
+                        break;
+                }
```
tadhgboyle: Same comment about unneeded conversions to/from int and string

tadhgboyle (changes requested):

modules/Core/includes/endpoints/ServerInfoEndpoint.php
supercrafter100: Renamed to `query_type`

modules/Core/module.php
supercrafter100: Storing as strings now, making this check just comparing strings

custom/languages/en_UK.json
```diff
@@ -219,7 +219,7 @@
     "admin/internal": "Internal",
     "admin/external": "External",
     "admin/plugin": "Plugin",
-    "admin/external_query_help": "If the default internal server query does not work, try another option.",
+    "admin/query_type_help": "If the default internal server query does not work, try another option.",
```
tadhgboyle, line 222: Would we want to change this to recommend plugin query instead of internal query?

modules/Core/module.php
```diff
@@ -923,14 +923,7 @@ public function onPageLoad(User $user, Pages $pages, Cache $cache, Smarty $smart
...
                         // Get query type
-                        $query_type = Util::getSetting('external_query');
-                        if ($query_type == QueryType::INTERNAL) {
-                            $query_type = 'internal';
-                        } else if ($query_type == QueryType::EXTERNAL) {
-                            $query_type = 'external';
-                        } else if ($query_type == QueryType::PLUGIN) {
-                            $query_type = 'plugin';
-                        }
+                        $query_type = Util::getSetting('query_type', 'internal');
```
tadhgboyle, line 926: We can remove this fallback, since we have it in the DatabaseInitialiser and the upgrade script we can safely assume it exists

custom/languages/en_UK.json
Derkades: If internal query is available (and it is on most hosting providers), external query is slower and less reliable by comparison. In my opinion issues with internal query are rare enough that it can remain the default

modules/Core/module.php
supercrafter100: Isn't it always safe to have a fallback just in case?

tadhgboyle (dismissed):

tadhgboyle (approved):

PR #3246, opened 2023-02-02
Move `ROOT_PATH` constant to a class

Derkades:
Why?
- It no longer needs to be set manually (like in CLI installer, seeder, etc), reducing the possibility of errors when files are moved
- Tools like phpstan do not like global constants


CHANGELOG.md
```diff
@@ -136,7 +136,7 @@
 - Fix DefaultRevamp dropdown not respecting light/dark mode
 - Fix edge case exception when no `$_active_page` is set in `Pages` class
 - Fix issue with `Forum::getLatestDiscussions()` returning an array of objects randomly
-- Fix missing `ROOT_PATH` definition in Phinx migration file
+- Fix missing `Constants::ROOT_PATH` definition in Phinx migration file
```
supercrafter100: Probably don't want to update this

core/classes/Misc/Constants.php
```diff
@@ -0,0 +1,7 @@
+<?php
+
```
supercrafter100, line 2: Header text?

core/migrations/phinx.php
```diff
@@ -1,7 +1,7 @@
 <?php
 
-if (!defined('ROOT_PATH')) {
-    define('ROOT_PATH', __DIR__ . '/../..');
+if (!defined('Constants::ROOT_PATH')) {
+    define('Constants::ROOT_PATH', __DIR__ . '/../..');
 }
```
supercrafter100: Don't think we need this?

dev/scripts/seeder/db_seeder.php
```diff
@@ -3,7 +3,7 @@
     die('This script must be run from the command line.');
 }
 
-define('ROOT_PATH', __DIR__ . '/../..');
+define('Constants::ROOT_PATH', __DIR__ . '/../..');
```
supercrafter100: Don't need this either anymore?

install.php
```diff
@@ -18,14 +18,14 @@
 // Definitions
 if (!defined('PATH')) {
     define('PATH', '/');
-    define('ROOT_PATH', __DIR__);
+    define('Constants::ROOT_PATH', __DIR__);
```
supercrafter100, line 21: Same as above

supercrafter100 (dismissed):

core/migrations/phinx.php
Derkades: Good find, I didn't check for define() usages

dev/scripts/seeder/db_seeder.php
tadhgboyle: Ye ol' find and replace all

tadhgboyle:
I think we should keep the `ROOT_PATH` constant for 2.1.0, and then remove it in 2.2.0 - just to spread out the massive breaking changes a bit for module devs

index.php
```diff
@@ -43,7 +43,6 @@
 }
 
 const PATH = '/';
-const ROOT_PATH = __DIR__;
```
partydragen, line 46: Should stay as deprecated, this being a huge breaking change for all templates and modules

partydragen (dismissed):

PR #3247, opened 2023-02-02
Remove query interval cache

Derkades:


tadhgboyle (dismissed):

Derkades:
Or do you think this can go into 2.1.0? If so I will rename the upgrade script to 203

tadhgboyle:
> Or do you think this can go into 2.1.0? If so I will rename the upgrade script to 203

Yeah 2.1.0 is fine !

tadhgboyle (approved):

PR #3248, opened 2023-02-05
Run release script with oldest suppported PHP version

Derkades:
For some reason, composer requires at least PHP 8.1 if `composer install` was run on PHP 8.1, even if the platform version is configured differently. To work around this issue, run the release script using the oldest PHP version we support.

dev-oldphp tag was added in commit https://github.com/NamelessMC/Nameless-Docker/commit/7e271820e69412874cd4e21409d8f9de6efb9d4d

### DIFF
--- a/release.sh
+++ b/release.sh
@@ -1,3 +1,3 @@
 #!/usr/bin/env bash
 set -e
-docker run --rm -u $(id -u) -v "$(pwd):/data" --entrypoint="/data/dev/scripts/release-entrypoint.sh" namelessmc/php:dev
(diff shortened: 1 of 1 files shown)


Derkades:
After this is merged, I was planning to backport the commit to the `v2` branch and update the archives attached to the 2.0.3 release. Is that alright?

tadhgboyle (approved):
Thank you

Derkades:
Apparently `v2` is a protected branch. I will open a second PR.

PR #3249, opened 2023-02-05
Run release script with oldest suppported PHP version (backport)

Derkades:


tadhgboyle (approved):

Derkades:
I am not able to merge this pull request, @samerton ?

samerton:
How come we're targeting the v2 branch?

Derkades:
The current 2.0.3 release zip file was built with PHP 8.1, and for some reason that means composer requires at least PHP 8.1 to be able to run it (even though `platform` is set to `7.4`). With this PR, the release script runs as PHP 7.4. Ideally we'd produce new release archives for 2.0.3 without another patch release, which is why this PR targets the `v2` branch.

Derkades:
I have replaced the release archives

PR #3251, opened 2023-02-07
Add unused language term script to CI workflow

tadhgboyle:
Also removes empty `""` strings from files


Derkades:
Please undo changes to language files other than en_UK, as these are managed by Weblate (and Weblate won't be able to handle these conflicts automatically). Weblate will automatically remove unused terms once they are removed from the base language file.

Derkades (approved):
Nice!

PR #3257, opened 2023-02-17
Fixing the duplication bug in messaging.php

Knuvex:
- Fixed issue where message content was duplicated on page refresh

### DIFF
--- a/modules/Core/pages/user/messaging.php
+++ b/modules/Core/pages/user/messaging.php
@@ -405,13 +405,14 @@ if (!isset($_GET['action'])) {
 
                     // Display success message
                     $smarty->assign('MESSAGE_SENT', $language->get('user', 'message_sent_successfully'));
-                    unset($_POST['content']);if ($_SERVER['REQUEST_METHOD'] === 'POST') {
-
-                    // Redirect to same page to avoid resubmitting form on refresh
-                    header('Location: ' . $_SERVER['REQUEST_URI']);
-                    exit;
-                }
-
+                    unset($_POST['content']);
+                    
+                    // Use strict comparison operator to compare request method
+                    if ($_SERVER['REQUEST_METHOD'] === 'POST') {
+                       // Redirect to same page to avoid resubmitting form on refresh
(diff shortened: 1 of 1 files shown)


tadhgboyle:
Thanks for the PR!
Are you able to do a screenrecording of the bug before and after the change?

Derkades:
I will explain
- The browser does a POST request when submitting a form
- Nameless replies to that content with HTML
- The browser renders that HTML and shows it to the user. When the user refreshes the page, the browser has to do the previous request again, which in this case is a POST request. This is potentially unwanted, which is why browsers warn before doing this. However, users ignore warnings and the warning is ugly.

The fix is to do this instead:
- When receiving the POST request, instead of replying with HTML, reply with a redirect instead
- The browser does a second GET request, this time the proper HTML is returned
- Now, when refreshing, the browser will only reissue the second GET request which has no side-effects

modules/Core/pages/user/messaging.php
```diff
@@ -406,6 +406,13 @@
                     // Display success message
                     $smarty->assign('MESSAGE_SENT', $language->get('user', 'message_sent_successfully'));
                     unset($_POST['content']);
+                    
+                    // Use strict comparison operator to compare request method
+                    if ($_SERVER['REQUEST_METHOD'] === 'POST') {
```
samerton, line 411: There should be no need to do this check here as we'll already be within a POST request.

modules/Core/pages/user/messaging.php
```diff
@@ -406,6 +406,13 @@
...
+                       // Redirect to same page to avoid resubmitting form on refresh
+                       header('Location: ' . $_SERVER['REQUEST_URI']);
```
samerton, line 413: We can remove this line and the `exit;` afterwards and simply do the following:
```php
Redirect::to($_SERVER['REQUEST_URI']);
```

samerton (changes requested):
Thanks for this! Just a couple of small things left as comments

PR #3266, opened 2023-02-25
Include missing smarty.js file

Derkades:
This is fixed in `develop`, but I guess it was never backported to v2.0.x.


tadhgboyle:
Don't think we are going to do a 2.0.4!

samerton:
> Don't think we are going to do a 2.0.4!

Indeed, the next release will be 2.1.0 😄

PR #3270, opened 2023-02-28
Misc fixes + improvements

### LINKED ISSUE #2861: Switch off auto language selection
### Describe the feature

Currently the auto language selection seems to be enabled for all users. However I have a site in English and it looks weird if half of it is in another language. It would therefor be nice if the auto language selection could be turned off.

### LINKED ISSUE #3267: Add OG Image upload section
We currently set og:image meta tag on each page load to an image named site_image.png (which doesn't exist), but don't have anywhere for users to set this image.
Additionally, we ship a file in `core/assets/images` named `favicon.png`, but never do anything with it. It can be deleted

tadhgboyle:
- Ability to change user language + timezone from staffcp
- Display active user template in staffcp
- Add "last seen" to user popup, and add tooltips for exact date of last seen + registered
- Limit online staff and users widgets to only showing 10 users
- Add ability to create header in dropdown menus (will be utilized by members module when merged)
- Use `.gitkeep` instead of `placeholder.txt` files
- Add og:image uploader (closes https://github.com/NamelessMC/Nameless/issues/3267)
	- Allow setting specific og:images for each page via SEO config
- Remove old fomantic toast assets
- Remove old datatables plugin files
- Improve email validation error message
- Add `avatar_url` to `UserInfoEndpoint`
- Add fallback message to `LatestPostsWidget`
- Cleanup root directory (move files into `.github` or `dev`)
- Allow disabling auto language feature & improve detection (closes https://github.com/NamelessMC/Nameless/issues/2861)
- Fix latest posts profile tab no posts message not displaying


core/avatar/cache/placeholder.txt
```diff
@@ -1 +0,0 @@
-Placeholder file for avatar cache
```
samerton, line 1: Do we create this directory in code if it doesn't exist?

core/classes/Core/User.php
```diff
@@ -157,32 +126,24 @@ public function find(string $value = null, string $field = 'id'): bool {
      *
      * @param int $group_id ID of group to give.
      * @param int $expire Expiry in epoch time. If not supplied, group will never expire.
-     * @param object|null $group_data Load data from existing query.
      *
      * @return bool True on success, false if they already have it.
      */
-    public function addGroup(int $group_id, int $expire = 0, $group_data = null): bool {
-        if (array_key_exists($group_id, $this->_groups)) {
+    public function addGroup(int $group_id, int $expire = 0): bool {
```
samerton, line 132: Will tag @partydragen for comment on the group rework

core/classes/Core/User.php
```diff
@@ -560,9 +516,38 @@ public function admLogout(): void {
...
      */
     public function getGroups(): array {
+        if (isset($this->_groups)) {
+            return $this->_groups;
+        }
+
+        if (isset(self::$_group_cache[$this->data()->id])) {
+            $groups_query = self::$_group_cache[$this->data()->id];
+        } else {
+            $groups_query = $this->_db->query('SELECT nl2_groups.* FROM nl2_users_groups INNER JOIN nl2_groups ON group_id = nl2_groups.id WHERE user_id = ? AND deleted = 0 ORDER BY `order`;', [$this->data()->id]);
```
samerton: Small thing but the semicolon can be removed from the end of the query here

core/classes/Database/PhinxAdapter.php
```diff
@@ -34,21 +33,7 @@ static function ($file_name) {
             foreach ($missing as $missing_migration) {
                 echo " - $missing_migration" . '<br>';
             }
-        }
-
-        // Something went wonky, either they've deleted migration files,
```
samerton: Are we not expecting this to get called anymore? What if, in a dev environment, I check out from a branch with new migrations back to develop, would I see a descriptive error message?

core/migrations/20230302090631_add_image_column_to_page_descriptions_table.php
```diff
@@ -0,0 +1,13 @@
...
+declare(strict_types=1);
+
+use Phinx\Migration\AbstractMigration;
+
+final class AddImageColumnToPageDescriptionsTable extends AbstractMigration
+{
+    public function change(): void
+    {
+        $table = $this->table('nl2_page_descriptions');
+        $table->addColumn('image', 'string', ['limit' => 512, 'null' => true])->update();
```
samerton: Does `'null' => true` default to null? If not, should we be making null default explicitly?

core/templates/frontend_init.php
```diff
@@ -106,12 +106,18 @@
...
 if (!defined('PAGE_DESCRIPTION')) {
-    $page_metadata = DB::getInstance()->get('page_descriptions', ['page', $route])->results();
-    if (count($page_metadata)) {
+    $page_metadata = DB::getInstance()->get('page_descriptions', ['page', $route]);
+    if ($page_metadata->count()) {
+        $page_metadata = $page_metadata->first();
         $smarty->assign([
-            'PAGE_DESCRIPTION' => str_replace('{site}', Output::getClean(SITE_NAME), $page_metadata[0]->description),
-            'PAGE_KEYWORDS' => $page_metadata[0]->tags
+            'PAGE_DESCRIPTION' => str_replace('{site}', Output::getClean(SITE_NAME), $page_metadata->description),
```
samerton: I wonder if we should be looking to purify description, keywords and image in case there's malicious input?

modules/Core/pages/profile.php
```diff
@@ -711,10 +710,9 @@
             }
             $replies_query = null;
 
-            $target_user = new User($post_user[0]->id);
             $wall_posts[] = [
                 'id' => $nValue->id,
-                'user_id' => Output::getClean($post_user[0]->id),
+                'user_id' => $target_user->data()->id,
```
samerton, line 715: Where is $target_user defined if not on line 714 anymore?

core/avatar/cache/placeholder.txt
tadhgboyle: Magnificat catch, replaced with `.gitkeep` but it was ignored

core/classes/Core/User.php
tadhgboyle: From my investigation, the ability to call `addGroup` and pass existing group data was never used - but could be wrong here!

core/classes/Database/PhinxAdapter.php
tadhgboyle: The situation could still arise in development where migrations have been executed and stored in `nl2_phinxlog` - but I can't see a way that alerting the developer that this happened is beneficial. All the times it has happened to me so far I just go delete the row(s) that shouldn't exist on the current branch and then refresh the page - a slight annoyance.

For example:
- Checkout Member list branch, run the migration to create the member lists table
- Checkout develop, see an error that a migration exists in `nl2_phinxlog` which does not have a related migration file on the branch
I don't know in what case this would be a helpful message to know, since it can't be harmful?

> in a dev environment, I check out from a branch with new migrations back to develop, would I see a descriptive error message?

Not with this code removed, do you think it would be helpful information to know? I'd be fine to add this check back, but so far it's just been an added annoyance when working on multiple features.

modules/Core/pages/profile.php
tadhgboyle: Line 639!

core/classes/Database/PhinxAdapter.php
samerton: If removing this message means the website would load normally despite extra migrations being run, then I'm all for it.

If, and only if, we'd get a fatal error instead then I think we should keep it?

Derkades: > Checkout develop, see an error that a migration exists in nl2_phinxlog which does not have a related migration file on the branch
I don't know in what case this would be a helpful message to know, since it can't be harmful?

The warning makes sense to me. If the migration file does not exist on a branch, that means it still expects the "old" database format. If the migration has already been run (is in the phinxlog table), the database is in the "new" format.

Without this check, you may run into random database errors

core/classes/Core/User.php
partydragen: Was used several places until aber removed it all to requery each time instead instead of using existing data in another group pr you did

core/classes/Database/PhinxAdapter.php
tadhgboyle: Makes sense, reverted. Thanks for the input

core/classes/Core/User.php
```diff
@@ -157,32 +126,24 @@ public function find(string $value = null, string $field = 'id'): bool {
...
      * @param int $group_id ID of group to give.
      * @param int $expire Expiry in epoch time. If not supplied, group will never expire.
-     * @param object|null $group_data Load data from existing query.
      *
      * @return bool True on success, false if they already have it.
      */
-    public function addGroup(int $group_id, int $expire = 0, $group_data = null): bool {
-        if (array_key_exists($group_id, $this->_groups)) {
+    public function addGroup(int $group_id, int $expire = 0): bool {
+        if (array_key_exists($group_id, $this->getGroups())) {
```
supercrafter100, line 133: Doesn't this have to be negated?

core/includes/updates/203.php
```diff
@@ -18,6 +17,10 @@ public function run(): void {
             Util::setSetting('minecraft_query_interval', $query_interval);
         }
 
+        // Replace`external_query` with `query_type`
```
supercrafter100: Spelling error

supercrafter100 (dismissed):

core/classes/Core/User.php
supercrafter100: oh nvm

core/classes/Core/User.php
```diff
@@ -157,32 +126,24 @@ public function find(string $value = null, string $field = 'id'): bool {
...
-            }
-        } else {
-            $this->_groups[$group_id] = new Group($group_data);
+        $this->_db->query('INSERT INTO `nl2_users_groups` (`user_id`, `group_id`, `received`, `expire`) VALUES (?, ?, ?, ?)', [
+            $this->data()->id,
+            $group_id,
+            date('U'),
+            $expire
+        ]);
+
```
supercrafter100, line 143: Wouldn't it be better to check if the group exists before we add it into the database?

core/classes/Core/Navigation.php
```diff
@@ -139,8 +139,9 @@ public function addDropdown(string $name, string $title, string $location = 'top
      * @param string|null $target HTML target attribute (eg '_blank')
      * @param string $icon Icon to prepend to nav item
      * @param int $order Nav item order
+     * @param string $type Type of item, either 'item' or 'header'
      */
-    public function addItemToDropdown(string $dropdown, string $name, string $title, string $link, string $location = 'top', string $target = null, string $icon = '', int $order = 10): void {
+    public function addItemToDropdown(string $dropdown, string $name, string $title, string $link, string $location = 'top', string $target = null, string $icon = '', int $order = 10, string $type = 'item'): void {
```
partydragen: Undo the changes in this file

core/classes/Core/User.php
partydragen: it do check already in the code

modules/Core/includes/endpoints/UserInfoEndpoint.php
```diff
@@ -17,18 +17,14 @@ public function __construct() {
     }
 
     public function execute(Nameless2API $api, User $user): void {
-        $query = 'SELECT nl2_users.id, nl2_users.username, nl2_languages.short_code as `locale`, nl2_users.nickname as displayname, nl2_users.joined as registered_timestamp, nl2_users.last_online as last_online_timestamp, nl2_users.isbanned as banned, nl2_users.active as validated, nl2_users.user_title as user_title FROM nl2_users LEFT JOIN nl2_languages ON nl2_users.language_id = nl2_languages.id';
-
-        // Ensure the user exists
-        $results = $api->getDb()->query($query . ' WHERE nl2_users.id = ?', [(int) $user->data()->id]);
-
-        $return = $results->first();
+        $return = new stdClass();
```
partydragen: Does not user info endpoint now missing alot of values such at displayname, user_title , locale and more

as before it got it all from the sql object directly, but now its a empty object where you need to enter all of them yourself

modules/Core/pages/panel/users_edit.php
```diff
@@ -106,6 +106,10 @@
                     Validate::REQUIRED => true,
                     Validate::MIN => 3,
                     Validate::MAX => 20
+                ],
+                'timezone' => [
```
partydragen, line 110: Are we sure timezones working this time? So we don't get the same issue we had with registration before as some browsers returned invalid timezones or something idk

partydragen (changes requested):

core/classes/Core/User.php
partydragen: it do check already in the code

modules/Core/pages/panel/users_edit.php
tadhgboyle: This won't have the same issue as the one that used to exist during registration, because it does not use JS on the frontend to try and "predict" the timezone; it is just a dropdown select

partydragen (approved):

PR #3274, opened 2023-03-04
Queue system

samerton:
- Create new queue system which allows for tasks to be scheduled and run in the background (using AJAX or cron)


core/classes/Queue/Queue.php
```diff
@@ -96,6 +96,9 @@ public static function process(): array {
                 $fragment = '';
 
                 if (class_exists($task['task'])) {
+                    DB::getInstance()->query('START TRANSACTION');
```
Derkades: I'm not sure if this will work, if PDO remains in autocommit mode? I think you should use `PDO::beginTransaction` instead

samerton: I did some manual testing and it seemed fine, but I'd agree doing this via PDO would be nicer

samerton: Done 👍

samerton:
Screenshot 2023-03-27 at 19 31 05
Forum post HTML seems fine after running the convert forum post task

.gitignore
```diff
@@ -1,3 +1,4 @@
+/.fleet/
```
tadhgboyle, line 1: Ooohh nice, didn't know you were using fleet :o

core/classes/Core/Validate.php
```diff
@@ -71,6 +81,11 @@ class Validate {
      */
     public const NUMERIC = 'numeric';
 
+    /**
+     * @var string Check that the value is one of a set of values
+     */
+    public const ONE_OF = 'one_of';
```
tadhgboyle: Would `IN` work here too? Just a thought, matches `in_array` more...

core/classes/Database/DB.php
```diff
@@ -92,15 +117,13 @@ public function transaction(Closure $closure) {
         $result = null;
```
tadhgboyle, line 117: This `transaction` method should probably just be dropped, not being used anywhere, and now that you've exposed those PDO things, they are much better to use

core/classes/Database/DatabaseInitialiser.php
```diff
@@ -41,7 +41,7 @@ private function initialiseGroups(): void {
...
...:1,"admincp.core.placeholders":1,"admincp.members":1,"admincp.integrations":1,"admincp.integrations.edit":1,"admincp.discord":1,"admincp.minecraft":1,"admincp.minecraft.authme":1,"admincp.minecraft.servers":1,"admincp.minecraft.query_errors":1,"admincp.minecraft.banners":1,"admincp.modules":1,"admincp.pages":1,"admincp.security":1,"admincp.security.acp_logins":1,"admincp.security.template":1,"admincp.styles":1,"admincp.styles.panel_templates":1,"admincp.styles.templates":1,"admincp.styles.templates.edit":1,"admincp.styles.images":1,"admincp.update":1,"admincp.users":1,"admincp.users.edit":1,"admincp.groups":1,"admincp.groups.self":1,"admincp.widgets":1,"modcp.ip_lookup":1,"modcp.punishments":1,"modcp.punishments.warn":1,"modcp.punishments.ban":1,"modcp.punishments.banip":1,"modcp.punishments.revoke":1,"modcp.reports":1,"modcp.profile_banner_reset":1,"usercp.messaging":1,"usercp.signature":1,"admincp.forums":1,"usercp.private_profile":1,"usercp.nickname":1,"usercp.profile_banner":1,"profile.private.bypass":1, "admincp.security.all":1,"admincp.core.hooks":1,"admincp.security.group_sync":1,"admincp.core.emails_mass_message":1,"modcp.punishments.reset_avatar":1,"usercp.gif_avatar":1}',
```
tadhgboyle, line 44: Keep meaning to move these into some nicer arrays 😅

partydragen: Admin group really only need the first administrator perm, they cant view admin perms section anyways

.gitignore
samerton: Gave it a go, couldn't get intellisense working with PHP so it didn't last long 😅

core/classes/Database/DB.php
samerton: On the flipside is it worth keeping for future use since it saves you having to begin, try/catch with a commit/rollback each time?

modules/Core/queries/queue.php
```diff
@@ -0,0 +1,46 @@
...
+
+if ($runner == 'cron') {
+    if (!isset($_GET['cron'])) {
+        return_json('Invalid cron URL', true);
+    }
+
+    if (Input::get('key') != Util::getSetting('cron_key')) {
+        return_json('Invalid cron key', true);
+    }
+}
```
tadhgboyle, line 30: What would an example cron file look like for this?

core/classes/Queue/Queue.php
```diff
@@ -0,0 +1,189 @@
...
+                (
+                 module_id,
+                 task,
+                 name,
+                 data,
+                 scheduled_at,
+                 scheduled_for,
+                 fragment,
+                 fragment_total,
+                 user_id,
```
tadhgboyle, line 36: Struggling to grasp how a task could have a `user_id` when it seems like they will be enqueued and executed by modules?

tadhgboyle (approved):
Code looks great, what kind of additions to our documentation do you think we need?
- Cron file example
- Module task creation example
- ...

tadhgboyle:
A few more thoughts (will add to this comment if more come up):
- How can modules define that a task should repeat, and on a specific schedule?
- It would probably be nice if we displayed somewhere "last cron task execution: xyz", so that people can easily see if their cron file config worked (i recall this being a pain point for the tebex module cron file)

core/classes/Queue/Queue.php
samerton: If a task can be manually triggered by a user, e.g. a mass message, it can be used to store the ID of that user.

samerton:
> A few more thoughts (will add to this comment if more come up):
> 
> * How can modules define that a task should repeat, and on a specific schedule?
> * It would probably be nice if we displayed somewhere "last cron task execution: xyz", so that people can easily see if their cron file config worked (i recall this being a pain point for the tebex module cron file)

The idea for now is that a task is responsible for rescheduling itself if needed, we could maybe look at changing this in the future depending on how we get on with the current system

Agreed RE cron, I'll create a follow-up PR showing when the queue was last run and also prepare to add something to the docs with an example cron

PR #3104, opened 2022-10-10
Add user sessions page in staffcp

supercrafter100:
image

PR #3278, opened 2023-03-08
Add user sessions page

tadhgboyle:
Replaces #3104
image
image


modules/Core/pages/panel/users_sessions.php
```diff
@@ -0,0 +1,120 @@
...
+    } else {
+        $errors[] = $language->get('general', 'invalid_token');
+    }
+}
+$timeago = new TimeAgo(TIMEZONE);
+$sessions = $view_user->getActiveSessions();
+$user_sessions_list = [];
+
+// TODO: Should we display all sessions, or just active ones? Over time, the list could get very long if we display all sessions.
+// Not really any reason to show inactive ones, since they can't action on them.
```
partydragen: Showing inactive ones can help us find if the account has been compromised or not, Hiding sessions could miss unwanted logins

modules/Core/pages/user/sessions.php
```diff
@@ -0,0 +1,109 @@
...
+    $smarty->assign([
+        'ERROR' => $language->get('general', 'error'),
+        'ERROR_MESSAGE' => Session::flash('user_sessions_error'),
+    ]);
+}
+
+$sessions = $user->getActiveSessions();
+$sessions_list = [];
+// TODO: Should we display all sessions, or just active ones? Over time, the list could get very long if we display all sessions.
+// Not really any reason to show inactive ones, since they can't action on them.
```
partydragen: Showing inactive ones can help us find if the account has been compromised or not, Hiding sessions could miss unwanted logins

core/assets/GeoLite2-Country.mmdb
samerton, line 1: Does this file need adding as a Deny to the .htaccess and blocking within nginx too, so that it can't be accessed via the browser

Derkades: It's a public file, but redistribution requires attribution so just to avoid any nastiness making it inaccessible can't hurt

samerton (dismissed):

PR #3281, opened 2023-03-21
Create `HasWebhookParams` interface

### LINKED ISSUE #3275: available_hooks for events no longer working
available_hooks for events no longer working, Forum events keep firing the discord messages when the webhooks are not even selected

tadhgboyle:
*Alongside https://github.com/NamelessMC/Nameless/pull/3226*

Allows events to set specific params to send when being transformed into a JSON object for webhooks.
This is useful for when events contain some params which are objects/classes, since they cannot be nicely formatted into JSON.

For example, if we want to be able to pass the users ID to a webhook when the `userRegistered` event is fired:
```php
class UserRegisteredEvent implements HasWebhookParams {
  public User $user;

  function __construct(User $user) {
    $this->user = $user;
  }

  function webhookParams(): array {
    return [
      'user_id' => $this->user->data()->id
    ];
  }
}
```

When previously this would have instead returned something like:
```php
['user' => User object]
```

### DIFF
--- a/core/classes/Events/EventHandler.php
+++ b/core/classes/Events/EventHandler.php
@@ -42,7 +42,8 @@ class EventHandler {
         if (class_exists($event)  && is_subclass_of($event, AbstractEvent::class)) {
             $class_name = $event;
             $name = $event::name();
-            $description = $event::description();
+            // We lazy load descriptions for class-based events to avoid loading new Language instances unnecessarily
+            $description = fn () => $event::description();
             $return = $event::return();
             $internal = $event::internal();
         } else {
@@ -210,6 +211,9 @@ class EventHandler {
 
         foreach (self::$_events as $name => $meta) {
             if (!$meta['internal'] || $internal) {
+                if (is_callable($meta['description'])) {
+                    $meta['description'] = $meta['description']();
+                }
                 $return[$name] = $meta['description'];
             }
         }
--- a/core/classes/Events/EventHandler.php
+++ b/core/classes/Events/EventHandler.php
@@ -205,11 +205,11 @@ class EventHandler {
      *
      * @return array List of all currently registered events
      */
-    public static function getEvents(): array {
+    public static function getEvents(bool $showInternal = false): array {
         $return = [];
 
         foreach (self::$_events as $name => $meta) {
-            if ($meta['internal']) {
+            if ($meta['internal'] && !$showInternal) {
                 continue;
             }
 
--- a/core/classes/Events/EventHandler.php
+++ b/core/classes/Events/EventHandler.php
@@ -203,19 +203,28 @@ class EventHandler {
     /**
      * Get a list of events to display on the StaffCP webhooks page.
      *
-     * @param bool $internal Whether to include internal events or not
      * @return array List of all currently registered events
      */
-    public static function getEvents(bool $internal = false): array {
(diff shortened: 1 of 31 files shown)


partydragen:
This pr is not done right, As all events need to get the params added

core/classes/Events/EventHandler.php
```diff
@@ -202,16 +203,28 @@ public static function executeEvent($event, array $params = []): ?array {
     /**
      * Get a list of events to display on the StaffCP webhooks page.
      *
-     * @param bool $internal Whether to include internal events or not
      * @return array List of all currently registered events
      */
-    public static function getEvents(bool $internal = false): array {
+    public static function getEvents(): array {
```
samerton: What are the effects of removing `$internal` here?

custom/panel_templates/Default/core/hooks_edit.tpl
```diff
@@ -64,18 +64,27 @@
...
                                 <div class="form-group custom-control custom-switch">
-                                    <input type="checkbox" id="inputevents[{$key|escape}]" name="events[{$key|escape}]"
-                                        class="custom-control-input" value="1" {if in_array($key|escape,
-                                        $ENABLED_HOOKS)} checked{/if}>
+                                    <input type="checkbox" id="inputevents[{$key|escape}]" name="events[{$key|escape}]" class="custom-control-input" value="1" {if in_array($key|escape, $ENABLED_HOOKS)} checked{/if}>
                                     <label class="custom-control-label" for="inputevents[{$key|escape}]">
-                                        {$item|escape}
+                                        {$meta.description|escape}
+                                        {if $meta.supports_discord}
+                                            <span class="badge badge-info">Supports Discord</span>
```
samerton: Can these be translated?

custom/panel_templates/Default/core/hooks_new.tpl
```diff
@@ -64,19 +64,24 @@
...
-                                        {$item|escape}
-                                    </label>
-                                </div>
+                                {foreach from=$ALL_EVENTS key=key item=meta}
+                                    <div class="form-group custom-control custom-switch">
+                                        <input type="checkbox" id="inputevents[{$key|escape}]" name="events[{$key|escape}]" class="custom-control-input">
+                                        <label class="custom-control-label" for="inputevents[{$key|escape}]">
+                                            {$meta.description|escape}
+                                            {if $meta.supports_discord}
+                                                <span class="badge badge-info">Supports Discord</span>
```
samerton: Can these be translated?

samerton (dismissed):
To confirm, is this for 2.1.0 or 2.2.0?

partydragen:
Closes https://github.com/NamelessMC/Nameless/issues/3275

core/classes/Events/EventHandler.php
tadhgboyle: The only place this is used internally is for the event list in the webhook creation page, and by modules is the websend module, which does not need internal events as well.
There should be no issues removing this

partydragen (dismissed):

tadhgboyle:
> To confirm, is this for 2.1.0 or 2.2.0?

2.1.0, as it has no more breaking changes than already existed in the main event revamp PR

partydragen (approved):

PR #3285, opened 2023-03-24
Add support to customize OAuth login buttons

tadhgboyle:
The main goal of this PR is to make our Google OAuth login button adhere to their branding guidelines, which has been an issue for multiple users previously, since they cannot get their site approved on the verification process.

Buttons now look like this:
image
image


modules/Core/pages/panel/registration.php
```diff
@@ -163,7 +163,8 @@
     $oauth_provider_data[$provider_name] = [
         'enabled' => NamelessOAuth::getInstance()->isEnabled($provider_name),
         'setup' => NamelessOAuth::getInstance()->isSetup($provider_name),
-        'icon' => $provider_data['icon'],
+        'icon' => $provider_data['icon'] ?? null,
```
samerton, line 166: Forgive me here but is `?? null` not necessary because `??` will only evaluate as true if `$provider_data['icon']` is already null?

modules/Core/pages/register.php
```diff
@@ -429,6 +429,7 @@
     'OAUTH_FLOW' => $oauth_flow,
     'OAUTH_AVAILABLE' => NamelessOAuth::getInstance()->isAvailable(),
     'OAUTH_PROVIDERS' => NamelessOAuth::getInstance()->getProvidersAvailable(),
+    'CONTINUE_WITH' => $language->get('user', 'continue_with'),
```
samerton: I think this needs to be changed to be within each provider, with the term being `Continue with {{provider}}`, and the provider name getting substituted in via i18next if that makes sense - the reason being not every language will have the same structure for this sentence

for example there'd be an array map with all providers, adding in the following item (off the top of my head 😄)
```php
[
  // ... rest of provider
  'continue_with' => $language->get('user', 'continue_with_x', ['provider' => $provider->getName()])
]
```

samerton (changes requested):

modules/Core/pages/register.php
tadhgboyle: Ah yeah great call, always forget about this

modules/Core/pages/register.php
```diff
@@ -428,7 +436,8 @@
     'OR' => $language->get('general', 'or'),
     'OAUTH_FLOW' => $oauth_flow,
     'OAUTH_AVAILABLE' => NamelessOAuth::getInstance()->isAvailable(),
-    'OAUTH_PROVIDERS' => NamelessOAuth::getInstance()->getProvidersAvailable(),
+    'OAUTH_PROVIDERS' => $providers,
+    'CONTINUE_WITH' => $language->get('user', 'continue_with'),
```
samerton: Do we still need this now? 😄

modules/Core/pages/login.php
```diff
@@ -272,7 +280,8 @@
     'ERROR' => ($return_error ?? []),
     'NOT_REGISTERED_YET' => $language->get('general', 'not_registered_yet'),
     'OAUTH_AVAILABLE' => NamelessOAuth::getInstance()->isAvailable(),
-    'OAUTH_PROVIDERS' => NamelessOAuth::getInstance()->getProvidersAvailable(),
+    'OAUTH_PROVIDERS' => $providers,
+    'LOGIN_WITH' => $language->get('user', 'log_in_with'),
```
samerton: Do we still need this now? 😄

samerton (approved):

PR #3290, opened 2023-03-28
Fix permissions not being respected for news posts

### LINKED ISSUE #3271: Forum `getLatestNews` does not consider user permissions
### Describe the issue

When a topic which a user cannot view (admin-only forum, but the logged in user is a moderator, for example) is made a news post, it will still be displayed on the front page, and when the user clicks on it to view more, they get a 403.

### NamelessMC Version

Development version

### To Reproduce

_No response_

### Expected Behaviour

_No response_

### Screenshots

_No response_

### Additional Information

_No response_

samerton:
Fixes #3271

### DIFF
--- a/modules/Forum/classes/Forum.php
+++ b/modules/Forum/classes/Forum.php
@@ -480,23 +480,56 @@ class Forum {
     /**
      * Get the latest news posts to display on homepage.
      *
+     * TODO: can be optimised by including posts in the initial query?
+     *
      * @param int $number The number of posts to get.
+     * @param array $groups Array containing user's group IDs, default [0] for guests
      * @return array The latest news posts.
      */
-    public function getLatestNews(int $number = 5): array {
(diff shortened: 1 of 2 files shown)


modules/Forum/classes/Forum.php
```diff
@@ -480,23 +480,56 @@ public function getIndividualPost(int $post_id) {
...
+                            FROM nl2_forums_permissions p
+                            WHERE p.group_id IN ($groups_in)
+                            AND p.view = 1
+                        )
+                    )
+                    AND deleted = 0
+                ORDER BY topic_date
+                DESC LIMIT ?
+            SQL,
+            [...$groups, $number]
```
Aberdeener, line 521: I don't see where `...$groups` goes in the query, am I missing something?

tadhgboyle (approved):

PR #3291, opened 2023-03-28
Disable the cursor pointer on the Core Module

PadowYT2:
As you cannot disable/enable it, why bother to confuse the user by adding a cursor pointer?

### DIFF
--- a/core/installation/steps/select_modules.php
+++ b/core/installation/steps/select_modules.php
@@ -35,6 +35,10 @@ if ($_SERVER['REQUEST_METHOD'] === 'POST') {
     $_SESSION['modules_selected'] = true;
     Redirect::to('?step=conversion');
 }
+
+if ($module !== 'Core') {
+    $pointer = "cursor: pointer;";
+}
 ?>
 
 <form action="" method="post">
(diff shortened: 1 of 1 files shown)


tadhgboyle (dismissed):

core/installation/steps/select_modules.php
PadowYT2, line 1: *code-style

core/installation/steps/select_modules.php
```diff
@@ -35,6 +35,10 @@
     $_SESSION['modules_selected'] = true;
     Redirect::to('?step=conversion');
 }
+
+if ($module !== 'Core') {
```
tadhgboyle: This should be a function instead, which takes a $module parameter and returns the string, for two reasons:
- we shouldn't echo a variable which may not exist 
- $module will always be the same value at this point, since you did not place this logic in the for loop, no matter what module we are making html for, $module will be the same unchanged value (whatever was the last value of it in the for loop above)

tadhgboyle (changes requested):

core/installation/steps/select_modules.php
partydragen: Core module should just not be listed at all is my point 😆

PadowYT2: The thing is it works, but unlisting it will be just kinda bad to hide from the user, no?

Derkades: What does a user get from knowing a core module exists? It seems like a technical detail to me

tadhgboyle (changes requested):
As mentioned before, this does not actually work - I've tested locally now (everything has the `cursor: pointer;` applied).
The $module variable on line 39 will never not be the last element of the foreach loop on line 25.
Instead this should be a function which takes a module name and returns the css depending if the module name is Core.

tadhgboyle (approved):

PR #3292, opened 2023-03-28
Remove the blank space for the modules

PadowYT2:
Before:
image

After:
image


PadowYT2:
Closed due to aber saying that if there will be more modules after forum, there will be a large gap, the best way is to just keep the current way

PR #3293, opened 2023-03-28
Remove the v1 converter during installation

PadowYT2:
Due to https://canary.discord.com/channels/246705793066467328/434751012428054530/1090382131202633771


core/installation/steps/upgrade_perform.php
```diff
@@ -1,46 +0,0 @@
-<?php
-
-$scripts = [
-    '
-    <script>
-        $(document).ready(function() {
-            $.post("?step=ajax_initialise&initialise=upgrade", {perform: "true"}, function(response) {
```
tadhgboyle, line 7: In the `ajax_initialise` file, there are more references and code that does the actual conversion - that should all be torn out too

core/installation/installer.php
```diff
@@ -6,7 +6,7 @@
 }
 
 if (isset($_GET['do'])) {
-    $_SESSION['action'] = ($_GET['do'] == 'upgrade' ? 'upgrade' : 'install');
+    $_SESSION['action'] = 'install';
```
tadhgboyle, line 9: These lines can be removed if we link directly to `requirements_validation` (see suggestion comment)

core/installation/installer.php
```diff
@@ -56,15 +56,10 @@
...
                                 <p><?php echo $language->get('installer', 'terms_and_conditions'); ?></p>
                                 <div class="ui message"><?php echo $nameless_terms; ?></div>
-                                <div class="ui divider"></div>
-                                <p><?php echo $language->get('installer', 'new_installation_question'); ?></p>
                             </div>
                             <div class="ui right aligned secondary segment">
-                                <a href="?do=upgrade" class="ui small button">
-                                    <?php echo $language->get('installer', 'upgrading_from_v1'); ?>
-                                </a>
                                 <a href="?do=install" class="ui small primary button">
```
tadhgboyle, line 61:
```suggestion
                                <a href="?step=requirements_validation" class="ui small primary button">
```

core/installation/steps/ajax_initialise.php
```diff
@@ -27,23 +23,6 @@
                 ];
 
                 $_SESSION['site_initialized'] = true;
```
tadhgboyle, line 25: This `if` statement can be pulled into the `else` right above it just to remove a layer of indentation

tadhgboyle (changes requested):

core/installation/installer.php
PadowYT2: But then what about
https://github.com/NamelessMC/Nameless/blob/4ce36c752e15f8f8b714eacde3f71264ecf9d063/core/installation/installer.php#L75-L78

and

https://github.com/NamelessMC/Nameless/blob/4ce36c752e15f8f8b714eacde3f71264ecf9d063/core/installation/steps/requirements_validation.php#L3-L7

tadhgboyle (approved):

PR #3295, opened 2023-03-29
Fix weird errors during admin creation

PadowYT2:
### Before
image

### After
image


core/installation/steps/admin_account_setup.php
```diff
@@ -45,7 +45,15 @@ function display_error(string $message) {
...
-                display_error($language->get('installer', 'input_required'));
+                if (str_contains($item, 'username is required')) {
+                    display_error($language->get('installer', 'username_required'));
+                } else if (str_contains($item, 'email is required')) {
+                    display_error($language->get('installer', 'email_required'));
+                } else if (str_contains($item, 'password is required')) {
+                    display_error($language->get('installer', 'password_required'));
+                } else if (str_contains($item, 'password_again is required')) {
+                    display_error($language->get('installer', 'passwords_must_match'));
+                }
```
supercrafter100: I think a better solution is to provide the messages straight to the validation instead of doing this wonky solution. See https://phpdoc.namelessmc.com/classes/Validate.html#method_messages . We can then just display the error messages that came from the validation

supercrafter100 (changes requested):

supercrafter100 (approved):

tadhgboyle (approved):

PR #3299, opened 2023-03-29
:tada: Implement dark mode for the installer

PadowYT2:
(in the select modules, i fixed the text in the 47f6f49848615dbf0b3a47cb086e2a14da37a8f6 commit)

https://user-images.githubusercontent.com/71085027/228527334-da0382bb-e5f7-4807-ba69-e1528ab46c16.mp4

image


tadhgboyle:
Not really in favor of the added maintenance costs here for minimal gain

PR #3301, opened 2023-03-30
Use CSS instead of the script to put the footer at the bottom

PadowYT2:


samerton:
Before merging we could do with testing on Chrome, Edge, Firefox, Safari

Also different screen sizes

tadhgboyle:
Hmm, this seems to not have any display issues, but still has the flicker that the JS solution had. I *assumed* that by doing this via CSS we wouldn't have any flicker on page load...

tadhgboyle (approved):

PR #3307, opened 2023-04-02
Revert validateToken

partydragen:
Broke in https://github.com/NamelessMC/Nameless/pull/3175/files


PadowYT2:
But the function is not used?

partydragen:
> But the function is not used?

It is used, its used for registration and any frontend validation this completly disabled recaptcha from working this fix was added to namelessmc website a while ago

tadhgboyle (approved):

PR #3313, opened 2023-04-11
Revamp avatar system, remove MC avatars from core

tadhgboyle:


custom/panel_templates/Default/core/avatars.tpl
```diff
@@ -38,78 +38,59 @@
...
-                                <div class="form-group">
-                                    <label for="selectDefaultAvatar">{$SELECT_DEFAULT_AVATAR}</label>
-                                    <select class="image-picker show-html" id="selectDefaultAvatar" name="avatar">
-                                        {foreach from=$IMAGES key=key item=item}
-                                        <option data-img-src="{$key}" value="{$item}" {if $DEFAULT_AVATAR_IMAGE eq
-                                            $item} selected{/if}>{$item}</option>
+                                <table class="table table-borderless table-striped">
+                                    <thead>
+                                        <tr>
+                                            <th>Name</th>
```
partydragen, line 53: Hardcoded text

modules/Core/module.php
```diff
@@ -435,14 +435,24 @@ public function __construct(Language $language, Pages $pages, User $user, Naviga
...
+            MinecraftAvatarSource::registerSource(new CrafatarMinecraftAvatarSource());
+            MinecraftAvatarSource::registerSource(new CraftheadMinecraftAvatarSource());
+            MinecraftAvatarSource::registerSource(new CravatarMinecraftAvatarSource());
+            MinecraftAvatarSource::registerSource(new MCHeadsMinecraftAvatarSource());
+            MinecraftAvatarSource::registerSource(new MinotarMinecraftAvatarSource());
+            MinecraftAvatarSource::registerSource(new NamelessMCMinecraftAvatarSource($language));
+            MinecraftAvatarSource::registerSource(new VisageMinecraftAvatarSource());
+        }
+
+//        EventHandler::registerListener(UserIntegrationUnlinkedEvent::class, ResetAvatarCacheHook::class);
```
partydragen, line 455: Those should be uncommented?

partydragen: or wait, this is listeners

modules/Core/classes/Avatars/UploadedImageAvatarSource.php
```diff
@@ -0,0 +1,38 @@
...
+    public function __construct(Language $language) {
+        $this->_name = $language->get('admin', 'avatar_source_uploaded_image');
+        $this->_module = 'Core';
+        $this->_settings = ROOT_PATH . '/modules/Core/includes/admin_avatar_settings/uploaded_image.php';
+    }
+
+    public function get(User $user): ?string {
+        $base_url = ($this->_full_url ? rtrim(URL::getSelfURL(), '/') : '') . ((defined('CONFIG_PATH')) ? CONFIG_PATH . '/' : '/') . 'uploads/avatars';
+
+        if (Settings::get('custom_user_avatars') && $user->data()->has_avatar) {
```
partydragen, line 14: Allow custom user avatars? setting on custom avatar source does not save on enabled state

modules/Core/classes/Avatars/UploadedImageAvatarSource.php
```diff
@@ -0,0 +1,38 @@
...
+            $exts = ['png', 'jpg', 'jpeg'];
+
+            if ($user->hasPermission('usercp.gif_avatar')) {
+                $exts[] = 'gif';
+            }
+
+            foreach ($exts as $ext) {
+                if (file_exists(ROOT_PATH . "/uploads/avatars/{$user->data()->id}.{$ext}")) {
```
partydragen, line 22: Need to be a way to remove or unselect default custom avatar

People most likey want Custom avatar to be priority one, and minecraft next but if they have uploaded a custom default image then they are not able to fallback to minecraft avatar

partydragen (changes requested):

partydragen (changes requested):
And that Minecraft Avatar Source should be default priority, Well below Uploading Image source type

So new sites still using Minecraft avatars by default as its still a CMS designed for Minecraft

core/classes/Avatars/AvatarSource.php
```diff
@@ -1,241 +1,127 @@
...
-                $uuid = $data->uuid;
-            } else {
-                $uuid = $data->username;
-                // Fallback to steve avatar if they have an invalid username
-                if (preg_match('#[^][_A-Za-z0-9]#', $uuid)) {
-                    $uuid = 'Steve';
+            $cache_key = $user_id . '_' . $source->getSafeName() . '_' . $size . '_' . (int) $full_url;
+            if ($this->_cache->isCached($cache_key)) {
+                $avatar = $this->_cache->retrieve($cache_key);
+                if ($avatar) {
```
supercrafter100, line 45: Isn't this check a bit redundant? Seeing as you already checked if it's in there?

core/classes/Avatars/AvatarSource.php
```diff
@@ -1,241 +1,127 @@
...
             }
-        } catch (Exception $ignored) {
         }
 
-        $cache->store($url, $is_valid, 3600);
-        return $is_valid;
-    }
+        // Fallback to initials avatar
+        if (!isset($url)) {
+            $url = $this->_sources[InitialsAvatarSource::class]->getAvatar($user, $size, $full_url);
```
supercrafter100, line 65: You never store this in cache if it fails, which would mean we attempt to re-fetch it from the respective avatar source every page load until it works. Is this intentional?

modules/Core/classes/Avatars/MinecraftAvatarSource.php
```diff
@@ -0,0 +1,135 @@
...
+            $identifier = $user->data()->username;
+            // Fallback to steve avatar if they have an invalid username
+            // TODO: what is this regex?
+            if (preg_match('#[^][_A-Za-z0-9]#', $identifier)) {
+                $identifier = 'Steve';
+            }
+        }
+
+        return self::getAvatarFromIdentifier($identifier, $this->_size);
+    }
```
supercrafter100, line 134: We should probably check if the uuid is an offline one and use usernames instead if this is the case. This should fix the fact that when force premium accounts is disabled, people only receive steve skins. Nameless generates an offline uuid when force premium accounts is enabled which would still make this use that offline uuid for retrieving accounts

PR #3317, opened 2023-04-16
Use utf8mb4_unicode_ci for DB class to match Phinx

### LINKED ISSUE #3179: phinx and db classes having mismatching collation
phinx will need to listen on the config to avoiding mismatching collation for old installations and if we change the default in future to prevent this issue happning again

And update so the default match for both phinx and db class

samerton:
Fixes #3179


partydragen (approved):
Tested and working :D

tadhgboyle (approved):

PR #3323, opened 2023-04-21
refactored the web.config.example file

anubhav1206:


Derkades:
Can you explain your changes? To me it seems like the previous indentation was correct

anubhav1206:
Looks cleaner in my opinion. We can revert the changes if we want.

PR #3326, opened 2023-04-23
Allow correct webhook action types in endpoint

supercrafter100:
The webhook types are numbers (1 = normal, 2 = discord) but these types were not accepted in the createwebhook endpoint. This fixes that


tadhgboyle (approved):

tadhgboyle:
Would be nice to maybe make these constants in some class, like `Webhook::TYPE_NORMAL`, `Webhook::TYPE_DISCORD`

PR #3329, opened 2023-04-27
Cancellable events

partydragen:


tadhgboyle:
Should EventHandler also be updated so that it checks if the event is instance of cancellable and not continue executing listeners if it has been cancelled?

partydragen:
> Should EventHandler also be updated so that it checks if the event is instance of cancellable and not continue executing listeners if it has been cancelled?

Hm not sure, It could may be listeners still want to know they should check if event is already cancelled or not etc but i guess none will need that so it saves performance to not continue executing listeners

samerton (approved):

PR #3331, opened 2023-04-30
Make sure module is enabled to process task

partydragen:
@samerton Why is in_progress status not there? https://github.com/NamelessMC/Nameless/blob/develop/core/classes/Queue/Queue.php#L67-L78

What is continue the processing? if any at all

### DIFF
--- a/core/classes/Queue/Queue.php
+++ b/core/classes/Queue/Queue.php
@@ -65,9 +65,9 @@ class Queue {
 
         $pendingTasks = DB::getInstance()->query(
             <<<SQL
-            SELECT `id`, `task`, `name`, `attempts`, `output`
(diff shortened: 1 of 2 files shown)


samerton:
> @samerton Why is in_progress status not there? https://github.com/NamelessMC/Nameless/blob/develop/core/classes/Queue/Queue.php#L67-L78
> 
> What is continue the processing? if any at all

In progress is when the task is actually being executed:

1. Queue picks up tasks in ready/error status
2. Queue sets task to in progress
3. Queue runs code for task
4. Queue updates task status accordingly

So as a developer, you should never set a task to in progress.

samerton (approved):

PR #3332, opened 2023-04-30
Add dependency injection library + task to generate sitemap

samerton:


modules/Core/queries/queue.php
```diff
@@ -40,7 +52,12 @@ function return_json($output, ?bool $error = false) {
 // Update last run immediately
 $cache->store('last_run', $date);
 
+// Load modules + template
+Module::loadPage($user, $pages, $cache, $smarty, [$navigation, $cc_nav, $staffcp_nav], $widgets, $template);
```
samerton, line 56: Adding this to the queue query allows tasks to tap into more data such as loaded pages on the site

partydragen (approved):

supercrafter100 (approved):

PR #3351, opened 2023-05-03
Quick fix for ghost player.

Wgis:
If players on server is equal 0 - i see one head, but without information. May be it's problem with $SERVER.format_player_list ? Anyway this quick fix can help...

### DIFF
--- a/core/classes/Core/Paginator.php
+++ b/core/classes/Core/Paginator.php
@@ -176,9 +176,6 @@ class Paginator {
             $html .= '</ul>';
         }
 
-        if ($start == $end){
-            $html = '';
-        }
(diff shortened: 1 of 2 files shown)


core/classes/Core/Paginator.php
```diff
@@ -176,6 +176,9 @@ public function generate(int $links, string $href = '?'): string {
             $html .= '</ul>';
         }
 
+        if ($start == $end){
+            $html = '';
+        }
```
tadhgboyle: This change seems unrelated to the description of the PR?

tadhgboyle (approved):

PR #3360, opened 2023-05-08
Remove unable to update groups catch

Derkades:
Catching all exceptions and returning a custom 500 error doesn't seem to improve API usability, and it hides valuable debug info. If the API returns a stack trace, the plugin will show it to the user making bug reports easier.


partydragen:
Its not hidden at all its displayed in meta while it make it return json

all api results should return in json even if its a error, API is also not just for the plugin only

tadhgboyle (approved):

Derkades:
> all api results should return in json even if its a error

It should for expected errors, but I wouldn't say it should still return json for programming errors? There is an actual bug here that is not supposed to ever happen. We happen to have a try-catch here, but usually programming errors wouldn't be catched like this

PR #3374, opened 2023-05-14
Remove discord discriminator requirement

partydragen:


samerton:
Does this still work for users that have not converted to the new Discord username system yet?

partydragen:
> Does this still work for users that have not converted to the new Discord username system yet?

Yes, Just that they does not get the tag anymore with is soon removed anyways

the username field is not really used for anything anyways, the username field is already outdated if they change their name or tag anyways as there is nothing that keeping the username up to date

PadowYT2:
> > Does this still work for users that have not converted to the new Discord username system yet?
> 
> Yes, Just that they does not get the tag anymore with is soon removed anyways
> 
> the username field is not really used for anything anyways, the username field is already outdated if they change their name or tag anyways as there is nothing that keeping the username up to date

With the old system, you know you can have the same username on different accounts, wouldn't that be a risk?

tadhgboyle (approved):

Derkades:
> With the old system, you know you can have the same username on different accounts, wouldn't that be a risk?

I don't think the Discord username is ever used as an identifier, so that shouldn't be an issue

tadhgboyle:
Worth noting: as of JSA 5.0.0 beta10 `getName()` returns just their username. No discriminator. This is why a user saw this error https://discord.com/channels/246705793066467328/766713002489282680/1118283515352580116

We should update NamelessLink to use `getAsTag()` instead, until the username changes are rolled out globally on Discords end:
https://github.com/discord-jda/JDA/blob/e1103d99de516815c8b8f001c3b5187987e57c56/src/main/java/net/dv8tion/jda/internal/entities/UserImpl.java#L119-L125

Derkades:
I switched to getName because getAsTag was deprecated in a [recent JDA update](https://github.com/discord-jda/JDA/releases/tag/v5.0.0-beta.10). I will update Nameless-Link to include a discriminator for now

PR #3380, opened 2023-05-23
Support custom migration tables

### LINKED ISSUE #3213: Make phinx support modules
Allow modules to use the phinx system 

Example as @Aberdeener mentioned
```php
class Module {
	public static function migrationDirs() {
	  $dirs = [];
	  $modules = Module::getModules();
	  foreach ($module as $module) {
	    if (dir_exists($migration_path = '/path/to/$module/migrations') {
	      $dirs[] = $migration_path;
	    }
	  }
	
	  return $dirs;
	}
}

// then in phinx.php
'paths' => [
  'migrations' => [__DIR__, ...Module::migrationDirs()],
],
```

samerton:
Closes #3213

### DIFF
--- a/core/classes/Database/PhinxAdapter.php
+++ b/core/classes/Database/PhinxAdapter.php
@@ -95,10 +95,12 @@ class PhinxAdapter {
         string $module,
         ?string $migrationDir = null
     ): string {
-        if ($module === 'Core') {
+        $module = strtolower($module);
+
+        if ($module === 'core') {
             $table = 'nl2_phinxlog';
         } else {
-            $module = strtolower(preg_replace('/[^a-zA-Z]+$/', '', $module));
+            $module = preg_replace('/[^a-zA-Z]+$/', '', $module);
             $table = "nl2_phinxlog_$module";
         }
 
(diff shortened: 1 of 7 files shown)


core/classes/Database/PhinxAdapter.php
```diff
@@ -73,4 +121,52 @@ public static function migrate(): string {
...
+     *
+     * @param string $module Module name
+     * @param string $migrationDir Migration directory to use
+     * @param int $since Version of earliest migration to rollback, default 0 for all
+     *
+     * @return string Output of the migration command from Phinx as if it was executed in the console.
+     *
+     * @throws Exception If unable to rollback
+     */
+    public static function rollback(
```
tadhgboyle, line 136: Good addition - I wonder where/how we could implement this

tadhgboyle (approved):
Looks good! We should remember to update the docs to show a quick guide on how module devs can use this

core/classes/Database/PhinxAdapter.php
samerton: The use case I had in mind was for `onUninstall` for modules - we can remove the database tables when uninstalling a module

partydragen:
HYPE!

PR #3384, opened 2023-06-04
Add Logo Image to Navbar

BijjuXD:
If LOGO_IMAGE is added in Layout->Images, the image will show instead of SITE_NAME, if no logo set, it will show SITE_NAME


PadowYT2:
Where is `LOGO_IMAGE` defined? Also this is completly not necessary to do. Also 100px is completly not accurate, you should be using the size x * 2 starting from 1, so 128px is ideal

BijjuXD:
It was predefined in line 117,
Well going to fix size when I'm back on laptop

supercrafter100:
Wouldn't it make more sense to put this behind a toggle switch? In case we want to use the logo image somewhere else? Ideally somewhere in the template settings then I guess. People might want both worlds as well (text & logo) 🤷‍♂️

BijjuXD:
Oh yeah that's a great idea!

PR #3386, opened 2023-06-09
Call `compileQueries` only when needed

tadhgboyle:
Shaves off a few ms each page load and lots of method calls to `SQLFormatter`


Derkades:
Nice

supercrafter100 (approved):

PR #3389, opened 2023-06-10
Small misc improvements

samerton:


core/classes/Core/URL.php
```diff
@@ -85,6 +85,16 @@ private static function buildNonFriendly(string $url, string $params): string {
         return (defined('CONFIG_PATH') ? CONFIG_PATH : '') . '/index.php?route=' . $url . ((substr($url, -1) == '/') ? '' : '/');
     }
 
+    /**
+     * Build an asset path
+     *
+     * @param string $path Contains the asset path relative to the root Nameless directory
+     * @return string
+     */
+    public static function buildAssetPath(string $path): string {
```
tadhgboyle, line 94: Reminder to self to do a bulk replacement and utilize this everywhere! Great addition

core/classes/Database/DB.php
```diff
@@ -474,9 +474,9 @@ public function addColumn(string $table, string $column, string $attributes): bo
      * column, operator (default =), value, and glue (default AND).
      * @return array The where clause string, and parameters to bind.
      */
-    private function makeWhere(array $clauses): array {
+    public static function makeWhere(array $clauses): array {
```
tadhgboyle, line 477: I admit I am still unsure how making this public helps modules at all - but won't consider this blocking.
Could you explain just so I can get a better understanding of how this method could be used by a module without going through the `where` method (which implicitly calls this already)?

tadhgboyle (approved):

core/classes/Database/DB.php
samerton: Here's an example: https://github.com/samerton/Nameless-Resources/blob/develop/upload/modules/Resources/classes/Resources.php#L121

It's to enable an easier way to pass parameters such as pagination and filters into methods which will perform some SQL, e.g. a repository class

partydragen: SMART!!!! This ise usefull many places!

partydragen (approved):

PR #3391, opened 2023-06-11
Remove redundant language term

Justman10000:


PadowYT2 (changes requested):
The files permissions were changed, should be bad...?

Justman10000:
> The files permissions were changed, should be bad...?

?

Justman10000:
@partydragen

core/classes/DTO/IntegrationData.php
```diff
@@ -5,15 +5,15 @@ class IntegrationData {
     public int $id;
     public string $name;
     public bool $enabled;
-    public bool $can_unlink;
+    public bool $can_be_unlink;
```
Aberdeener, line 8: This change is not something we will consider merging - changing a column of the user database is not something we want to do. Your initial commit was good, could you revert back to just that?

tadhgboyle (dismissed):

core/classes/DTO/IntegrationData.php
Justman10000: Then that's the way it is? What's wrong with that?

Aberdeener: Sorry, I don't understand what you mean 😄 
`can_unlink` is valid enough English for it's use case here - we will not be changing it. Your first commit was good so if you'd revert to that we can merge it 🙂

Justman10000: > Sorry, I don't understand what you mean 😄 `can_unlink` is valid enough English for it's use case here - we will not be changing it.

Dude, can_unlink can be used as a permission name! Here it is the question whether an integration can be unlinked! So: Can be unlink = can_be_unlink

Derkades: "can be unlink" is not valid English. "can be unlinked" would be, but the existing name is also fine. Please keep to one change per PR :)

Justman10000: > "can be unlink" is not valid English. "can be unlinked" would be, but the existing name is also fine. Please keep to one change per PR :)

As I said before

> Dude, can_unlink can be used as a permission name! Here it is the question whether an integration can be unlinked!


And one change per PR? Sorry, but... Stupid? How many PRs do you want? Literally a few hundred thousand?

supercrafter100: > And one change per PR? Sorry, but... Stupid? How many PRs do you want? Literally a few hundred thousand?

Yes, makes it easier to revert them

> As I said before

Your updated term was actually wrong. Though we are not planning on changing the name of the column in the database as this is just simply unnecessary and would be a major breaking change. It is fine the way it is right now. So that's why we're asking to revert that change so we can just fix the redundant language term.

supercrafter100: We're talking about the `adding_server` language term. The `can unlink` changes will not be merged. So that's why we are asking to revert those changes.

supercrafter100: All we're asking is if you can just revert the changes made related to the `can_unlink` string. As previously mentioned the way it is currently worded `can_unlink` is completely valid the way it is used right now. Your updated string is actually not correct english. Additionally, we are not planning on changing that database column. 

So if you can *just* include the original changes you made (eg: the `NEW_SERVER` one) then we can merge this. We will not be merging this PR in it's current state due to the can_unlink change.

Justman10000: Already, yes, I have not changed the title above! The **_Update_** as the title was intentional! And whether you merge the two, or both separately, this major breaking change would be either way! And `can_unlink` is simply wrong, because this sounds like an permission, but not an action!

In German this is called _Kann entlinken_! However, the _can_unlink_ refers to the integration, not to the user

Justman10000: > There won't be a breaking change if we don't do the change, which we aren't planning on doing anyways. can_unlink is completely correct in it's current context. It indicates if the integration can be unlinked (true or false). So I don't see anything wrong with it? It can be used in both contexts.

Exactly not! If we keep it in this context, it sounds as if the integration could unlink other things! An integration can not do this! _can_unlink_ is related to the user, that is, can be unlinked

Justman10000: From me! Fixed

5659cf2

supercrafter100 (dismissed):

Derkades:
It would be nice if permissions were restored from 755 to 644 like padow mentioned

Justman10000:
> It would be nice if permissions were restored from 755 to 644 like padow mentioned

Why?

Justman10000:
@PadowYT2 @Derkades

Better?

Justman10000:
@PadowYT2 @Derkades

Better?

Derkades (approved):

PR #3396, opened 2023-06-16
Create `CacheCollector` to display `Cache` stats in the debug bar

tadhgboyle:
image


core/classes/Core/Cache.php
```diff
@@ -65,14 +65,20 @@ public function isCached(string $key): bool {
...
-                return isset($cachedData[$key]['data']);
             }
         }
 
-        return false;
+        if (!isset($is_cached)) {
+            $is_cached = false;
+        }
+
+        CacheCollector::getInstance()->recordCheck($key, $is_cached);
```
samerton: Can we be sure that, if we call this method, we won't see any errors in production websites (i.e. not using developer dependencies) because `CacheCollector` imports `DebugBar`

samerton (changes requested):

core/classes/Core/Cache.php
tadhgboyle: we shouldn't - the events collector imports debugbar and I've not had issues there!

tadhgboyle: thanks, good call - I added a check to ensure debugbar is loaded!

partydragen (approved):

PR #3398, opened 2023-06-17
Rework user group cache issue

partydragen:
Fixed user group cache does not update when groups change causes duplication errors in database when user being loaded at new after the groups changed


tadhgboyle (approved):
Thanks for fixing this - when we do the cache DX improvement changes this can be made more readable

PR #3410, opened 2023-07-07
Fix forum index showing topics without view other topics permission

samerton:


modules/Forum/classes/Forum.php
```diff
@@ -808,4 +797,47 @@ public static function getAccessibleLabels(array $labels, array $user_groups): a
...
+                WHERE t.forum_id = ?
+                  AND (t.topic_creator = ? OR t.sticky = 1)
+                ORDER BY p.created, p.post_date DESC LIMIT 1
+            SQL,
+            [$forumId, $userId]
+        );
+
+        if ($latest_post->count() && $latest_post = $latest_post->first()) {
+            return [
+                $latest_post->created ?? strtotime($latest_post->post_date),
```
tadhgboyle, line 835: Out of scope for this PR but this definitely feels like code smell

tadhgboyle (approved):

modules/Forum/classes/Forum.php
Derkades: Can I mark this conversation as resolved to be able to merge this PR?

Derkades:
please squash when merging, easier to backport :)

samerton:
> please squash when merging, easier to backport :)

Absolutely - I'll always squash if there is >1 commit in the diff, if cherry-picking there is no difference either way

Derkades:
Squashing is still useful even when a merge request only contains a single commit, because:
- the commit title is changed to the PR title, with a reference to the PR number (this means the backport commit will show up in the PR, as a confirmation that the backport has been done)
- the date of the commit is updated to be the date of the PR merge
- there is no additional merge commit which in my opinion is only clutter in this case

PR #3411, opened 2023-07-07
Small reactions fixes

samerton:
- Fix saving negative reactions in StaffCP
- Fix message element background (used for reactions box in forums) being too bright on dark mode

### DIFF
--- a/core/classes/Core/Validate.php
+++ b/core/classes/Core/Validate.php
@@ -159,7 +159,7 @@ class Validate {
                         if (empty($source[$array][$matches[1]])) {
                             $missing = true;
                         }
-                    } else if (empty($value)) {
+                    } else if (empty($value) && $value !== '0') {
                         $missing = true;
                     }
 
--- a/custom/panel_templates/Default/core/reactions_form.tpl
+++ b/custom/panel_templates/Default/core/reactions_form.tpl
@@ -73,7 +73,7 @@
                                         <select name="type" class="form-control" id="InputReactionType">
                                             <option value="2" {if $TYPE_VALUE eq 2} selected{/if}>{$POSITIVE}</option>
                                             <option value="1" {if $TYPE_VALUE eq 1} selected{/if}>{$NEUTRAL}</option>
-                                            <option value="-1" {if $TYPE_VALUE eq 0} selected{/if}>{$NEGATIVE}</option>
+                                            <option value="0" {if $TYPE_VALUE eq 0} selected{/if}>{$NEGATIVE}</option>
                                             <option value="3" {if $TYPE_VALUE eq 3} selected{/if}>{$CUSTOM_SCORE}</option>
                                         </select>
                                     </div>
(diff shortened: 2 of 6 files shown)


core/classes/Core/Validate.php
```diff
@@ -159,7 +159,7 @@ public static function check(array $source, array $items = []): Validate {
                         if (empty($source[$array][$matches[1]])) {
                             $missing = true;
                         }
-                    } else if (empty($value)) {
+                    } else if (empty($value) && $value !== '0') {
```
samerton, line 162: Wasn't completely sure about this one, let me know what you think

tadhgboyle: Only concern is that this could cause unintended/surprise behaviour elsewhere where it's used...

custom/panel_templates/Default/core/reactions_form.tpl
```diff
@@ -73,7 +73,7 @@
                                         <select name="type" class="form-control" id="InputReactionType">
                                             <option value="2" {if $TYPE_VALUE eq 2} selected{/if}>{$POSITIVE}</option>
                                             <option value="1" {if $TYPE_VALUE eq 1} selected{/if}>{$NEUTRAL}</option>
-                                            <option value="-1" {if $TYPE_VALUE eq 0} selected{/if}>{$NEGATIVE}</option>
+                                            <option value="0" {if $TYPE_VALUE eq 0} selected{/if}>{$NEGATIVE}</option>
```
tadhgboyle, line 76: Nice find, out of curiosity how did `-1` not work here?

samerton: There's a `Validate::IN` with the various different reaction types - negative is 0 so -1 is therefore an invalid value

tadhgboyle: Ohh 🤦🏼 Nice one, don't know how I missed that originally

custom/templates/DefaultRevamp/reactions_modal.tpl
```diff
@@ -1,4 +1,4 @@
-<div class="ui menu" style="overflow: scroll;">
```
tadhgboyle, line 1: Pretty sure this was there to gracefully handle having a few dozen reactions nicely, might be misremembering. You can use the seeder to check though (edit the forum reaction count constant to get lots of reactions)

custom/templates/DefaultRevamp/css/custom.css
```diff
@@ -1486,6 +1501,27 @@ body.dark .cc-window.cc-floating .cc-highlight .cc-btn.cc-allow {
     border-color: transparent;
 }
 
+body.dark #modal-reactions .ui.menu {
+    background-color: #282828;
+    overflow-x: auto;
```
tadhgboyle, line 1506:
```suggestion
    overflow: auto;
```
Sorry, meant reactions as in reaction types but also reactions as in reactions given by users 😅 so both directions need this

samerton: The `scrolling content` class on the outer div should do this already 👍

tadhgboyle (approved):

supercrafter100 (approved):

core/classes/Core/Validate.php
samerton: Should be okay but let's keep an eye out for reports in case we are actually relying on `REQUIRED` for non-'0' values in places

PR #3431, opened 2023-08-18
Add forum description variable

Gamerboy59:
Add a forum description variable that can be used in templates. I did not find a way to add a sub header (this is how fomatic classes are called for this) in DefaulRevamp without breaking the ui but custom templates can make more use of it.


samerton (approved):

samerton:
Thanks!

PR #3433, opened 2023-08-24
Ignore group sync request instead of returning error

Derkades:
When NamelessMC received a group sync request while the Minecraft integration was disabled, or if it received the request from a different server than the selected group sync server, Nameless returned an API error.

I think "invalid server id" is not really an appropriate error for this situation, and simply ignoring the request would be a better user experience.

Example user confusion:
image

### DIFF
--- a/modules/Core/includes/endpoints/UpdateGroupsEndpoint.php
+++ b/modules/Core/includes/endpoints/UpdateGroupsEndpoint.php
@@ -14,7 +14,7 @@ class UpdateGroupsEndpoint extends KeyAuthEndpoint {
 
         $server_id = $_POST['server_id'];
 
-        if (!Settings::get('mc_integration') && $server_id != Settings::get('group_sync_mc_server')) {
+        if (!Settings::get('mc_integration') || $server_id != Settings::get('group_sync_mc_server')) {
             $api->returnArray(['message' => $api->getLanguage()->get('api', 'groups_updates_ignored')]);
             return;
         }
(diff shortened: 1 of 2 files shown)


modules/Core/includes/endpoints/UpdateGroupsEndpoint.php
```diff
@@ -13,25 +13,29 @@ public function execute(Nameless2API $api): void {
         $api->validateParams($_POST, ['server_id', 'player_groups']);
 
         $server_id = $_POST['server_id'];
-        $group_sync_log = [];
 
-        if (Settings::get('mc_integration') && $server_id == Settings::get('group_sync_mc_server')) {
-            $integration = Integrations::getInstance()->getIntegration('Minecraft');
+        if (!Settings::get('mc_integration') || $server_id != Settings::get('group_sync_mc_server')) {
+            $api->returnArray(['message' => $api->getLanguage()->get('api', 'groups_updates_ignored')]);
+            return;
```
tadhgboyle: I *think* we don't need this `return` - `returnArray` should `die`

tadhgboyle (dismissed):

modules/Core/includes/endpoints/UpdateGroupsEndpoint.php
Derkades: A that's why

samerton (approved):

partydragen (approved):

PR #3449, opened 2023-10-01
Add uninstall option to modules

samerton:


partydragen (dismissed):

core/classes/Core/Util.php
```diff
@@ -49,8 +49,13 @@ public static function cyrillicToLatin(string $string): string
...
     {
-        // safety precaution, only allow deleting files in "custom" directory
-        if (!str_contains($directory, 'custom')) {
+        // safety precaution, only allow deleting files in "custom", "modules" or "uploads" directory
+        if (
+            str_contains($directory, 'Core') ||
+            !str_contains($directory, 'custom') ||
+            !str_contains($directory, 'modules') ||
+            !str_contains($directory, 'uploads')
+        ) {
```
supercrafter100, line 60: This does not look right. Won’t this always return false unless the string contains custom, modules AND uploads?

Derkades: Looks fine to me. It aborts if:
 - The path contains "Core" (to prevent deleting modules/Core)
 - OR the path does not contain "custom"
 - OR the path does not contain "modules"
 - OR the path does not contain "uploads"

A resolved absolute path check would be better. Although this is not a security feature so it doesn't matter if it can be bypassed.

partydragen (approved):

PR #3460, opened 2023-11-02
Merging of OAuth + Connections

partydragen:
Merge OAuth into connections instead of 2 seperate linking systems making alot of users confused and to not need to link two times on same provider

Aswell to solve current problems releated with connections using oauth for linking aswell
Also its confusing to setup integration that requires oauth for then you need to setup oauth + integration instead of just doing it all in integration page

Progress:
- [x] Remove OAuth pages
- [x] Make users login/register using the connections data
- [x] Add google integration
- [x] Move OAuth settings to integration setting page instead of squeeze it all into same page and single row
- [x] Add option to allow user to login/register with the oauth (Currently if you add linking with it it also requires login/register)


core/classes/Misc/NamelessOAuth.php
```diff
@@ -46,19 +46,18 @@ public function getProviders(): array
...
     /**
-     * Determine if OAuth is available if at least one provider is setup.
+     * Get or create an instance of a specific provider.
      *
-     * @return bool If any provider is setup
+     * @param string $provider The provider name
+     * @return null|array An array of registered OAuth provider.
      */
-    public function isAvailable(): bool
+    public function getProvider(string $provider): ?array
```
samerton, line 54: @partydragen can you confirm that this is correct please - this was the one merge conflict I had to resolve for this PR

custom/panel_templates/Default/core/integrations_edit.tpl
```diff
@@ -88,6 +88,70 @@
...
+                                                <label for="enable" id="enable"
+                                                       class="custom-control-label">
+                                                    {$REGISTER_LOGIN_WITH_OAUTH}
+                                                </label>
+                                            </div>
+
+                                            <div class="form-group">
+                                                <label for="client-id">{$CLIENT_ID}</label>
+                                                <input type="text" name="client-id" class="form-control"
+                                                       id="client-id" placeholder="Client ID"
```
samerton: Some hardcoded terms in the placeholders here - `Client ID`, `Client Secret`, and then there's no placeholder at all for the client URL below

core/classes/Misc/NamelessOAuth.php
partydragen: I have no idea however it looks to be working

custom/panel_templates/Default/core/integrations_edit.tpl
partydragen: Fixed, Also Client URI is never empty as its set by NamelessMC so no Placeholder needed

samerton (approved):

PR #3468, opened 2023-12-15
Bump recommended PHP version to 8.3

tadhgboyle:
We should be okay re: https://www.php.net/manual/en/migration83.php

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -8,7 +8,7 @@ jobs:
     strategy:
       matrix:
         operating-system: ["ubuntu-latest"]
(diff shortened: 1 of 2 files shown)


Derkades:
Not sure if showing a compatibility warning is a good idea, I think it's perfectly fine for people to use PHP 8.2. For example, if their distribution ships PHP 8.2 (like the current Debian version) they probably shouldn't upgrade to PHP 8.3 via third party packages to get rid of the warning.

tadhgboyle:
Ahhh right right thanks, I forgot we only bump this when it's the automatically installed version on major distros 🤦🏼

PR #3469, opened 2023-12-15
Move core language files into `modules/Core`

tadhgboyle:
File mask here will need to be updated to `modules/Core/language/*.json` https://translate.namelessmc.com/settings/namelessmc/nameless/#files
Additionally, the Monolingual base language file will need to be updated to `modules/Core/language/en_UK.json`


supercrafter100 (approved):

Derkades:
thanks, I like this!

PR #3470, opened 2023-12-15
Prevent directory traversal using language parameter

Derkades:
Fixes a very minor security issue, where in the installer anyone could check whether arbitrary json files exist on the system.

### DIFF
--- a/install.php
+++ b/install.php
@@ -56,6 +56,9 @@ if (Config::exists() && Config::get('core.installed') === true) {
 }
 
 if (isset($_GET['language'])) {
+    if (str_contains($_GET['language'], '/')) {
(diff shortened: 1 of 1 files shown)


supercrafter100 (approved):

samerton:
Should probably set open_basedir in the same was as we do in index.php, didn't realise this was not already there for the installer: <https://github.com/NamelessMC/Nameless/blob/develop/index.php#L76>

tadhgboyle:
> Should probably set open_basedir in the same was as we do in index.php, didn't realise this was not already there for the installer: https://github.com/NamelessMC/Nameless/blob/develop/index.php#L76

Can this be done in this PR or should a new one be made?

PR #3484, opened 2024-01-07
Add ability to view users by group

partydragen:
Add ability to view users by group, integration, banned and active

Also allows view users by  integration, banned and active by adding the tags in url manually


custom/panel_templates/Default/template.php
```diff
@@ -304,6 +304,23 @@ public function onPageLoad() {
                                 AssetTree::DATATABLES,
                             ]);
 
+                            $url_parameters = [];
+                            if (isset($_GET['group'])) {
+                                $url_parameters[] = 'group=' . $_GET['group'];
```
samerton: Should these $_GET params all be sanitised in case somebody sends a malicious link to a staff member

samerton (approved):

PR #3488, opened 2024-03-06
Add ability to ban terms in forum content

samerton:
New section added to forum settings to add terms banned from forum content - useful as an anti spam tool


supercrafter100 (approved):

tadhgboyle (approved):
Textarea is unideal but not sure we have anything better accessible with our UI packages 😦

PR #3490, opened 2024-03-08
Strip HTML tags from page description

samerton:


partydragen (dismissed):

partydragen (approved):

Derkades:
Why `addslashes`? Isn't that function specifically for PHP strings? I think `htmlspecialchars()` is the right function to use here.

samerton:
> Why `addslashes`? Isn't that function specifically for PHP strings? I think `htmlspecialchars()` is the right function to use here.

This is just the page description so it shouldn't have any HTML inside - we have already stripped HTML tags, now we need to ensure that quotation marks `"` are also escaped so that they do not interfere with the meta tags

Derkades:
Still, addslashes is for PHP which means it for example escapes `"` to `\"` while htmlspecialchars (with the default flags including ENT_QUOTES) is meant for HTML which means it correctly escapes `"` to `&quot;`. addslashes might coincidentally seem to work fine but it probably has subtle security holes. It's like using urlencode to escape for a SQL query. The addslashes documentation (and comments) warn about this.

PR #3493, opened 2024-03-09
Allow custom endpoint params

partydragen:
Add ability to add custom params to execute function

Example my OAuth2 module have a AccessTokenAuthEndpoint that allow Access Token to access its api
However the Endpoints itself don't know who user authed as im not able to transfer that data to endpoint, This will allow that

As now my AccessTokenAuthEndpoint auth method can add User into CustomParams so now the endpoint can restrict api usage for that user ONLY

Usage by my OAuth2 Module: https://github.com/partydragen/Nameless-OAuth2/commit/ee01af45c2dbbdc7fecad6d648136ed464d8b9b7


tadhgboyle (commented):
I'm confused what functionality this provides, in your example commit the `customParams` is never read from/called. Can you perhaps explain a different way?

partydragen:
> I'm confused what functionality this provides, in your example commit the `customParams` is never read from/called. Can you perhaps explain a different way?

Its being added to execute function
https://github.com/partydragen/Nameless-OAuth2/commit/ee01af45c2dbbdc7fecad6d648136ed464d8b9b7#diff-c64a5bfe66e837e1b28172db725aaccd0a4bbeb0dbbaddc8abee369cb0570196R11

`public function execute(Nameless2API $api, AccessToken $token): void {`
So it being `execute(API, CUSTOM AUTH PARAMS, Transfomers)`

Also this this OAuth endpoint that adding AccessToken https://github.com/partydragen/Nameless-OAuth2/commit/158fd22875e0f691616068313da0a6cddd3f8443

So the endpoint know who user authoriced to limit the api data from that user only

Pretty much that Auth methods being able to register params to endpoints instead of just Transformers

And the actuall call happning here https://github.com/NamelessMC/Nameless/pull/3493/commits/f1aec3c3f73ce06799a195241457a77efb5cfd74#diff-52869f0ee82251d95009587780c0e900fa379fa99957f42b97d13cde939b05f3R72 thats add it to execute

samerton (approved):

PR #3504, opened 2024-05-02
Add UserUpdatedEvent

partydragen:
Added UserUpdatedEvent allows modules to eaiser detect user changes like when username change

Will need to discover possible places to execute event

### DIFF
--- a/modules/Core/classes/Events/UserUpdatedEvent.php
+++ b/modules/Core/classes/Events/UserUpdatedEvent.php
@@ -1,0 +1,30 @@
+<?php
+
+class UserUpdatedEvent extends AbstractEvent implements HasWebhookParams {
+
+    public User $user;
+    public UserData $updated_data;
+
+    public function __construct(User $user) {
+        $this->user = $user;
+
+        $data = DB::getInstance()->get('users', ['id', $user->data()->id]);
+        $this->updated_data = new UserData($data->first());
+    }
+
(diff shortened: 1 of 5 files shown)


modules/Core/classes/Events/UserUpdatedEvent.php
```diff
@@ -0,0 +1,30 @@
...
+class UserUpdatedEvent extends AbstractEvent implements HasWebhookParams {
+
+    public User $user;
+    public UserData $updated_data;
+
+    public function __construct(User $user) {
+        $this->user = $user;
+
+        $data = DB::getInstance()->get('users', ['id', $user->data()->id]);
+        $this->updated_data = new UserData($data->first());
```
tadhgboyle, line 12: Would be nice if we could provide something like `data_before`, `data_after` and then a diff of them via `data_changed`

partydragen: Well this PR did not really go as planned

I want to trigger certain functions when event is triggered but as the functions require User param it will get the outdated data :/ 

Like to solve that i need to directly update userData on update
```
foreach ($fields as $key => $value) {
    $this->data()->{$key} = $value;
}
```
But updating the userdata directly could have bad side effects idk

Another problem i also have is integrations cache is outdated if any integrations is added/removed

PR #3509, opened 2024-06-07
Fix Discord webhook entity decoding

### LINKED ISSUE #3486: Discord webhook | Changing í to &iacute; etc...
### Describe the issue

Hi there,
I'm having this bug with discord webhook.

When I post something on forum, that contains characters like í, ý, á, é etc. It change into tags like &iacute; 
So forum post like this:
image

Looks on discord webhook like this:
image

### NamelessMC Version

2.1.0

### To Reproduce

_No response_

### Expected Behaviour

_No response_

### Screenshots

_No response_

### Additional Information

_No response_

samerton:
Also fix topicReply enabled webhooks not showing in StaffCP forum setup

Closes #3486

### DIFF
--- a/core/classes/Misc/Text.php
+++ b/core/classes/Misc/Text.php
@@ -175,6 +175,8 @@ class Text
             $content
         ));
 
+        $content = html_entity_decode($content);
+
         return self::truncate($content, 512, [
             'html' => true,
         ]);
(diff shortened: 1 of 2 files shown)


tadhgboyle (approved):

Derkades:
As discussed in Discord, it might be a good idea to change this function to strip HTML tags as well, as Discord does not expect HTML content.

samerton:
> As discussed in Discord, it might be a good idea to change this function to strip HTML tags as well, as Discord does not expect HTML content.

Fortunately this is done just before we perform the decoding, leaving us with just special characters at this point: https://github.com/NamelessMC/Nameless/blob/fb1bc1f813f0bfb0099630113c42963394b03b0d/core/classes/Misc/Text.php#L172

PR #3515, opened 2024-06-12
Check if members module is enabled

partydragen:
Check if members module is enabled, Uninstalling the module breaks the site

### DIFF
--- a/modules/Core/module.php
+++ b/modules/Core/module.php
@@ -548,14 +548,16 @@ class Core_Module extends Module {
         Email::addPlaceholder('[Message]', static fn(Language $viewing_language, string $email) => $viewing_language->get('emails', $email . '_message'));
         Email::addPlaceholder('[Thanks]', static fn(Language $viewing_language) => $viewing_language->get('emails', 'thanks'));
 
-        MemberListManager::getInstance()->registerListProvider(new RegisteredMembersListProvider($language));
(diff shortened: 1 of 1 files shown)


tadhgboyle (approved):

modules/Core/module.php
```diff
@@ -548,14 +548,16 @@ public function __construct(Language $language, Pages $pages, User $user, Naviga
...
-        MemberListManager::getInstance()->registerListProvider(new RegisteredMembersListProvider($language));
-        MemberListManager::getInstance()->registerListProvider(new StaffMembersListProvider($language));
-
-        MemberListManager::getInstance()->registerMemberMetadataProvider(function (User $member) use ($language) {
-            return [
-                $language->get('general', 'joined') => date(DATE_FORMAT, $member->data()->joined),
-            ];
-        });
+        if (Util::isModuleEnabled('Members')) {
+            MemberListManager::getInstance()->registerListProvider(new RegisteredMembersListProvider($language));
```
samerton, line 552: Should these providers be registered within the Members module itself?

tadhgboyle: No, the idea here is that they belong to the Core module, the member list module doesn't have any lists of it's own. Similar to how we have `forum post count` lists that are initialized in the forum module

samerton: The reason I ask is because it feels a bit strange to have logic depending on other modules in the core. If we are definitely happy with this approach then I'm okay with it.

PR #3517, opened 2024-06-13
Allow store module bypass maintenance

partydragen:
Allow Store module listener to bypass maintenance mode

Temporary fix until we get a better system to whitelist pages to bypass maintenance, Many people testing their store in maintenance mode but the listeners don't work when its in maintenance mode


tadhgboyle (approved):
Definitely not an ideal state but ✅ for now if this helps you

PR #3519, opened 2024-06-20
Add Persian (Iran) langauge

Derkades:
was requested via weblate


core/classes/Core/Language.php
```diff
@@ -44,6 +44,10 @@ class Language
             'name' => 'English US',
             'htmlCode' => 'en',
         ],
+        'fa_IR' => [
+            'name' => 'Persian (Iran)',
+            'htmlCode' => 'fa',
```
samerton, line 49: Could you please add `'rtl' => true,` here?

samerton (approved):

PR #3522, opened 2024-06-28
Fix array for not start with validation

partydragen:
The foreach being breaked after the first check, It should only break when result is found


tadhgboyle (approved):

tadhgboyle:
I think this would be fine to leave or keep, it depends if we want to show all the errors at once (current behaviour?) or one at a time. The former seems better

partydragen:
> I think this would be fine to leave or keep, it depends if we want to show all the errors at once (current behaviour?) or one at a time. The former seems better

What? The issue here is other denied values is never being checked, Only the first one as it break the loop instantly
Making them bypass all other denied values

This will still only show 1 error at a time, Same as all other array releated checks

PR #3524, opened 2024-07-14
Record server ID when creating report via API and display in staff panel

### LINKED ISSUE #3190: Extend report API with field for Minecraft server id
### Describe the feature

With the implementation of the NamelessMC plugins in-game, using /report has become really useful for universal reporting across websites and in-game. I would love to see the ability to see the server where reports are coming from. This is really useful information.

samerton:
Closes #3190 

Screenshot 2024-07-14 at 12 27 01
Screenshot 2024-07-14 at 12 27 42

### DIFF
--- a/modules/Core/includes/endpoints/CreateReportEndpoint.php
+++ b/modules/Core/includes/endpoints/CreateReportEndpoint.php
@@ -6,6 +6,7 @@
  * @param string $content The content of the report
  * @param string $reported_username The username of the reported user (optional, required if reported not provided)
  * @param string $reported_uid A unique ID for the reported user (optional, required if reported not provided)
+ * @param string $server_id A server ID the report is associated with
  *
  * @return string JSON Array
  */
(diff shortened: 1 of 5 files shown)


partydragen (approved):

modules/Core/includes/endpoints/CreateReportEndpoint.php
```diff
@@ -79,8 +80,9 @@ public function execute(Nameless2API $api): void {
             'reported_id' => $user_reported_id,
             'report_reason' => $_POST['content'],
             'updated_by' => $user_reporting_data->id,
-            'reported_mcname' => $_POST['reported_username'] ? $_POST['reported_username'] : $reported_user->getDisplayname(),
-            'reported_uuid' => $_POST['reported_uid'] ? $_POST['reported_uid'] : $reported_uuid ?? 'none',
+            'reported_mcname' => $_POST['reported_username'] ?: $reported_user->getDisplayname(),
+            'reported_uuid' => $_POST['reported_uid'] ?: $reported_uuid ?? 'none',
+            'server_id' => $_POST['server_id'],
```
Derkades, line 85: The endpoint should probably check if the server id is valid, if one is specified. Of course, not specifying one at all should be allowed for backwards compatibility

modules/Core/pages/panel/users_reports.php
```diff
@@ -89,36 +89,52 @@
...
+                $reported_user_profile = URL::build('/panel/user/' . urlencode($report->reported_id . '-' . $reported_user->data()->username));
+                $reported_user_style = $reported_user->getGroupStyle();
+                $reported_user_avatar = $reported_user->getAvatar();
             }
 
             $updated_by_user = new User($report->updated_by);
 
+            $server = null;
+            if ($report->server_id) {
+                $server = Output::getClean(DB::getInstance()->get('mc_servers', ['id', $report->server_id])->first()->name);
```
Derkades, line 127: What if the server does not exist or no longer exists?

samerton: Good point, will add a null check in here

modules/Core/includes/endpoints/CreateReportEndpoint.php
samerton: I can add a check in to prevent a string for example to avoid a MySQL error

Derkades:
I forgot to say - this is a useful addition and I've implemented it into the plugin just now

tadhgboyle (approved):

PR #3536, opened 2024-08-11
Don't hardcode oauth scopes

partydragen:
Don't hardcode oauth scopes, Such as email is forced for all but integrations such as twitch does not have that scope name when trying to auth it returns invalid scope

+ ability to add several scopes


tadhgboyle (approved):
An array merge here might be better instead of making the provider manually add back the email and ID scope

partydragen:
> An array merge here might be better instead of making the provider manually add back the email and ID scope

We can't merge as some providers don't have the email scope, that causes the it to return invalid scope error on their end

Also when scopes is null, the league library will use the providers default scopes (Not currently will current backwards compability)

PR #3538, opened 2024-09-12
New group sync endpoints with separate add/remove lists

Derkades:
The current group sync system is based on an endpoint where the source (minecraft, discord) submits a list of current groups. I believe this is fundamentally flawed. The easiest to understand example of this is the first sync. What currently happens is, either Discord sends groups to the website or the website sends groups to Discord, whichever is first. One of the two sides decides, on the other side all previously set roles are lost.

See for example https://github.com/NamelessMC/Nameless-Link/issues/449, but this issue has also been brought up in Discord.

The website cannot distinguish between these two cases:
 - A group is not present in the request because it was removed -> The group should be removed
 - A group is not present in the request because it was always missing -> The group should remain on the website, and in fact be added in the opposite direction.

This can, I believe, be solved by an endpoint with separate `add` and `remove` lists. On the initial sync, both sides send all their current groups in the `add` list to the other side. Afterwards, every time a group is added or removed, the Minecraft server or Discord bot makes a request with the added or removed groups.

These endpoints were tested manually using curl, but not yet with the plugin or nameless-link. I first wanted to see if anyone had good ideas and feedback.

Disclosure: I am being paid to fix this group sync issue in Nameless-Link, which required me to add these new endpoints. It would be unfair to demand someone to review this PR for free, so I wil provide compensation to members of the Nameless team.

### DIFF
--- a/core/classes/Group_Sync/GroupSyncManager.php
+++ b/core/classes/Group_Sync/GroupSyncManager.php
@@ -153,6 +153,10 @@ final class GroupSyncManager extends Instanceable
      * @param string $sending_injector_class Class name of injector broadcasting this change
      * @param array  $group_ids              Array of Group IDs native to the sending injector which were added/removed to the user
      *
+     * @deprecated broadcastGroupChange should be used instead. This function has issues, such as that on the first sync it will replace
+     *             groups on one side with groups from the other side, depending on which side happens to sync first. On the first sync,
+     *             the desired behaviour is for the new roles on both sides to become the union of roles on both sides.
+     *
      * @return array Array of logs of changed groups
      */
     public function broadcastChange(User $user, string $sending_injector_class, array $group_ids): array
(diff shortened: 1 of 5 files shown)


modules/Discord Integration/includes/endpoints/SyncDiscordRolesEndpoint.php
```diff
@@ -0,0 +1,40 @@
+<?php
+
+/**
+ * @param int $user The NamelessMC user ID to edit
+ * @param string $roles An array of Discord Role ID to give to the user
+ *
+ * @return string JSON Array
+ */
```
tadhgboyle:
```suggestion
```

modules/Core/includes/endpoints/UpdateGroupsEndpoint.php
```diff
@@ -1,5 +1,8 @@
 <?php
 
+/**
+ * @deprecated SyncMinecraftGroupsEndpoint should be used instead
+ */
```
tadhgboyle, line 5: Sidebar, never considered that API endpoints could be deprecated. Wouldn't be too hard to add something like `extends DeprecatedEndpoint` and make Nameless log every time a deprecated endpoint is hit for visibility

modules/Core/includes/endpoints/SyncMinecraftGroupsEndpoint.php
```diff
@@ -0,0 +1,38 @@
...
+
+    public function __construct() {
+        $this->_route = 'minecraft/sync-groups';
+        $this->_module = 'Core';
+        $this->_description = 'Update a users groups based on added or removed groups from the Minecraft server';
+        $this->_method = 'POST';
+    }
+
+    public function execute(Nameless2API $api): void {
+        $api->validateParams($_POST, ['server_id', 'user']);
```
tadhgboyle:
```suggestion
    public function __construct() {
        $this->_route = 'minecraft/{user}/sync-groups';
        $this->_module = 'Core';
        $this->_description = 'Update a users groups based on added or removed groups from the Minecraft server';
        $this->_method = 'POST';
    }

    public function execute(Nameless2API $api, User $user): void {
        $api->validateParams($_POST, ['server_id']);
```

Cant also then remove line 22

core/classes/Group_Sync/GroupSyncManager.php
```diff
@@ -283,6 +287,129 @@ public function broadcastChange(User $user, string $sending_injector_class, arra
...
+            $add = $data['add'];
+            $remove = $data['remove'];
+
+            /** @var GroupSyncInjector&BatchableGroupSyncInjector $injector */
+            $injector = $this->getInjectorByClass($injector_class);
+            $injector_column = $injector->getColumnName();
+
+            if ($injector instanceof BatchableGroupSyncInjector) {
+                /** @var GroupSyncInjector&BatchableGroupSyncInjector $injector */
+                $batchable_injector = $injector;
```
tadhgboyle: Unused assignment?

tadhgboyle (dismissed):

modules/Core/includes/endpoints/SyncMinecraftGroupsEndpoint.php
Derkades: Good idea, I'll also apply this to the discord endpoint

core/classes/Group_Sync/GroupSyncManager.php
Derkades: Right, I was rewriting this code but forgot to finish it

modules/Discord Integration/includes/endpoints/SyncDiscordRolesEndpoint.php
Derkades: It is a bit strange, but other endpoints have this documentation too. In any case, I forgot to update it so it has to go.

Derkades:
Thanks for the review. It has now also been implemented in [Nameless-Java-API](https://github.com/NamelessMC/Nameless-Java-API/commit/e0e155c087dd1d35c488156a5921ac5e1589dd05) and [Nameless-Link](https://github.com/NamelessMC/Nameless-Link/commit/e628dc98cbd8460a3512c9274ee587c811445989), with a fallback to the previous endpoint. Tested with a local instance of the Nameless website and Nameless-Link bot, so it should be safe to merge

modules/Core/includes/endpoints/UpdateGroupsEndpoint.php
Derkades: It might be nice. I do worry that it will result in too much log spam

supercrafter100 (approved):

tadhgboyle (approved):

tadhgboyle (approved):

Issue #3480, opened 2023-12-24
Determine the right group for chown


PadowYT2:
### Describe the feature

This is based on 
https://github.com/NamelessMC/Nameless/blob/fcaa51708c23d592a92209d9ff7728f557070647/core/init.php#L43
basically, on CentOS, it is either nginx/apache and not www-data. Not sure if possible, but PHP should determine the right group

Derkades:
get_current_user() should work

samerton:
According to the docs, `get_current_user()` returns the owner of the current file as opposed to the user running the PHP process: https://www.php.net/manual/en/function.get-current-user.php

PR #3547, opened 2024-10-18
Revert is_writable should be negative

partydragen:
Currently the error is shown when the folders are actually writeable!

Revert the change from https://github.com/NamelessMC/Nameless/pull/3534 https://github.com/NamelessMC/Nameless/issues/3480


tadhgboyle (approved):

Derkades:
Oops, must have left this in while debugging

PR #3557, opened 2024-12-01
Permissions for posting on profiles

### LINKED ISSUE #3121: Way to limit new user ability to wall post
### Describe the feature

I approve people to play on my server and use my website, but the way that NameslessMC works right now, anyone can register and immediately post to their wall or private message anybody. I've seen a few trolls take advantage of this.

I propose 2 features to help with this:

1. An option to restrict access to wall posts and private messaging to new users until approved, or given a "group" which has access to those things.

2. Sections in the Staff Panel where selected staff can monitor the activity for any inappropriate, illegal, troll, etc. behaviour.

samerton:
Closes #3121 

Also tidies reactions CSS (overflows when there are a lot of reactions, dark mode reaction count colour)


supercrafter100 (commented):
Seeing as we added a new permission, don't we have to add it to the admin group by default? Both here https://github.com/NamelessMC/Nameless/blob/31fc854d4a76925e540ffca3a4eaedc649b6d2a5/core/classes/Database/DatabaseInitialiser.php#L49 and in a migration step?

supercrafter100:
Im just not sure if this is going to break posting for existing websites when they upgrade. Or should this just be handled in the upgrade script when we create that?

samerton:
> Seeing as we added a new permission, don't we have to add it to the admin group by default? Both here
> 
> https://github.com/NamelessMC/Nameless/blob/31fc854d4a76925e540ffca3a4eaedc649b6d2a5/core/classes/Database/DatabaseInitialiser.php#L49
> 
> and in a migration step?

Admin group gets all permissions by default so it's not needed to add them in individually anymore - I will update this file anyway (mostly for the benefit of member/moderator).

As for the migration script - I was thinking website owners could go in and manually assign the permission to their groups as we will not know how different websites have their groups set up.

tadhgboyle (approved):

PR #3560, opened 2024-12-01
Fix being unable to centre images via editor

### LINKED ISSUE #3553: Image Centering Bug
### Describe the issue

When making a Forum post, sometimes images will not stay centered after posting. Going back into edit and centering the images does not fix the issue. Only some images are affected, and only some of the time. 

### NamelessMC Version

2.1.0

### To Reproduce

The bug is hard to reproduce, as it seems random.
When it does appear, these are the steps;

1. Go into edit mode on a forum post
2. Copy paste, or upload an image
3. Center the image
4. Post the forum post

The picture will sometimes revert back to not being centered. 

### Expected Behaviour

The expected behavior is for images to stay centered after posting.

### Screenshots

_No response_

### Additional Information

Contact me on Discord at BDanecker if you need more information.
One Dev suggested that this bug can be worked around by centering the text first and then inserting the image.

samerton:
Margin properties were missing from allowed CSS styles via purifier

Closes #3553

### DIFF
--- a/core/classes/Core/Output.php
+++ b/core/classes/Core/Output.php
@@ -55,10 +55,10 @@ class Output
             $purifierConfig->set('HTML.Doctype', 'XHTML 1.0 Transitional');
             $purifierConfig->set('URI.DisableExternalResources', false);
             $purifierConfig->set('URI.DisableResources', false);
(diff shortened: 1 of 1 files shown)


supercrafter100:
would it be useful to include padding as well? As for future proofing?

samerton:
> would it be useful to include padding as well? As for future proofing?

I'm not sure if padding is used by the editor but I have added that in along with some other updates 👍

supercrafter100 (approved):

partydragen (approved):

PR #3570, opened 2024-12-14
Cache 2fa code for 1min when setting it up

### LINKED ISSUE #3495: Force Two Factor Authentication for group members.
### Describe the issue

I turned on Force Two Factor Authentication for group members.
But when entering the scan codes we are getting Invalid code, please try again.

### NamelessMC Version

Development version

### To Reproduce

1. Go to Groups.
2. Click on a group.
3. turn on Force Two Factor Authentication for group members
4. Scan QR code.
5. See error:  Invalid code, please try again.

Also when i just got to my account and enable 2fa from there it works!
But when Forcing Two Factor Authentication for group members you get the Invalid code, please try again.

### Expected Behaviour

Enable 2fa for the accounts.

### Screenshots

_No response_

### Additional Information

PHP 8.1.27 Updated to  PHP 8.3.3
NamelessMC Version: 2.1.2

2fa:
Okta Verify.
microsoft authenticator

supercrafter100:
While this should normally not be needed, there is currently an issue when force 2fa is enabled for a group that causes every request to be redirected to this page. A side effect of this is that every page load, the 2fa code gets regenerated and thus is no longer valid when the user enters their code. To address this, I cache the code for a minute to make sure that they have managed to enter the code and no query has overridden it. It is not a perfect solution but should at least help fix the issue for the time being.

### DIFF
--- a/modules/Core/pages/user/settings.php
+++ b/modules/Core/pages/user/settings.php
@@ -34,12 +34,17 @@ if (isset($_GET['do'])) {
         $tfa = new \RobThree\Auth\TwoFactorAuth(Output::getClean(SITE_NAME));
 
         if (!isset($_GET['s'])) {
-            // Generate secret
-            $secret = $tfa->createSecret();
 
-            $user->update([
-                'tfa_secret' => $secret
-            ]);
+            // Generate secret
+            $cache->setCache("users_tfa");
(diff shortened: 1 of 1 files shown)


supercrafter100:
This should hopefully close #3495

samerton (approved):

PR #3572, opened 2024-12-29
Add Put & Patch & Delete to HTTPClient

partydragen:


tadhgboyle (approved):
Approved, but we should really refactor this class so it’s not a bunch of duplicated methods with only 1 or 2 lines of difference

PR #3579, opened 2025-01-30
Increment array support

partydragen:
Make increment and decrement support several columns and custom number

### DIFF
--- a/core/classes/Database/DB.php
+++ b/core/classes/Database/DB.php
@@ -386,33 +386,67 @@ class DB
     }
 
     /**
-     * Increment a numeric column value by 1.
+     * Increment a numeric column value by 1 or array of data in "column => value" format to increment.
      *
-     * @param  string $table The table to use.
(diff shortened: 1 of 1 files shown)


core/classes/Database/DB.php
```diff
@@ -386,33 +386,67 @@ public function update(string $table, $where, array $fields): bool
...
-     * Increment a numeric column value by 1.
+     * Increment a numeric column value by 1 or array of data in "column => value" format to increment.
      *
-     * @param  string $table The table to use.
-     * @param  int    $id    The id of the row to increment a column in.
-     * @param  string $field The field to increment.
-     * @return bool   Whether an error occurred or not.
+     * @param  string       $table  The table to use.
+     * @param  int          $id     The id of the row to increment a column in.
+     * @param  string|array $fields Array of data to in "column => value" format to increment.
```
samerton, line 393: Perhaps, if `$fields` is in the format `['column_one', 'column_two']`, then we could default the `value` to 1?

PR #3610, opened 2025-04-19
Fix type error for widgets on newer php versions

Gamerboy59:
Property $page is typed as array but json_decode returns an object not an array which throws an error in recent php version (8 and newer). Setting the second parameter to true makes it return an associated array instead of an object to match the type declaration of the property.
Fixes: `Uncaught TypeError: Cannot assign stdClass to property WidgetData::$pages of type array`

### DIFF
--- a/core/classes/Widgets/WidgetData.php
+++ b/core/classes/Widgets/WidgetData.php
@@ -14,6 +14,6 @@ class WidgetData
             ? []
             : (is_array($data->pages)
                 ? $data->pages
-                : json_decode($data->pages));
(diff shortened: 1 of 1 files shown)


tadhgboyle (approved):
Thank you!

partydragen:
Older sites stored widges with no pages selected as {}, in latest version it changed to array yea []

Still working on confirming it as some of my sites did not break when i used {} as value

partydragen (approved):

partydragen:
Thank you :)

PR #3613, opened 2025-04-21
Bump minimum PHP version to 8.2

tadhgboyle:


index.php
```diff
@@ -25,9 +25,9 @@
     error_reporting(-1);
 }
 
-// Ensure PHP version >= 7.4
-if (PHP_VERSION_ID < 70400) {
-    die('NamelessMC is not compatible with PHP versions older than 7.4, you are running PHP ' . PHP_VERSION);
+// Ensure PHP version >= 8.2
```
supercrafter100, line 28: Do we really want to kill every site that still uses 7.4? A warning that stuff Might break is probably less intrusive than immediately nuking okder sites? Regardless, how will this work with the updater?

tadhgboyle: Good questions, as we will be using new language features, sites will break anyways on old PHP versions - even without this early explosion so I believe it’s alright to keep it here. 
In terms of the upgrade, we can put out a minor release that lets us check a PHP version requirement from an upgrade script before running, perhaps?

supercrafter100: yeah that sounds like a better idea, don't want the site to just suddenly break when updating. Perhaps we can just have a warning on the update page itself so we can specify a required version through the released module and it shows a warning if you don't meet the requirement?

tadhgboyle: Yeah great idea. We can update the releases API to display the required PHP version, then we can display a nice big red error if we detect a mismatch locally

supercrafter100 (approved):

PR #3620, opened 2025-04-25
Convert Reports to Notification system

tadhgboyle:


core/classes/Locale/Language.php
```diff
@@ -240,11 +240,11 @@ public function __construct(string $module = 'core', ?string $active_language =
      * Return a term in the currently active language.
      *
      * @param  string  $section   Section name.
-     * @param  ?string $term      The term to translate.
+     * @param  string  $term      The term to translate.
      * @param  array   $variables Any variables to pass through to the translation.
      * @return string  Translated phrase.
      */
-    public function get(string $section, ?string $term = null, array $variables = []): string
+    public function get(string $section, string $term, array $variables = []): string
```
tadhgboyle: couldn't see a case where we don't pass anything to term 🤷🏻

core/classes/Locale/LanguageKey.php
```diff
@@ -0,0 +1,37 @@
...
+    private string $section;
+    private string $term;
+    private string $module;
+    private array $variables;
+
+    public function __construct(string $section, string $term, array $variables = [], ?string $module = 'core')
+    {
+        $language = new Language($module, 'en_UK');
+        $translation = $language->get($section, $term);
+        if ($translation === $section . '/' . $term) {
```
tadhgboyle, line 22: have to do this as we don't offer a `Language#exists` method atm

core/classes/Misc/Report.php
```diff
@@ -35,25 +35,18 @@ public static function create(Language $language, User $user_reporting, User $re
...
-
-            $moderators = DB::getInstance()->query('SELECT DISTINCT(nl2_users.id) AS id FROM nl2_users LEFT JOIN nl2_users_groups ON nl2_users.id = nl2_users_groups.user_id WHERE group_id in ' . $groups)->results();
-
-            if (count($moderators)) {
-                foreach ($moderators as $moderator) {
-                    Alert::create($moderator->id, 'report', ['path' => 'core', 'file' => 'moderator', 'term' => 'report_alert'], ['path' => 'core', 'file' => 'moderator', 'term' => 'report_alert'], URL::build('/panel/users/reports/', 'id=' . $id));
-                }
-            }
-        }
+        $moderators = DB::getInstance()->query('SELECT DISTINCT(nl2_users.id) AS id FROM nl2_users LEFT JOIN nl2_users_groups ON nl2_users.id = nl2_users_groups.user_id WHERE group_id IN (SELECT id FROM nl2_groups WHERE permissions LIKE \'%"modcp.reports":1%\')')->results();
```
tadhgboyle, line 38: Remove need for 2nd query

core/classes/Locale/LanguageKey.php
```diff
@@ -0,0 +1,37 @@
...
+class LanguageKey
+{
+    private string $section;
+    private string $term;
+    private string $module;
+    private array $variables;
+
+    public function __construct(string $section, string $term, array $variables = [], ?string $module = 'core')
+    {
+        $language = new Language($module, 'en_UK');
```
partydragen, line 20: So this class is only intended for Notification class as you force english here? As i see the Notification class doing the part of getting the term for user language

modules/Core/classes/Core/Notification.php
```diff
@@ -76,25 +104,31 @@ public function send(): void {
...
         }
     }
 
-    private function sendAlert(int $userId, string $content): void {
-        Alert::send($userId, $this->_title, $content, null, $this->_skipPurify);
+    private function sendAlert(int $userId, string $title, ?string $content): void {
+        Alert::send(
+            $userId,
+            $title,
+            $this->_alertUrl ? null : $content,
```
partydragen, line 116: Why should content be null when alert url is set?

tadhgboyle: AFAIK, alerts will ignore content if the url/link is set, so no need to set content if we set the URL instead

core/classes/Locale/LanguageKey.php
tadhgboyle: Yes this is to ensure that the term passed to `LanguageKey` exists in at least the base language as a sort of typo check

supercrafter100 (dismissed):

supercrafter100 (approved):

PR #3621, opened 2025-04-25
Drop `nl2_users.warning_points` column from users table

tadhgboyle:
Unused internally and, afaik, not written to in many years

### DIFF
--- a/core/init.php
+++ b/core/init.php
@@ -119,7 +119,7 @@ if ($page != 'install') {
     }
 
     // Ensure database is up-to-date
-//    PhinxAdapter::ensureUpToDate('Core');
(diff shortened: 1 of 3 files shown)


core/init.php
```diff
@@ -119,7 +119,7 @@
     }
 
     // Ensure database is up-to-date
-//    PhinxAdapter::ensureUpToDate('Core');
+    PhinxAdapter::ensureUpToDate('Core');
```
tadhgboyle, line 122: @samerton was this intentional? 😄

arichardson02:
LGTM

core/init.php
samerton: Nice catch 😅

PR #3622, opened 2025-04-25
Convert Forum topic replies to Notification system

tadhgboyle:


modules/Forum/module.php
```diff
@@ -167,6 +167,13 @@ public function __construct(Language $language, Language $forum_language, Pages
         }
 
         ReactionContextsManager::getInstance()->provideContext(new ForumPostReactionContext($forum_language));
+
+        Notification::addType(
+            'forum_topic_reply',
+            $forum_language->get('forum', 'forum_topic_replies'),
+            Module::getIdFromName('Forum'),
+            ['alert' => false, 'email' => true],
```
partydragen: Alert should also be enabled by default thought like current version atleast since users must click follow topic themself

modules/Forum/pages/forum/view_topic.php
```diff
@@ -334,64 +334,41 @@
                     $available_hooks,
                 ));
 
-                // Alerts + Emails
-                $users_following = DB::getInstance()->get('topics_following', ['topic_id', $tid])->results();
-                if (count($users_following)) {
-                    $users_following_info = [];
-                    foreach ($users_following as $user_following) {
-                        if ($user_following->user_id != $user->data()->id) {
-                            if ($user_following->existing_alerts == 0) {
```
partydragen, line 343: You're no longer checking if user has already been notified of new activity since last visit

Can be very spammy if user get email/alert for every single new post

partydragen (changes requested):

modules/Forum/pages/forum/view_topic.php
tadhgboyle: resolved, added to where clause for fetching notification recipients

modules/Forum/pages/forum/view_topic.php
```diff
@@ -334,64 +334,41 @@
...
-                                Alert::create(
-                                    $user_following->user_id,
-                                    'new_reply',
-                                    ['path' => ROOT_PATH . '/modules/Forum/language', 'file' => 'forum', 'term' => 'new_reply_in_topic', 'replace' => ['{{author}}', '{{topic}}'], 'replace_with' => [Output::getClean($user->data()->nickname), Output::getClean($topic->topic_title)]],
-                                    ['path' => ROOT_PATH . '/modules/Forum/language', 'file' => 'forum', 'term' => 'new_reply_in_topic', 'replace' => ['{{author}}', '{{topic}}'], 'replace_with' => [Output::getClean($user->data()->nickname), Output::getClean($topic->topic_title)]],
-                                    URL::build('/forum/topic/' . urlencode($tid) . '-' . $forum->titleToURL($topic->topic_title), 'pid=' . $last_post_id)
-                                );
-                                DB::getInstance()->update('topics_following', $user_following->id, [
-                                    'existing_alerts' => 1
```
partydragen, line 352: Then you also need this back as of now its always false so user will still get spammed with notifications

samerton (dismissed):

partydragen (changes requested):

modules/Forum/pages/forum/view_topic.php
tadhgboyle: added, thanks!

partydragen (dismissed):

partydragen (approved):

PR #3628, opened 2025-04-27
Convert Mentions to Notification system

tadhgboyle:
wip


core/classes/Database/DB.php
```diff
@@ -520,10 +520,10 @@ public function addColumn(string $table, string $column, string $attributes): bo
      *                        column, operator (default =), value, and glue (default AND).
      * @return array The where clause string, and parameters to bind.
      */
-    public static function makeWhere(array $clauses): array
+    private function makeWhere(array $clauses): array
```
partydragen: It should stay public it has nice use cases, And sam uses it in the new resources module he is working on

core/classes/Misc/MentionsParser.php
```diff
@@ -11,49 +11,78 @@
...
-                        $value = preg_replace('/(?<!\/)' . preg_quote("@$possible_username", '/') . '/', '[user]' . $user->data()->id . '[/user]', $value);
-
-                        // Check if user is blocked by OP
-                        if ($value && $link && ($alert_full && $alert_short) && ($user->data()->id != $author_id) && !$user->isBlocked($user->data()->id, $author_id)) {
-                            Alert::create($user->data()->id, 'tag', $alert_short, $alert_full, $link);
-                            break;
-                        }
-                    }
-
-                    // chop last word off of it
```
partydragen, line 46: Why is that not needed anymore?

core/classes/Misc/MentionsParser.php
```diff
@@ -11,49 +11,78 @@
...
+
+    /**
+     * Get users from the database based on the provided nicknames. Filters out users have blocked the author.
+     */
+    private static function getRecipients(string $content, int $author_id): array
+    {
+        preg_match_all(self::USER_MENTIONS_REGEX, $content, $matches);
+        $nicknames = $matches[1];
+
+        return DB::getInstance()->query(
```
partydragen, line 72: User should be able to tag the user despite being blocked thought but not alert

core/classes/Misc/MentionsParser.php
tadhgboyle: I couldn't figure out why it was needed initially. Do you know what it does?

core/classes/Database/DB.php
tadhgboyle: Okay - thanks! Is the code for that one public at all? I'd like to see if we can instead improve our DB class instead of exposing its internal functions

partydragen: This is atleast one place where i see its used and looks to be a nice use case https://github.com/samerton/Nameless-Resources/blob/develop/upload/modules/Resources/classes/Resources.php#L121

core/classes/Misc/MentionsParser.php
tadhgboyle: fixed

core/classes/Database/DB.php
tadhgboyle: undid changes

partydragen (approved):

PR #3630, opened 2025-04-28
Remove internal deprecations

tadhgboyle:


core/classes/Group_Sync/GroupSyncManager.php
```diff
@@ -146,148 +146,6 @@ private function compileValidatorMessages(Language $language): array
...
-     * @param string $sending_injector_class Class name of injector broadcasting this change
-     * @param array  $group_ids              Array of Group IDs native to the sending injector which were added/removed to the user
-     *
-     * @deprecated broadcastGroupChange should be used instead. This function has issues, such as that on the first sync it will replace
-     *             groups on one side with groups from the other side, depending on which side happens to sync first. On the first sync,
-     *             the desired behaviour is for the new roles on both sides to become the union of roles on both sides.
-     *
-     * @return array Array of logs of changed groups
-     */
-    public function broadcastChange(User $user, string $sending_injector_class, array $group_ids): array
```
partydragen: @Derkades Whats the main reason for removing this where we just need a set of groups vs manually setting remove/add groups?

Your NamelessLink & Plugin is only submitting changed groups to api instead of full set of groups?

However i still pretty much depend on this function as other group sync providers such as Patreon Integration i only get to set Full set of groups as i myself wont be able to detect the changed groups

This function has apperard to be working fine thought for me in all cases as i always submit the full set of groups

Derkades: Sending a list of groups works for unidirectional sync with a single authoritative source, but not for bidirectional sync. If on Discord someone has groups A, B, C and on Discord they have groups C, D, you have no idea which groups should be added or removed on which side

tadhgboyle: I will revert this to the develop branch version and we can come back to this convo :)

modules/Core/pages/validate.php
```diff
@@ -29,10 +29,11 @@
             $user,
         ));
 
-        GroupSyncManager::getInstance()->broadcastChange(
+        GroupSyncManager::getInstance()->broadcastGroupChange(
             $user,
             NamelessMCGroupSyncInjector::class,
-            [$user->getMainGroup()->id]
+            [$user->getMainGroup()->id],
+            [],
```
partydragen: Same as above but this one can be removed, broadcastGroupChange is already triggered in the hook above

modules/Core/hooks/ValidateHook.php
```diff
@@ -21,10 +21,11 @@ public static function execute(UserValidatedEvent $event): void {
 
         $validate_user->setGroup(VALIDATED_DEFAULT);
 
-        GroupSyncManager::getInstance()->broadcastChange(
+        GroupSyncManager::getInstance()->broadcastGroupChange(
             $validate_user,
             NamelessMCGroupSyncInjector::class,
-            [VALIDATED_DEFAULT]
+            [VALIDATED_DEFAULT],
+            []
```
partydragen: A group might have been removed or no change at all at this point or several groups as user is losing all other groups at this point

You need to check all of users group if there has been any change and construct the array of removed groups, Or switch back to broadcastChange

partydragen (changes requested):

modules/Core/hooks/ValidateHook.php
tadhgboyle: done, thanks!

modules/Core/hooks/ValidateHook.php
```diff
@@ -14,17 +14,20 @@ public static function execute(UserValidatedEvent $event): void {
...
+        $groups_after = array_keys($validated_user->getGroups());
+        $groups_to_remove = array_diff($groups_before, $groups_after);
 
-        GroupSyncManager::getInstance()->broadcastChange(
-            $validate_user,
+        GroupSyncManager::getInstance()->broadcastGroupChange(
+            $validated_user,
             NamelessMCGroupSyncInjector::class,
-            [VALIDATED_DEFAULT]
+            [VALIDATED_DEFAULT],
```
partydragen: This still falsely triggering as user might already have the VALIDATED_DEFAULT group as many sites have Default and Validate group as same group

The old broadcastChange did this fine :S as it detected the real change in groups

Might just ignore this and lett go

partydragen (dismissed):

modules/Core/hooks/ValidateHook.php
tadhgboyle: fixed

PR #3631, opened 2025-04-28
Remove deprecated event system

partydragen:
Remove the old deprecated event system, Casuing to much trouble keeping it

+ Tagging is now also possible in Profile Posts and Messaging (No Alert)
+ Common Render events allowing for modules and all to use same rendering anywhere


modules/Forum/front_page.php
```diff
@@ -46,7 +46,7 @@
             'author_avatar' => $post_user->getAvatar(64),
             'author_group' => Output::getClean($post_user->getMainGroup()->name),
             'author_group_html' => $post_user->getMainGroup()->group_html,
-            'content' => EventHandler::executeEvent('renderPost', ['content' => $item['content']])['content'],
+            'content' => EventHandler::executeEvent(new RenderPostEvent($item['content']))['content'],
```
tadhgboyle: ❤️

modules/Core/hooks/ContentHook.php
```diff
@@ -17,10 +17,10 @@ class ContentHook extends HookBase {
      * @param array $params
      * @return array
      */
-    public static function codeTransform(array $params = []): array {
-        if (parent::validateParams($params, ['content'])) {
+    public static function codeTransform(AbstractEvent $event): void {
+        if (isset($event->content)) {
```
tadhgboyle, line 20: This won't work, AbstractEvent has no params set on it directly.

We perhaps should introduce a `AbstractContentEvent` which builtin events (such as your new `RenderPostEvent`) can extend to remove duplication of a `content `property, and make these content hooks more explicit.

core/classes/Events/AbstractEvent.php
```diff
@@ -65,9 +55,9 @@ final public function params(): array
      * EventHandler::executeEvent(new EventClass(...));
      * ```.
      *
-     * @return array|null The response from the event, or null if the event is not returning a response
+     * @return AbstractEvent The response from the event
      */
-    final public static function dispatch(): ?array
+    final public static function dispatch(): AbstractEvent
```
tadhgboyle: let's 🔥 this method, it was never used and defeats the purpose of the nice type checking we have now

core/classes/Events/EventHandler.php
```diff
@@ -28,41 +28,27 @@ public static function registerWebhooks(array $webhooks): void
      * Register an event.
      * This must be called in the module's constructor.
      *
-     * @param class-string|string $event       Name of event to add.
-     * @param string              $description Human readable description.
-     * @param array               $params      Array of available parameters and their descriptions.
-     * @param bool                $return      Whether to return $params afterwards
-     * @param bool                $internal    Whether to hide this hook from users in the StaffCP (ie for internal events)
+     * @param  class-string $event Event to add.
```
tadhgboyle:
```suggestion
     * @param  class-string<AbstractClass> $event Event to add.
```

This should work nicely here

core/classes/Events/EventHandler.php
```diff
@@ -72,30 +58,30 @@ public static function registerEvent(
...
     }
 
     /**
      * Register an event listener for a module.
      * This must be called in the module's constructor.
      *
-     * @param string                $event    Event name to listen to.
-     * @param callable|class-string $callback Listener callback to execute when event is executed. If class name is provided, we will assume there is a static "execute" method on the class.
-     * @param int                   $priority Execution priority - higher gets executed first
+     * @param  class-string          $event    Event to listen to.
```
tadhgboyle:
```suggestion
     * @param  class-string<AbstractEvent>          $event    Event to listen to.
```

core/classes/Events/EventHandler.php
```diff
@@ -111,52 +97,15 @@ public static function registerListener(string $event, $callback, int $priority
     /**
      * Execute an event.
      *
-     * @param  AbstractEvent|string $event  Event name to call, or instance of event to execute.
-     * @param  array                $params Params to pass to the event's function, not required when a class-based event is used.
-     * @return array|null           Response of lissteners, can be any type or null
+     * @param  AbstractEvent $event Instance of event to execute.
+     * @return AbstractEvent Response of listeners, can be any type or null
```
tadhgboyle:
```suggestion
      * @template T of AbstractEvent
      * @param T $event
      * @return T
```

This should work to give us better type hinting on the return value!

modules/Core/classes/Events/RenderContentEditEvent.php
```diff
@@ -0,0 +1,24 @@
+<?php
+
+class RenderContentEditEvent extends AbstractEvent {
+
+    public string $content;
+    public bool $skip_purify;
+
+    public function __construct(string $content, bool $skip_purify = false) {
```
tadhgboyle, line 12:
```suggestion
    public function __construct(string $content, bool $skip_purify) {
```

Should we make this mandatory? Otherwise I know from personal experience it's super easy to forget about 🙈 
Can we also add a phpdoc to indicate exactly what it means for the purification to happen?

modules/Core/classes/Events/RenderContentEvent.php
```diff
@@ -0,0 +1,24 @@
+<?php
+
+class RenderContentEvent extends AbstractEvent {
+
+    public string $content;
+    public bool $skip_purify;
+
+    public function __construct(string $content, bool $skip_purify = false) {
```
tadhgboyle, line 12: same as above ^

modules/Core/hooks/MentionsHook.php
```diff
@@ -18,74 +18,76 @@ class MentionsHook extends HookBase {
...
-            $params['content'] = MentionsParser::parseAndNotify(
-                $params['user']->data()->id,
-                $params['content'],
-                $params['alert_url'],
-                $params['mention_notification_type'],
-                $params['mention_notification_title'],
-            );
+    public static function preCreate(AbstractEvent $event): void {
+        if (!empty($event->content) && isset($event->user)) {
+            if (isset($event->alert_url, $event->mention_notification_type, $event->mention_notification_title)) {
```
tadhgboyle, line 23: Hmm, don't love all these required `isset` here... what do you think about some builtin event types such as `AbstractContentEvent extends AbstractEvent`?

modules/Core/hooks/MentionsHook.php
```diff
@@ -18,74 +18,76 @@ class MentionsHook extends HookBase {
...
                 static function (array $userData) {
                     [$userId, $userStyle, $userNickname, $userProfileUrl] = $userData;
                     return '<a href="' . $userProfileUrl . '" data-poload="' . URL::build('/queries/user/', 'id=' . $userId) . '" class="user-mention" style="' . $userStyle . '">@' . Output::getClean($userNickname) . '</a>';
                 }
             );
         }
+    }
 
-        return $params;
+    public static function stripPost(AbstractEvent $event): void {
```
tadhgboyle, line 71: hmm, this feels like a smell to me... can we not use the existing `stripContent`?

modules/Core/pages/profile.php
```diff
@@ -114,14 +114,17 @@
 
                             if ($validation->passed()) {
                                 // Validation successful
+                                $event = new ContentCreateEvent(Input::get('post'), $user);
+                                EventHandler::executeEvent($event);
```
tadhgboyle, line 118: ❤️

modules/Core/hooks/MentionsHook.php
partydragen: The Text::embedSafe function also used StripPost so i needed to make seperate function that returns string, while the one you tagged is using the event system

modules/Core/classes/Events/RenderContentEditEvent.php
partydragen: Should stay false by default, The user should clearly know what they're doing if they skip, Anyways i did add phpdoc

Sam also agrees It should be a conscious decision to skip

modules/Core/hooks/MentionsHook.php
partydragen: I don't get it you want a new event `AbstractContentEvent extends AbstractEvent`??

Or a class that the event extends/implements, However that class only only contain content being much having one for Content, User then alerts, and one for skip_content its all so different

Also if we do class then its bigger chance for future breaking changes for modules if we need to change the class

core/classes/Events/EventHandler.php
```diff
@@ -28,41 +28,27 @@ public static function registerWebhooks(array $webhooks): void
...
+        $description = fn () => $event::description();
+        $internal = $event::internal();
+
         // Don't re-register if the event already exists, just update the params
         // and description. This is to "fix" when registerListener is called
         // for an event that has not been registered yet.
         if (isset(self::$_events[$name])) {
             self::$_events[$name] = [
                 'description' => $description,
                 'internal' => $internal,
```
tadhgboyle, line 51:
```php
       foreach (self::$_events as $name => $meta) {
            $class = $meta['class_name'];

            if (!is_subclass_of($class, DiscordDispatchable::class) && !is_subclass_of($class, HasWebhookParams::class)) {
                continue;
            }

            if (is_callable($meta['description'])) {
                $description = $meta['description']();
            } else {
                $description = $meta['description'];
            }

            $return[$name] = [
                'description' => $description,
                'supports_discord' => is_subclass_of($class, DiscordDispatchable::class),
                'supports_normal' => is_subclass_of($class, HasWebhookParams::class),
            ];
        }
```

I don't think we need to have this 'internal' flag exist anymore

modules/Forum/classes/Events/PrePostCreateEvent.php
```diff
@@ -0,0 +1,30 @@
...
+        $this->content = $content;
+        $this->user = $user;
+        $this->alert_url = $alert_url;
+        $this->mention_notification_type = $mention_notification_type;
+        $this->mention_notification_title = $mention_notification_title;
+    }
+
+    public static function name(): string {
+        return 'prePostCreate';
+    }
```
tadhgboyle: `AbstractEvent#name` will do this for you (same applies to the other new event classes you've made)
```suggestion
```

core/classes/Events/AbstractEvent.php
```diff
@@ -28,16 +28,6 @@ public static function name(): string
      */
     abstract public static function description(): string;
```
tadhgboyle:
```suggestion
   public static function description(): string
   {
       return static::name();
   }
```

no need to define this manually, wouldn't hurt to default to the name here

tadhgboyle (approved):

PR #3633, opened 2025-04-29
Tighten unused language terms checks

tadhgboyle:
Removes language terms that were false positives with the old lenient grepping checks


custom/panel_templates/Default/core/groups_form.tpl
```diff
@@ -99,15 +99,6 @@
                                     <input type="number" min="1" class="form-control" id="InputOrder" name="order"
                                         value="{$GROUP_ORDER_VALUE}">
                                 </div>
-                                {if $DISCORD_INTEGRATION}
```
tadhgboyle, line 102: this variable was never set, this is old code from before the discord module

partydragen (approved):

PR #3638, opened 2025-05-02
Revised Email Templates

FluxxyBoi:


FluxxyBoi:
Looking for inputs on the updates. Let me know what I can do next! 
How the updated template looks:
image

FluxxyBoi:
Register:
image
TFA:
image
Reset Pass:
image

> The variables, i.e: server name, message, subject, thanks, greeting etc are set to my testing server's data, but are set as correct variables in the code files.

partydragen:
Amazing work, Thanks

I see its some hardcoded text thats need solving but lets wait on the new email templating system to be finished

FluxxyBoi:
@partydragen ,

Yeah, as I mentioned, the buttons are using hardcoded text. It's pretty straightforward to fix it, but yeah, waiting for the new email templating system would be a much better idea.

Just let me know when it's out, and I'll be happy to commit rest of the updates.

PR #3647, opened 2025-05-05
Remove redundant try/catch blocks

tadhgboyle:
Most of these are super ancient and guarded very foundational pieces of our code - like simple DB calls. Removing them as I see no reason to wrap in try/catch for simple stuff like this - kept other ones where the result was more important.


tadhgboyle:
Diff is weird. I just removed the try and catch, and the unintended the logic. no logic changes were made

PR #3649, opened 2025-05-08
Make sure error reporting is disabled by default

supercrafter100:
Since PHP 8 it seems like some defaults have changed, so this PR explicitly disabled error logging when debugging is not enabled, see https://www.php.net/manual/en/errorfunc.configuration.php.

### DIFF
--- a/index.php
+++ b/index.php
@@ -26,7 +26,7 @@ if (defined('DEBUGGING') && DEBUGGING) {
 } else {
     ini_set('display_startup_errors', 0);
     ini_set('display_errors', 0);
-    error_reporting(NULL);
+    error_reporting(null);
 }
 
 // Ensure PHP version >= 8.2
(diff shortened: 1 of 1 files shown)


index.php
```diff
@@ -23,6 +23,10 @@
     ini_set('display_startup_errors', 1);
     ini_set('display_errors', 1);
     error_reporting(-1);
```
Derkades: Since PHP 8.0.0, error_reporting is set to E_ALL by default so `error_reporting` can be removed

index.php
```diff
@@ -23,6 +23,10 @@
...
+} else {
+    ini_set('display_startup_errors', 0);
+    ini_set('display_errors', 0);
+    error_reporting(null);
```
Derkades: error_reporting should still be enabled I think? We don't want errors to be displayed on the page, but we still want to log them

Derkades (changes requested):
[display_errors](https://www.php.net/manual/en/errorfunc.configuration.php#ini.display-errors) indeed defaults to "1", that's unexpected! The documentation does not say the default has changed.

PR #3652, opened 2025-05-08
Add base Model system (ActiveRecord + Casts + Relations) for clean DB access

GIGABAIT93:
Add Model system with ActiveRecord pattern, CastManager, and Relation support

- Introduced abstract Model class with:
  - automatic table name resolution with prefix
  - CRUD methods: find, findOrFail, create, save, delete, all
  - attribute casting (primitive types, JSON, enums, custom Castable support)
  - eager-loaded relations (hasMany, belongsTo)
  - lifecycle event hooks

- Added QueryBuilder class:
  - where, whereIn, orderBy, limit clauses
  - eager loading support via with(), including nested relations
  - CRUD operations and increment/decrement helpers

- Added CastManager class for handling read/write attribute casts

- Lazy loading relations is prohibited (throws exception) to encourage clean eager loading

- Overall: simple, efficient and extensible base model system for use in core and modules


tadhgboyle (commented):
Thanks for this contribution! I've always wanted to get some type of proper Model support in Nameless. 

Off the bat, I'm very hesitant to take on maintaining our own ORM. This is a very complex problem area full of possible issues which I just don't think we will be able to properly maintain to keep this clean and secure. My long term plan has been to eventually implement something like illuminate/database - a well tested and maintained library. Things get really hairy quickly with ORMS - take a look at Rails ActiveRecord codebase for an example. 

Before we make any decisions, can you please push up an example usage of this?
IE: User model with hasMany Groups, and vice versa?
It's a bit unclear how this can fit in seamlessly right now without a huge amount of work to update every one of our models. Ideally this is a slow and steady rollout over time since we have so many database objects.

GIGABAIT93:
> Thanks for this contribution! I've always wanted to get some type of proper Model support in Nameless.
> 
> Off the bat, I'm very hesitant to take on maintaining our own ORM. This is a very complex problem area full of possible issues which I just don't think we will be able to properly maintain to keep this clean and secure. My long term plan has been to eventually implement something like illuminate/database - a well tested and maintained library. Things get really hairy quickly with ORMS - take a look at Rails ActiveRecord codebase for an example.
> 
> Before we make any decisions, can you please push up an example usage of this? IE: User model with hasMany Groups, and vice versa? It's a bit unclear how this can fit in seamlessly right now without a huge amount of work to update every one of our models. Ideally this is a slow and steady rollout over time since we have so many database objects.

Hi I added a file with examples of use [here](https://github.com/NamelessMC/Nameless/blob/d886dfe5397a91c96ef5dec831d961b52332f17b/core/classes/Database/ORM/examples.md)

samerton:
Thank you for your efforts with this PR, it's a very interesting subject.

I am wondering if, in its current form, Nameless actually needs something like an ORM just yet. I do understand that data models do need improving, however I'm of the opinion that _not_ using an ORM leads to more freedom for module authors - for example, I've been working on a new version of the Resources module on-and-off for a while now, and I've created some repository classes with SQL queries defined just as I need them.

Even if we do go down the route of an ORM one day, I agree with @tadhgboyle in that it may be best to use the likes of illuminate/database or Doctrine, as tried and tested libraries for this.

I'd be interested to hear thoughts of others on this.

GIGABAIT93:
> Thank you for your efforts with this PR, it's a very interesting subject.
> 
> I am wondering if, in its current form, Nameless actually needs something like an ORM just yet. I do understand that data models do need improving, however I'm of the opinion that _not_ using an ORM leads to more freedom for module authors - for example, I've been working on a new version of the Resources module on-and-off for a while now, and I've created some repository classes with SQL queries defined just as I need them.
> 
> Even if we do go down the route of an ORM one day, I agree with @tadhgboyle in that it may be best to use the likes of illuminate/database or Doctrine, as tried and tested libraries for this.
> 
> I'd be interested to hear thoughts of others on this.

In fact, using Illuminate/Database would be great. The only reason I have created it because I have a few modules in which I use my own models and I would not like to duplicate the code of the base models for every module of my module

PR #3657, opened 2025-05-14
Fix table html purification

### LINKED ISSUE #3461: Tables do not work in pages
### Describe the issue

Table outlines/borders do not work when being selected.

When creating a table it shows the borders. When the page gets published they disappear.
Even when selecting solid line borders and selecting a color for the border it does not translate to the published page.
In fact it gets reset as soon as you update the page.

### NamelessMC Version

Development version

### To Reproduce

1. Create any page, custom, in modules, etc
2. Create a table and fill it with items
3. Submit and view.
4. change border to solid line and select any color
5. safe changes
6. view, borders do not appear
7. go back to editor and see border line changes are reset.

### Expected Behaviour

A table with rows and columns being shown.

### Screenshots

image
image
image

### Additional Information

Ive tested it with default template as well as a custom template. Issue persists.
Version: 2.1.2

tadhgboyle:
Fixes https://github.com/NamelessMC/Nameless/issues/3461

Also:
- Makes forum post/topic editing take up full column width (right now its pretty slim and TinyMCE hides 90% of the options)
- Fixes forum topic "Guests" using invalid smarty key

### DIFF
--- a/core/classes/Core/Output.php
+++ b/core/classes/Core/Output.php
@@ -57,7 +57,7 @@ class Output
             $purifierConfig->set('URI.DisableExternalResources', false);
             $purifierConfig->set('URI.DisableResources', false);
             $purifierConfig->set('HTML.Allowed', 'u,a,p,p[style],b,i,small,blockquote,span[style],span[class],p,strong,em,li,ul,ol,div[align],br,img,figure,figcaption');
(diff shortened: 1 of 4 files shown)


TeemoCell:
Since you changed that already. Can you edit `Creating topic in Forum` tho? I was about to make a pull for that.

vtX7ixW

tadhgboyle:
> Since you changed that already. Can you edit `Creating topic in Forum` tho? I was about to make a pull for that.
> 
> vtX7ixW

Fixed

PR #3662, opened 2025-05-17
Create and implement `Cache#fetch`

tadhgboyle:
Replaces most instances of `isCached` -> `retrieve` or `store` with `fetch` which does it all in one. There are quite a few leftover cases of this pattern but I believe they are okay to be left alone


partydragen:
Merged without approvel hmmmm xD

PR #3664, opened 2025-05-21
Update Avatar config to use Settings system rather than Cache

tadhgboyle:
First of (hopefully) many PRs to stop using Cache for configuration!


modules/Core/pages/panel/avatars.php
tadhgboyle, line 1: Surprisingly messy, we seem to have been in a middle state where we were writing both to the cache and unused setting names

modules/Core/classes/Avatars/NamelessMCAvatarSource.php
tadhgboyle, line 1: I don't think this one has worked in forever, I also see no benefit in maintaining our own when it offers the same functionality of existing 3rd party ones

partydragen (dismissed):

PR #3667, opened 2025-05-21
Allow editing user titles from UserCP

tadhgboyle:


partydragen:
There should be permission to edit the user title

modules/Core/pages/user/settings.php
```diff
@@ -289,11 +297,14 @@
 
                     $gravatar = $_POST['gravatar'] == '1' ? 1 : 0;
 
+                    $user_title = $user->hasPermission('usercp.title') ? Input::get('user_title') : null;
```
samerton: Maybe instead of default `null` this should be `$user->data()->user_title` so that titles applied by a site administrator aren't lost

tadhgboyle: Good catch thanks

PR #3670, opened 2025-05-21
Update Discord widget config to use Settings system rather than Cache

tadhgboyle:


modules/Discord Integration/includes/admin_widgets/discord.php
```diff
@@ -10,28 +10,20 @@
...
-$cache->setCache('social_media');
-
 if (Input::exists()) {
     if (Token::check()) {
         if (isset($_POST['theme'])) {
-            $cache->store('discord_widget_theme', $_POST['theme']);
+            Settings::set('discord_widget_theme', $_POST['theme'], 'Discord Integration');
         }
 
-        $cache->store('discord', $guild_id);
```
tadhgboyle, line 21: This was never read, we use `Settings::get('discord')` for the guild Id

PR #3672, opened 2025-05-22
Update images config to use Settings system rather than Cache

tadhgboyle:


core/classes/Database/DatabaseInitialiser.php
```diff
@@ -296,8 +296,7 @@ private function initialiseTemplates(): void
             $config_path = '/' . trim($config_path, '/');
         }
 
-        $this->_cache->setCache('backgroundcache');
-        $this->_cache->store('banner_image', $config_path . '/uploads/template_banners/homepage_bg_trimmed.jpg');
+        Settings::set('banner_image_path', $config_path . '/uploads/template_banners/homepage_bg_trimmed.jpg');
```
tadhgboyle, line 300: Not 100% sure about the naming here. I think having the `_path` suffix is good for clarity?

PR #3673, opened 2025-05-22
Misc refactors + fixes

tadhgboyle:


partydragen:
Why remove server query, Its even still used for group sync selection

Anyways its still nice feature to have

PR #3674, opened 2025-05-22
Enhancements URL and Pages

GIGABAIT93:
### Enhancements to `Pages` class (`core/classes/Core/Pages.php`):

* Added a new method `getPageByName` to retrieve page details by name. This method iterates through `_pages` to find a matching name and returns the page information or `null` if not found.
* Introduced `getPagesByModule` to fetch all pages associated with a specific module, using `array_filter` for efficient filtering.
* Updated return types for `getPageById` and `getPageByURL` methods to explicitly allow `null` when no page is found, improving type safety. [[1]](diffhunk://#diff-6bf976a03e1913935fbb6ea7a3ee1bbbf857bbe19681fcb79039c89c0279ea92L130-R132) [[2]](diffhunk://#diff-6bf976a03e1913935fbb6ea7a3ee1bbbf857bbe19681fcb79039c89c0279ea92L149-R151)
* Added the `url` key to the page array in the `add` and `addCustom` methods, ensuring the URL is explicitly stored for each page. [[1]](diffhunk://#diff-6bf976a03e1913935fbb6ea7a3ee1bbbf857bbe19681fcb79039c89c0279ea92R51) [[2]](diffhunk://#diff-6bf976a03e1913935fbb6ea7a3ee1bbbf857bbe19681fcb79039c89c0279ea92R70)

### Enhancements to `URL` class (`core/classes/Core/URL.php`):

* Added `getSegments`, a utility to break a URI into an array of segments, removing leading/trailing slashes and empty segments. This simplifies URI parsing.
* Introduced `segment`, a method to retrieve a specific URI segment by index, with an optional default value if the segment is absent.


supercrafter100 (approved):

tadhgboyle:
Thanks for the contribution! Did you have specific use cases in mind for these? I'm hesitant to add more methods which aren't used at least internally

GIGABAIT93:
> Thanks for the contribution! Did you have specific use cases in mind for these? I'm hesitant to add more methods which aren't used at least internally

I use it in my module eg getting a segment to determine slug `$slug = URL::segment(2);`
Also in my module I for pages I give the names so it would be convenient to get a page by name 
`if($pages->getActivePage()['name'] === $pages->getPageByName('index')['name']){}`
Also as an example you can get all the pages of the module and automatically add them to navigation
I think it would be useful methods

PR #3677, opened 2025-05-23
Add auto upgrader

tadhgboyle:
- DB transactions don't seem to be working (ie: maintenance mode enabled, error happens, it's still enabled)
- Need a way to block multiple from running at once (locks on queue tasks?)
- Should we disable and then re-enable modules?
- Should we try to backup their DB first?
- Hide update banner if task is ongoing?
- Can we check ahead for any potential issues? (tmp not writable, ZipArchive doesn't exist)
- Utilize checksum to validate the zip before proceeding


partydragen:
Must check entire NamelessMC is writeable thought as ALOT of users dont have files writeable

Should also check disk usage

Currently a issue with disable/enable modules as i must load modules in a certain order

PR #3678, opened 2025-05-24
Add backup feature

tadhgboyle:
image

image


partydragen:
Amazing work!

Add a option to set automaticly reschedule backup task daily
Maybe even option to delete all backups after X latest backups

Display warning on update page to backup site first? :D

tadhgboyle:
> Amazing work!
> 
> Add a option to set automaticly reschedule backup task daily Maybe even option to delete all backups after X latest backups
> 
> Display warning on update page to backup site first? :D

Added those settings @partydragen!
Will add a link to create a backup on the update page too, thanks for the suggestions!

core/includes/updates/230.php
```diff
@@ -38,6 +38,16 @@ public function run(): void
         Settings::set('discord_widget_theme', $discord_widget_theme, 'Discord Integration');
         $this->_cache->eraseAll();
 
+        // Add admincp.core.backups permission to Admin group
+        $admin_group = Group::find(1);
+        $permissions = json_decode($admin_group->permissions, true);
+        if (!isset($permissions['admincp.core.backups'])) {
+            $permissions['admincp.core.backups'] = 1;
+            DB::getInstance()->update('groups', $admin_group->id, [
+                'permissions' => json_encode($permissions),
```
partydragen, line 69: This is not needed, All admins group has the administrator perm

custom/panel_templates/Default/core/backups.tpl
```diff
@@ -0,0 +1,150 @@
...
+                                <h5>{$EXISTING}</h5>
+
+                                <div class="table-responsive">
+                                    <table class="table table-striped table-hover">
+                                        <thead>
+                                            <tr>
+                                                <th>{$FILENAME}</th>
+                                                <th>{$DATE_CREATED}</th>
+                                                <th>{$FILE_SIZE}</th>
+                                                <th>Actions</th>
```
partydragen: Hardcoded language term

Derkades:
We must be very careful to deny access to /backups in .htaccess and the nginx configuration file. Maybe Nameless can try making a request to /backups/somefile, and refuse to make a backup if it succeeds?

tadhgboyle:
> We must be very careful to deny access to /backups in .htaccess and the nginx configuration file. Maybe Nameless can try making a request to /backups/somefile, and refuse to make a backup if it succeeds?

I changed this so that it stores backups in `/cache/backups`, I think this is much safer since we already guard the `/cache` folder from being publicly accessible - what do you think?
FWIW, in the near future I'd love to rename this to `/storage`, and have things like `/storage/cache`, `/storage/backups`, `/storage/logs`, etc

core/includes/updates/230.php
```diff
@@ -66,6 +66,16 @@ public function run(): void
...
```
partydragen, line 69: All of this can still be removed, We have not updated admin group perms for ages as it already had Adminstrator perm

partydragen (approved):
Should prop have sam review this too, Dont want any security leaks of this,

Also are we sure backup folder is properly protected that people cant guess the link to download the zip directly

Like i see it uses time hard to guess exact time but if its displayed in staffcp then any staff might find out about the direct download link

PR #3680, opened 2025-06-05
Remove group template permissions

tadhgboyle:


partydragen:
Must still keep per group perms aswell its still nice feature, dont want user switch to dev templates or if a site even testing a new template for the staff

Why remove working features at all that not making any problems

PR #3686, opened 2025-06-09
Add middleware system

tadhgboyle:
Introduces middleware system. I've implemented it in a few areas which are loaded globally and on the frontend which do redirects based on user/site state (maintenance mode, force TFA, etc). This helps us remove some global procedural code as we try to get to a cleaner internal state.


modules/Core/classes/Middleware/Global/MaintenanceModeMiddleware.php
```diff
@@ -0,0 +1,40 @@
...
+        }
+
+        // Allow admin users to bypass maintenance mode
+        if ($user->isLoggedIn() && $user->canViewStaffCP()) {
+            // Display notice to admin stating maintenance mode is enabled
+            define('BYPASS_MAINTENANCE', true);
+            return;
+        }
+
+        Redirect::to(URL::build('/maintenance'));
```
partydragen, line 38: Should not send users to maintenance page thats just annoying, And alot of users like just to refresh site to see if maintenance is complete

Also makes uses lose any details they might have in the URL

modules/Core/classes/Middleware/Frontend/EnsureUserIntegrationsLinkedMiddleware.php
```diff
@@ -0,0 +1,30 @@
+<?php
+
+class EnsureUserIntegrationsLinkedMiddleware extends AbstractMiddleware
+{
+    public MiddlewareType $type = MiddlewareType::Frontend;
+
+    public array $exemptRoutes = [
+        '/user/connections',
+        '/oauth',
+        '/user/settings',
```
partydragen, line 10: User should only be able to view /user/settings IF they are viewing enable_tfa

modules/Core/classes/Middleware/Global/TFAMiddleware.php
```diff
@@ -0,0 +1,48 @@
...
+    ];
+
+    public function handle(User $user, Request $request, Language $language): void
+    {
+        // Only process for logged-in users
+        if (!$user->isLoggedIn()) {
+            return;
+        }
+
+        // Skip if AJAX request, such as Alert or PM checks
```
partydragen, line 28: Was well also to prevent the code from refreshing when user switch to next step as that refreshing the site

modules/Core/classes/Middleware/Global/TFAMiddleware.php
```diff
@@ -0,0 +1,48 @@
...
+ * @package NamelessMC\Hooks
+ * @author Aberdeener
+ * @version 2.3.0
+ * @license MIT
+ */
+class TFAMiddleware extends AbstractMiddleware
+{
+    public array $exemptRoutes = [
+        '/logout',
+        '/user/settings' // Allow access to settings to enable TFA
```
partydragen, line 18: Same here, ONLY allow enable_tfa section, 

Now user details is exposed can can be modifyed and can even disable 2fa without using 2fa to login

partydragen (changes requested):

partydragen:
Able to make modules add/modify exemptRoutes?

PR #3687, opened 2025-06-26
add auto_reload config key to TwigTemplateEngine.php

graphiteisaac:
Howdy, this is a problem I've come across as I've tried making a template for Nameless using the new Twig engine, if you upload a template, it caches once and pretty much never refreshes the cache.

I believe the ideal behaviour would be for the cache to reload when the source file changes, the same as the Smarty engine, and this line of config should achieve that. To my knowledge, there isn't a big performance penalty or anything.

Reference to the config is available in the [Twig docs](https://twig.symfony.com/doc/3.x/api.html#environment-options)

Cheers :)


tadhgboyle (approved):
Need to familiarize myself with Twig internals - not sure what it does with the cache path we provide if the auto_reload essentially makes it re-parse on each page load

PR #3695, opened 2025-07-17
External resources installation

tadhgboyle:


partydragen:
Very nice so far, Glad this finally being made :D

PR #3732, opened 2026-06-04
Dynamic page loading

partydragen:
Add dynamic page loading, Allows to eaiser create friendly urls and having ids in path
such as

/community/{id}
/community/{id}/edit
/community/{id}/edit/members

current old behavior does not allow this ^^ other than smashing it all into same page, now we can properly split it all into own files

### DIFF
--- a/index.php
+++ b/index.php
@@ -109,39 +109,46 @@ $route = rtrim(strtok($_GET['route'], '?'), '/');
 
 $all_pages = $pages->returnPages();
 
-if (array_key_exists($route, $all_pages)) {
-    $pages->setActivePage($all_pages[$route]);
-    if (isset($all_pages[$route]['custom'])) {
-        require(implode(DIRECTORY_SEPARATOR, [ROOT_PATH, 'modules', 'Core', 'pages', 'custom.php']));
-        die;
-    }
+foreach ($all_pages as $page_route => $page) {
+    $regex = preg_replace(
+        '/\{([a-zA-Z0-9_]+)\}/',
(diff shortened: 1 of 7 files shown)


index.php
```diff
@@ -109,41 +109,48 @@
...
-        die;
-    }
-
-    $path = implode(DIRECTORY_SEPARATOR, [ROOT_PATH, 'modules', $all_pages[$route]['module'], $all_pages[$route]['file']]);
+foreach ($all_pages as $page_route => $page) {
+    $regex = preg_replace(
+        '/\{([a-zA-Z0-9_]+)\}/',
+        '(?P<$1>[^/]+)',
+        $page_route
+    );
```
supercrafter100, line 117: Maybe explain through a comment here what this code does instead of letting users guess

index.php
```diff
@@ -109,41 +109,48 @@
...
-    for ($i = count($path_array) - 2; $i > 0; $i--) {
-        $new_path = '/';
-        for ($n = 1; $n <= $i; $n++) {
-            $new_path .= $path_array[$n] . '/';
+    if (preg_match($regex, $route, $matches)) {
+        $route_params = [];
+        foreach ($matches as $key => $value) {
+            if (!is_int($key)) {
+                $route_params[$key] = $value;
+            }
```
supercrafter100, line 126: Mention why this is_int is needed, this is a side effect of the regex constructed earlier and the named parameters

supercrafter100 (changes requested):

